PR #44, opened 2022-03-23
chore: Add cause traceback to internal server exceptions

### LINKED ISSUE #40: Add cause traceback to internal server exceptions
https://github.com/momentohq/client-sdk-java/pull/127

poppoerika:
Closes #40

### DIFF
--- a/src/CacheServiceErrorMapper.ts
+++ b/src/CacheServiceErrorMapper.ts
@@ -24,7 +24,7 @@ export function cacheServiceErrorMapper(
     case Status.UNKNOWN:
     case Status.ABORTED:
     case Status.UNAVAILABLE:
-      return new InternalServerError(err?.message);
+      return new InternalServerError(`${err?.message}\n${err?.stack || ''}`);
     case Status.NOT_FOUND:
       return new NotFoundError(err?.message);
     case Status.OUT_OF_RANGE:
@@ -43,5 +43,7 @@ export function cacheServiceErrorMapper(
     case Status.ALREADY_EXISTS:
       return new AlreadyExistsError(err?.message);
   }
-  return new InternalServerError('unable to process request');
+  return new InternalServerError(
+    `unable to process request\n${err?.stack || ''}`
+  );
 }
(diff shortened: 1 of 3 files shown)


src/CacheServiceErrorMapper.ts
```diff
@@ -43,5 +43,7 @@ export function cacheServiceErrorMapper(
     case Status.ALREADY_EXISTS:
       return new AlreadyExistsError(err?.message);
   }
-  return new InternalServerError('unable to process request');
+  return new InternalServerError(
+    `unable to process request\n${err?.stack || ''}`
```
eaddingtonwhite, line 47: why not make this same as line 27? any reason not to include message as well?

src/CacheServiceErrorMapper.ts
```diff
@@ -24,7 +24,7 @@ export function cacheServiceErrorMapper(
     case Status.UNKNOWN:
     case Status.ABORTED:
     case Status.UNAVAILABLE:
-      return new InternalServerError(err?.message);
+      return new InternalServerError(`${err?.message}\n${err?.stack || ''}`);
```
bruuuuuuuce, line 27: I think instead of string formatting, you can add a second param to `InternalServerError` and set the stack to `err?.stack`

https://github.com/momentohq/client-sdk-javascript/blob/295045a94ba24592b0257382f3986915670ab299/src/Errors.ts#L83

so it would become
```
export class InternalServerError extends CacheServiceError {
  constructor(message: string, stack?: string) {
    super(message);
    this.name = 'InternalServerError';
    this.stack = stack
  }
}
```

src/CacheServiceErrorMapper.ts
kvcache: Can we use a `cause` in our exceptions? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error

poppoerika: Hmm... maybe I'm not understanding this but it doesn't seem that `ErrorConstructor` does not have `options` or `cause` property?

image

poppoerika: I implemented what Matt suggested below.

kvcache: just do matt's thing - I was only pushing back at the string formatting of a stack.

poppoerika: 👍

bruuuuuuuce (approved):
:shipit:

PR #50, opened 2022-03-25
fix: Update CD workflow

poppoerika:


.github/workflows/cd.yml
```diff
@@ -38,19 +52,13 @@ jobs:
...
-        with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          allow-initial-development-versions: true
-          force-bump-patch-version: true
-
-      - name: Publish npm package to Artifactory
+      - name: Publish package to npm
         run: |
-          export VERSION="${{ steps.semrel.outputs.version }}"
+          export VERSION="${{ needs.release.outputs.version }}"
```
eaddingtonwhite, line 57: Is this right with needs. Not step.?

poppoerika: Yup, following the same procedure as the client protos which successfully publish package to npm from the cd workflow.
https://github.com/momentohq/client_protos/blob/main/.github/workflows/push.yml#L61
I think I needed to add `needs: release` [here](https://github.com/momentohq/client-sdk-javascript/pull/50/files#diff-ea3ea8c9932adc7ba8161ceda844fedd43b006848ef1140c050cbd7ea0788a18R30)

eaddingtonwhite (approved):

PR #53, opened 2022-03-30
feat: add public lamba layer

bruuuuuuuce:


.github/workflows/cd.yml
```diff
@@ -61,3 +61,38 @@ jobs:
...
+            --license-info "Apache License 2.0" \
+            --zip-file layer.zip \
+            --compatible-runtimes nodejs14.x nodejs12.x | jq '.Version')
+          
+          aws lambda add-layer-version-permission \
+            --layer-name ${LAYER_NAME} \
+            --version-number ${LAYER_VERSION} \
+            --statement-id "Allow all aws accounts access to momento-sdk-javascript lambda layer, version: ${VERSION}" \
+            --principal * \
+            --action lambda:GetLayerVersion
```
cprice404, line 97: I have a minor pref for putting this logic in a shell script, maybe in a new directory called 'build-scripts' or something.  It might allow for re-usability for dev purposes, and in general I think it's best to keep the LOC of YAML for these things to a minimum.  But I don't feel too strongly about it if you disagree.

Also we might want to create some follow-up tickets about (a) creating some kind of canary or integration test for this, and (b) adding documentation for it.  If this PR is going to make this layer public right now, I think we should probably have some way of very explicitly declaring that it is experimental and subject to change.

.github/workflows/cd.yml
```diff
@@ -61,3 +61,38 @@ jobs:
...
+        run: |
+          export VERSION="${{ needs.release.outputs.version }}"
+          echo "Publishing lambda layer with version $VERSION"
+          mkdir -p nodejs/node_modules
+          pushd nodejs/node_modules
+            npm i @gomomento/sdk@${VERSION}
+          popd
+          zip -r layer.zip nodejs
+          LAYER_NAME="momento-sdk-javascript"
+          LAYER_VERSION=$(aws lambda publish-layer-version \
```
cprice404, line 85: Is there some way that we can explictly specify the layer version so that we can maintain some sort of mapping between the layer version number and our SDK version number?

poppoerika (commented):
This is just a question since I'm not familiar with Lambda layers.
This lambda layer helps users who uses our JS SDK in their lambda without them installing the SDK package from npm?

.github/workflows/cd.yml
bruuuuuuuce: Yep agree on the shell script. That was we can reuse in both ci.yaml and cd.yaml. 

+1 on the follow up tickets. I can publish it as `-snapshot` to imply that its expiremental for now. Or just `-expiremental`

.github/workflows/cd.yml
bruuuuuuuce: Unfortunately we can't do that :sad-vir:. Amazon auto increments the version of the lambda layer for us. It also has to be a whole number. I can see if I can tag the layer with the version of the sdk its using

.github/workflows/cd.yml
```diff
@@ -61,3 +61,38 @@ jobs:
...
+          role-duration-seconds: 1200
+          role-session-name: javascript-sdk-lambda-layer-deployer
+
+      - name: Publish lambda layer
+        run: |
+          export VERSION="${{ needs.release.outputs.version }}"
+          echo "Publishing lambda layer with version $VERSION"
+          mkdir -p nodejs/node_modules
+          pushd nodejs/node_modules
+            npm i @gomomento/sdk@${VERSION}
```
cprice404, line 81: one other ticket we should consider filing on this - auditing the dependency list of the sdk.

For lambda runtimes, historically one of the biggest contributors to cold start issues is the size of the layer/lambda source code.  And node_modules dirs are notorious for getting filled up with a large volume of dependencies and cruft that isn't always needed.  So for the lambda use case we're really going to want to audit all of our dependencies and make sure we need all of them.

Out of curiosity do you know how big your zip file currently ends up being?

.github/workflows/cd.yml
bruuuuuuuce: O I mean naming it, `momento-sdk-javascript-experimental`, there are no built in flags for it

cprice404: ah okay that works too

.github/workflows/cd.yml
bruuuuuuuce: Right now its 2.8M

cprice404: that's not too bad.  probably still worth filing a ticket, but we don't need to tackle it right away.

bruuuuuuuce: I think tho layers are a bit different than a normal lambda runtime. These layers don't get pulled down from s3 with the normal lambda code. There is some special sauce amazon does behind the scenes with them.

cprice404: yeah i agree that the process must be different but i would still expect that the binary has to be transferred to the hosts, and that the startup cost of that is still going to be relative to the size of the binary.  probably worth mentioning your point in the ticket we create though, and if/when we pick up that ticket we can do some googling on how layers relate to cold starts.

cprice404 (approved):
awesome thanks for filing the tickets!

PR #62, opened 2022-03-30
chore: fix illegal statement id chars part 3

bruuuuuuuce:


rlinehan:
It seems like it would be really nice if something could catch this before PRs are merged... but it's a shell script so testing sucks Would it be reasonable to do this in cdk instead? Is that even possible?

bruuuuuuuce:
> It seems like it would be really nice if something could catch this before PRs are merged... but it's a shell script so testing sucks Would it be reasonable to do this in cdk instead? Is that even possible?

Agreed testing like this is :dumpster_fire:. We can do this via cdk, and I think this is definitely something to consider. I didn't initially do this via cdk because,
1. its just 2 aws commands
2. we don't have a user it github that can deploy cdk code to the public artifacts account
3. these sdk repos don't have any cdk in them currently

We can make that user ^^^, but I wanted the users perms to be scoped as minimally as possible, so literally only have access to update lambda layers and thats it.

Would be interested to hear @cprice404 opinions about this.

poppoerika (approved):

PR #63, opened 2022-03-30
fix: Update integration tests to delete existing cache before creating one

poppoerika:
Based on our conversations [here](https://gomomento.slack.com/archives/C037XLHFC79/p1648677574935629).

### DIFF
--- a/integration/SimpleCacheClient.test.ts
+++ b/integration/SimpleCacheClient.test.ts
@@ -43,10 +43,10 @@ const deleteExistingCache = async (
   momento: SimpleCacheClient,
   cacheName: string
 ) => {
-  const listCacheResp = await momento.listCaches();
-  const cacheList = listCacheResp.getCaches().map(c => c.getName());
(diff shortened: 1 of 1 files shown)


integration/SimpleCacheClient.test.ts
```diff
@@ -39,10 +39,22 @@ const removeSystemCredentials = () => {
...
 };
 
+const deleteExistingCache = async (
+  momento: SimpleCacheClient,
+  cacheName: string
+) => {
+  const listCacheResp = await momento.listCaches();
+  const cacheList = listCacheResp.getCaches().map(c => c.getName());
+  if (cacheList.includes(cacheName)) {
+    await momento.deleteCache(cacheName);
```
bruuuuuuuce, line 49: Since we now have limits on control plane api calls, you can just surround the delete call with a try catch and swallow the `CacheDoesntExist` error

kvcache (approved):
lgtm

PR #67, opened 2022-04-21
feat: add support for signing keys APIs

tylerburdsall:
Adds support for `CreateSigningKey`, `RevokeSigningKey`, and `ListSigningKeys`

### DIFF
--- a/integration/SimpleCacheClient.test.ts
+++ b/integration/SimpleCacheClient.test.ts
@@ -4,6 +4,7 @@ import * as os from 'os';
 import {SimpleCacheClient, TimeoutError} from '../src';
 import {AlreadyExistsError, NotFoundError} from '../src/Errors';
 import {TextEncoder} from 'util';
+import { ListSigningKeysResponse } from '../src/messages/ListSigningKeysResponse';
 
 const AUTH_TOKEN = process.env.TEST_AUTH_TOKEN;
 if (!AUTH_TOKEN) {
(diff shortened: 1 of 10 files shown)


integration/SimpleCacheClient.test.ts
```diff
@@ -4,6 +4,7 @@ import * as os from 'os';
 import {SimpleCacheClient, TimeoutError} from '../src';
 import {AlreadyExistsError, NotFoundError} from '../src/Errors';
 import {TextEncoder} from 'util';
+import { ListSigningKeysResponse } from '../src/messages/ListSigningKeysResponse';
```
wandaitzuchen, line 7: nit: remove spaces around {} to be consistent in formatting?

tylerburdsall: Oh good catch, I'm surprised `eslint` didn't fix this. Thanks, I'll fix it!

wandaitzuchen (dismissed):
lgtm!

src/Momento.ts
```diff
@@ -115,9 +118,78 @@ export class Momento {
...
+        {interceptors: this.interceptors},
+        (err, resp) => {
+          if (err) {
+            reject(cacheServiceErrorMapper(err));
+          } else {
+            resolve(new CreateSigningKeyResponse(endpoint, resp));
+          }
+        }
+      );
+    });
```
kvcache, line 140: nit: Instead of copy-pasting this block around you could make a higher order function.

src/Momento.ts
```diff
@@ -115,9 +118,78 @@ export class Momento {
...
     if (!name.trim()) {
       throw new InvalidArgumentError('cache name must not be empty');
     }
   };
+
+  private validateTtlMinutes = (ttlMinutes: number) => {
+    if (ttlMinutes && ttlMinutes < 0) {
+      throw new InvalidArgumentError('ttlMinutes must be positive');
+    }
+  };
```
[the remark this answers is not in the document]
tylerburdsall: `null/undefined` can't be passed in since the parameter is a non-null `number`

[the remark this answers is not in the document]
tylerburdsall: Honestly this was just to keep with the existing pattern, happy to move this to a standard utility function declaration

src/Momento.ts
tylerburdsall: That's true, I can do that

src/Momento.ts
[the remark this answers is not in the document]
kvcache: Every other function in this file is a normal function. The `validateCacheName` function property is the one that is nonconforming. Instead of adding to the nonconformity maybe fix it (since it's a small cosmetic change next to stuff you're already changing) and do your thing the way the majority of the file does things?

wandaitzuchen (approved):

PR #70, opened 2022-06-08
chore: Update end-user facing documentation.

### LINKED ISSUE #69: update javascript-sdk documentation
Document all end-user facing cache client methods.

malandis:
Closes #69

### DIFF
--- a/src/SimpleCacheClient.ts
+++ b/src/SimpleCacheClient.ts
@@ -11,18 +11,29 @@ import {CreateSigningKeyResponse} from './messages/CreateSigningKeyResponse';
 import {RevokeSigningKeyResponse} from './messages/RevokeSigningKeyResponse';
 import {ListSigningKeysResponse} from './messages/ListSigningKeysResponse';
 
+/**
+ * Momento Simple Cache Client.
+ *
+ * Features include:
+ * - Get, set, and delete data
+ * - Create, delete, and list caches
+ * - Create, revoke, and list signing keys
+ * @export
+ * @class SimpleCacheClient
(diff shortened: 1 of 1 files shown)


src/SimpleCacheClient.ts
```diff
@@ -16,9 +16,13 @@ export class SimpleCacheClient {
...
 
   /**
-   * @param {string} authToken
-   * @param {number} defaultTtlSeconds
-   * @param {number} requestTimeoutMs
+   * Create a SimpleCacheClient.
+   * @param {string} authToken - Momento token to authenticate requests with Simple Cache Service.
+   * @param {number} defaultTtlSeconds - A default time to live, in seconds, for cache objects
+   * created by this client.
+   * @param {number} requestTimeoutMs - An optional timeout in milliseconds for get and set operations
```
bruuuuuuuce, line 23: This is probably my fault, but looking at this pr I notice that we are defining optional params in 3 different ways 

- saying its optional in the help text
- `{number=}`
- `{number | undefined}`

while we are updating the docs, maybe we should choose 1 approach and consolidate all the jsdocs. My vote would be for `{number=}` as its the most concise.

bruuuuuuuce: wdyt?

malandis: I opted to use the JSDoc style square bracket style  `{number} [requestTimeoutMs]`

bruuuuuuuce (approved):
:shipit:

tylerburdsall (approved):

PR #73, opened 2022-06-30
fix: copy down package.json into dist during publish

eaddingtonwhite:
We currently read our package.json file in programmatically to obtain sdk version. This commit copies down package.json into dist folder during publish so code can access it in a consistent way.

Loosely following advice off [this](https://stackoverflow.com/questions/38858718/npm-script-copy-package-json-to-dist-when-bundling) SO post. Feels like there should be cleaner way to do this like we do in python for example with `pkg_resources.get_distribution("momento").version` but javascript ecosystem does not seem to have a very consistent strategy for doing this.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -52,5 +52,6 @@ jobs:
           echo "Publishing package with version $VERSION"
           mv package.json package.json.ORIG
           cat package.json.ORIG|jq ". += {\"version\": \"$VERSION\"}" > package.json
+          cp package.json dist/package.json
           npm publish
         shell: bash


malandis (approved):
Reasonable stopgap  until we can investigate further.

cprice404:
this seems reasonable, i've definitely had to do things like this before, but what I'm missing is how our package.json isn't getting included in the `dist` in the current setup, prior to this PR?

cprice404 (approved):
curious about why it's not already ending up in dist, if we know that, but not opposed to this.

eaddingtonwhite:
@cprice404 its copied in at top level of package but not down in dist. So structure looks like this currently which doesnt work w/ how code imports.

```console
➜  sdk git:(main) ✗ ll
total 40
-rw-r--r--  1 eaw  staff    11K Jun 30 15:29 LICENSE
-rw-r--r--  1 eaw  staff   1.9K Jun 30 15:29 README.md
drwxr-xr-x  3 eaw  staff    96B Jun 30 15:29 dist
-rw-r--r--  1 eaw  staff   1.4K Jun 30 15:29 package.json
➜  sdk git:(main) ✗ ls dist
src
➜  sdk git:(main) ✗
```

You can see full package file structure on publish [here](https://github.com/momentohq/client-sdk-javascript/runs/7140398549?check_suite_focus=true) see how isnt copied down where code expects it now :/

PR #74, opened 2022-06-30
fix: try different way including package.json in published bundle

eaddingtonwhite:
Still having issues with package json after last fix. Was able to verify this locally w/ 
```
npm run build
npm pack
```

I believe it should work on publish now and copy step added previously might not actually be needed.
```
✗ npm pack
npm notice
npm notice 📦  @gomomento/sdk@0.0.1
npm notice === Tarball Contents ===
npm notice 11.4kB LICENSE
npm notice 1.9kB  README.md
npm notice 1.6kB  dist/package.json   ## <--- Can see now!!!
npm notice 186B   dist/src/CacheServiceErrorMapper.d.ts
npm notice 6.7kB  dist/src/CacheServiceErrorMapper.js
npm notice 3.0kB  dist/src/Errors.d.ts
npm notice 13.1kB dist/src/Errors.js
...
```

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -57,7 +57,6 @@ jobs:
           echo "Publishing package with version $VERSION"
           mv package.json package.json.ORIG
           cat package.json.ORIG|jq ". += {\"version\": \"$VERSION\"}" > package.json
-          cp package.json dist/package.json
           npm publish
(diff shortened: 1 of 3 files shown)


malandis (approved):

cprice404 (commented):
nice

PR #82, opened 2022-07-07
fix: memory issues for retries; improvements to logging

cprice404:
This commit fixes a memory issue that surfaced on the retry logic
under higher load.

It also adds a broader array of log levels for users to choose
from, and moves the noisiest log messages to `trace`.  In addition,
there are some places where we had been logging the cache value;
those are removed because for large cache values the log messages
were overwhelming.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2427,9 +2427,9 @@
       }
     },
     "node_modules/electron-to-chromium": {
-      "version": "1.4.181",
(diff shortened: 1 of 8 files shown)


mjindalmotiv8 (dismissed):

danielamiao (approved):
giphy

eaddingtonwhite (approved):

PR #84, opened 2022-07-13
feat: performance improvements for high load

cprice404:
This commit is focused on improving performance for high load,
but also includes some other general cleanups.

Highlights:
* tune grpc memory setting to avoid client-side RESOURCE_EXHAUSTED
* tune grpc subchannel setting to improve performance with multiple
  clients
* SimpleCacheClient now round-robins requests across multiple instances
  of MomentoCache, to spread load across a larger number of grpc channels
* Fixes to logging implementation that were impacting performance and
  stability
* Simplify logging configuration
* Rename files to `kebab-case` to be more idiomatic to typescript
* Fixes all outstanding compiler warnings

For discussion:
* Adds 'request-id' header to requests


src/grpc/headers-interceptor.ts
```diff
@@ -36,10 +37,12 @@ export class HeaderInterceptor {
     return (options, nextCall) => {
       return new InterceptingCall(nextCall(options), {
         start: (metadata, listener, next) => {
+          metadata.add('request-id', randomUUID());
```
cprice404, line 40: we should talk about this.  I add this because it was useful in helping me debug some things related to retries (otherwise it is very difficult to distinguish requests from one another).  I also assume that it will be generally useful for server-side traces, and possibly for client-side traces/metrics.  But it does add some bytes to the payload, so happy to discuss, and possibly remove / defer to a later time.

src/grpc/headers-interceptor.ts
```diff
@@ -36,10 +37,12 @@ export class HeaderInterceptor {
...
           this.headersToAddEveryTime.forEach(h =>
             metadata.add(h.name, h.value)
           );
           if (!HeaderInterceptor.areOnlyOnceHeadersSent) {
+            HeaderInterceptor.areOnlyOnceHeadersSent = true;
```
cprice404, line 45: this wasn't being set before.  That meant that the agent header was being sent on every request.  I "fixed" it but have mixed feelings about it so would be open to discussing whether this is actually the behavior we want here.

src/grpc/retry-interceptor.ts
```diff
@@ -77,45 +72,48 @@ export class RetryInterceptor {
...
                 retries++;
                 const newCall = nextCall(options);
-                // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
                 newCall.start(metadata, {
                   onReceiveMessage: function (message) {
-                    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
                     savedReceiveMessage = message;
                   },
                   onReceiveStatus: function (status) {
+                    const requestId: string = JSON.stringify(
```
cprice404, line 104: if we decide not to add `request-id` for now then this stuff will need to be removed as well.

src/momento-cache.ts
```diff
@@ -37,30 +38,37 @@ export class MomentoCache {
...
     );
     this.client = new cache.cache_client.ScsClient(
       props.endpoint,
-      ChannelCredentials.createSsl()
+      ChannelCredentials.createSsl(),
+      {
+        // default value for max session memory is 10mb.  Under high load, it is easy to exceed this,
+        // after which point all requests will fail with a client-side RESOURCE_EXHAUSTED exception.
+        // This needs to be tunable: https://github.com/momentohq/dev-eco-issue-tracker/issues/85
+        'grpc-node.max_session_memory': 256,
```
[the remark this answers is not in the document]
cprice404: did a few more test runs and I think 256 is high enough for now.

src/grpc/headers-interceptor.ts
eaddingtonwhite: my 2 cents is its 100% worth the extra bytes having this on every request will make operations much easier prevent us wasting time trouble shooting issues especially now that we are starting to do client side retrys.

Would rather start with this on by default and wait for customer to have issue with bandwidth and then provide an optional way to toggle this header off for very advanced high load users.

I feel similarly about user agent header being sent on every request. Would rather we default to easier operations and reasonable performance and come back and solve for the extreme case when needed and provide ways to toggle off for advanced users.

src/grpc/headers-interceptor.ts
eaddingtonwhite: It was performance optimization to cut down on bytes sent. TBH feel its bit of pre-mature optimization see comment on request-id header. That being said I dont care about this one as much vs request-ids so would not be end of world if we keep on only first request like it is now.

src/grpc/headers-interceptor.ts
mjindalmotiv8: If we are not tracking this at the service layer I am not sure how we will use this? Is it for client side debugging?

src/simple-cache-client.ts
```diff
@@ -54,22 +61,36 @@ export class SimpleCacheClient {
...
+    // https://github.com/momentohq/oncall-tracker/issues/186
+    const numClients = 6;
+    this.dataClients = range(numClients).map(
+      () =>
+        new MomentoCache({
+          authToken,
+          defaultTtlSeconds,
+          endpoint: dataEndpoint,
+          requestTimeoutMs: options?.requestTimeoutMs,
+        })
```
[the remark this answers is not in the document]
cprice404: could do that, let me think about it.  I don't know if we are expecting that customers might ever instantiate this class directly.  I'll keep this in mind when we do the follow-up PR to make this configurable.

src/utils/logging.ts
```diff
@@ -31,37 +31,57 @@ export interface LoggerOptions {
...
-): Logger {
-  const loggerOptions = options ?? defaultLoggerOptions();
+interface InitializedLoggerOptions {
+  level: LogLevel;
+  format: LogFormat;
+}
+
+const _MOMENTO_LOGGING_OPTIONS: InitializedLoggerOptions = {
+  level: LogLevel.WARN,
+  format: LogFormat.CONSOLE,
```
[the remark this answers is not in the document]
cprice404: interesting.  I can change the default to JSON and just use CONSOLE in the load test then.  Thanks for the heads up.

src/grpc/headers-interceptor.ts
bruuuuuuuce: I am 💯 on board for request ids

danielamiao: I'm onboard with adding this but it might be helpful to cross check this with OTel tracing specs (not metrics, don't hate, their tracing protocol is pretty good :P) to ensure we use the standard header tag. There is grpc middleware that automatically links up your traces if you use standard headers (we probably don't have that middleware wrapped around the service yet but we could add it easily later). I can help with this later today if needed.

cprice404:
@eaddingtonwhite do these integration test failures make any sense to you

src/grpc/headers-interceptor.ts
kvcache: I'm pretty sure the grpc middleware we have on mr2 already does this @danielamiao 

Agree otlp tracing is not bad at all (otel tracing library though...)

I'm not a fan of doing this for every request. We're a cache, not a database. We should support this but have it default to off - it's an extra 32 bytes per request of overhead paid for by users. Also, it's unnecessary tracking data which we should be cautious with and mindful of.

bruuuuuuuce: > We should support this but have it default to off 

We seem to keep running into these types of issues that require us to go to the customer to tell them to do something so we can debug. I would prefer if we could debug customer requests without having to ask the customer to help us.

src/simple-cache-client.ts
```diff
@@ -195,9 +220,19 @@ export class SimpleCacheClient {
...
     );
   }
+
+  private getNextDataClient(): MomentoCache {
+    if (this.nextDataClientIndex >= this.dataClients.length) {
+      this.nextDataClientIndex = 0;
+    }
+    const client = this.dataClients[this.nextDataClientIndex];
+    this.nextDataClientIndex++;
+    return client;
```
kvcache, line 236:
```suggestion
    const client = this.dataClients[this.nextDataClientIndex];
    this.nextDataClientIndex = (this.nextDataClientIndex + 1) % this.dataClients.length;
    return client;
```
nit: unnecessary condition

eaddingtonwhite:
@cprice404 no that is new one to me have not seen. Id be curious if we ran again if failed in same way consistently.

src/grpc/headers-interceptor.ts
cprice404: If we are generally okay with the idea of this existing, but possibly being configurable, we can defer the decision about how to configure it and what the defaults should be until https://github.com/momentohq/dev-eco-issue-tracker/issues/85 , which I intend to pick up very soon.

I don't want to make it configurable in this PR.  For this PR I think the options are either always-on, or remove this code for now.

I can double-check against the otel specs for header name once we have agreed whether or not to even include this in this PR.

src/grpc/headers-interceptor.ts
cprice404: need others to weigh in here.  honestly i'd prefer to send this on every request as well.  I'm pretty sure I know what Kenny's opinion is.  Would like to know where @danielamiao , @mjindalmotiv8 , anyone else who is interested in weighing in on this stand.

src/grpc/headers-interceptor.ts
kvcache: And you're going to do that with the client's uuid how matt?

It's extra cost footed by the user per request, bloating 100% of requests with tracking information that is extremely debatably useful for 0.000001% of requests.. and again, how are you going to do anything with it without asking the user for something? It's logged client side at debug. If debug logging isn't enabled, then it's completely wasted network transfer - a client side id generator for our service to.. idk what actually

Also, fyi, we do head sampling of request traces. There is a real possibility that uuids a user asks you about will have no trace on an active system and you won't be able to disambiguate "your request never reached us" from "I didn't trace that request"

What is the value offered to users that is worth the cost of 32 bytes additional per request overhead to their cache? Happy to take this to slack

src/grpc/headers-interceptor.ts
mjindalmotiv8: I think there is value in this during debugging and having the feature to enable it on per customer selectively during debugging is most likely what we should be doing. Doing on each request is going to add more processing on mr2 and given that we care about microseconds worth of operations, I am worried this will add latency.

kvcache (dismissed):
The request id code as it sits is a bug from my perspective. I understand you're making changes to how the client is configured so whatever change ordering you like seems fine. Ideally we wouldn't merge this as-is to the release branch though if that was the intent.

src/grpc/headers-interceptor.ts
danielamiao: @cprice404 let's remove it from this PR to unblock this PR for now. That way we can agree on defaults + use the otlp tracing header in the follow up PR in one go.

src/grpc/headers-interceptor.ts
cprice404: configurability is not in scope for this PR, so I will leave this as-is, meaning that it will only be sent once per client.

src/grpc/headers-interceptor.ts
cprice404: @danielamiao agreed, I will do that.

cprice404: removed in https://github.com/momentohq/client-sdk-javascript/pull/84/commits/912471012bc256cf275766de7e83ad7e06efd1e9

src/grpc/retry-interceptor.ts
cprice404: removed in https://github.com/momentohq/client-sdk-javascript/pull/84/commits/912471012bc256cf275766de7e83ad7e06efd1e9

src/utils/logging.ts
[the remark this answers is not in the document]
cprice404: changed default to JSON in https://github.com/momentohq/client-sdk-javascript/pull/84/commits/912471012bc256cf275766de7e83ad7e06efd1e9

cprice404:
> @cprice404 no that is new one to me have not seen. Id be curious if we ran again if failed in same way consistently.

fixed in https://github.com/momentohq/client-sdk-javascript/pull/84/commits/23034ee0f83cbfa55e794901ef1cd359591c3256 along with some other test cleanups

kvcache (approved):

eaddingtonwhite (approved):

cprice404:
@danielamiao anything else you want to see on this one before it goes in?

danielamiao:
@cprice404 nope looks good from my end, thanks!

PR #85, opened 2022-07-14
fix: attempt to work around semver plugin's branch restriction

cprice404:
The semver plugin that we are using appears to try to enforce that
the releases only be done from the default branch.  This commit
attempts to disable that check.  Filed this issue to see if they
might support configuring an allowed branch to do releases from:

https://github.com/go-semantic-release/action/issues/25

### DIFF
--- a/.github/workflows/on-push-to-release-branch.yml
+++ b/.github/workflows/on-push-to-release-branch.yml
@@ -25,7 +25,7 @@ jobs:
           #  to actually have anything to do with CI:
           # https://github.com/go-semantic-release/semantic-release/blob/master/cmd/semantic-release/main.go#L173-L194
           # https://github.com/go-semantic-release/condition-github/blob/4c8af3fc516151423fff2f77eb08bf7082570676/pkg/condition/github.go#L42-L44
-	      custom-arguments: '--no-ci'
+          custom-arguments: '--no-ci'
 
       - name: Output release
         id: release
--- a/.github/workflows/on-push-to-release-branch.yml
+++ b/.github/workflows/on-push-to-release-branch.yml
@@ -1,4 +1,4 @@
-name: Push to Release Branch
+name: On push to release branch
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


malandis (dismissed):
Looks like there's no other workaround besides changing the default branch to release.

.github/workflows/on-push-to-release-branch.yml
```diff
@@ -20,6 +20,12 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           allow-initial-development-versions: true
           force-bump-patch-version: true
+	  # For whatever reason, this silly tool won't let you do releases from branches
+          #  other than the default branch unless you pass this flag, which doesn't seem
+          #  to actually have anything to do with CI:
+          # https://github.com/go-semantic-release/semantic-release/blob/master/cmd/semantic-release/main.go#L173-L194
+          # https://github.com/go-semantic-release/condition-github/blob/4c8af3fc516151423fff2f77eb08bf7082570676/pkg/condition/github.go#L42-L44
+	  custom-arguments: '--no-ci'
```
eaddingtonwhite, line 28: Spacing seems off on this. I think maybe will fail with yaml format error? :skeletor:

.github/workflows/on-push-to-release-branch.yml
```diff
@@ -20,6 +20,12 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           allow-initial-development-versions: true
           force-bump-patch-version: true
+          # For whatever reason, this silly tool won't let you do releases from branches
+          #  other than the default branch unless you pass this flag, which doesn't seem
+          #  to actually have anything to do with CI:
+          # https://github.com/go-semantic-release/semantic-release/blob/master/cmd/semantic-release/main.go#L173-L194
+          # https://github.com/go-semantic-release/condition-github/blob/4c8af3fc516151423fff2f77eb08bf7082570676/pkg/condition/github.go#L42-L44
+	      custom-arguments: '--no-ci'
```
eaddingtonwhite, line 28: Still seems off?

cprice404: oh i didn't even notice the red thing, that's a nice feature.  thanks github :)

PR #86, opened 2022-07-14
chore: add a workflow for triggering manual release

cprice404:
This commit adds a manual workflow to start the release process.
The workflow does a merge from the `main` branch to the `release`
branch, which will then trigger the release via the push to the
`release` branch.

### DIFF
--- a/.github/workflows/manual-release.yml
+++ b/.github/workflows/manual-release.yml
@@ -1,0 +1,20 @@
+name: Manual Release
+# Triggers a merge from main->release, which will then trigger a release
+# from the release branch.
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  sync-branch:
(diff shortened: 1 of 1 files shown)


.github/workflows/manual-release.yml
```diff
@@ -0,0 +1,20 @@
...
+  workflow_dispatch:
+
+jobs:
+  sync-branch:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@main
+
+      - name: Merge main -> release
+        uses: devmasx/merge-branch@master
```
danielamiao, line 15: I wonder if this works if the branch is protected? I thought there is protection against directly merging without a PR.

cprice404: Check out the branch protection for that branch and lmk what you think.  I set it up so that only repo admins can push to it, but it does not require a PR.

PR #98, opened 2022-08-06
feat: Add MomentoSigner class

rlinehan:
Add new class for creating presigned URLs and user signed access tokens, using signing keys registered with the service via `CreateSigningKey`.

Pulls in the node module `jose` ([github](github.com/panva/jose)/[npm](https://www.npmjs.com/package/jose)) for handling importing JWKs and signing JWTs. This library is linked from the [openID docs](https://openid.net/developers/jwt/) and has 2 million weekly downloads. It's MIT licensed and appears to be active. It seems to support both Browsers and Node. It is somewhat large but doesn't have any dependencies, from the [docs](https://github.com/panva/jose#bundle-size-package-size-tree-shaking):
> Yes the bundle size is on the larger side, that is because each module is actually published multiple times so that it can remain truly without dependencies and be universal / isomorphic.
>
> Nevertheless, since each module can be required independently and is fully tree-shakeable, the install size should not be a cause for concern.

So it seems like a reasonable choice for our use case.

### DIFF
--- a/src/momento-signer.ts
+++ b/src/momento-signer.ts
@@ -7,16 +7,50 @@ export enum CacheOperation {
 }
 
 export interface signingRequest {
+  /**
+   * @param {string} cacheName - Name of the cache the signed token should be allowed to operate on.
+   */
   cacheName: string;
+  /**
+   * @param {number} expiryEpochSeconds - Timestamp in epoch seconds for
+   * when the signed token should expire.
+   */
   expiryEpochSeconds: number;
-  cacheOperation?: CacheOperation;
+  /**
+   * @param {string} [cacheKey] - The cache key the signed token is allowed to
+   * access. If not present, auth token is allowed to be used for any key.
+   */
   cacheKey?: string;
+  /**
+   * @param {CacheOperation} [cacheOperation] - The operation the signed token
+   * should be valid for. If not present the signed token is allowed to be used
+   * for any operation.
+   */
+  cacheOperation?: CacheOperation;
+  /**
+   * @param {number} [ttlSeconds] - The time to live for cache items that this
+   * signed token can set. If not present the signed token is allowed to be
+   * used for set operations with any ttl.
+   */
   ttlSeconds?: number;
 }
 
 export interface presignedUrlRequest extends signingRequest {
-  cacheOperation: CacheOperation;
+  /**
+   * @param {string} cacheKey - The cache key the presigned
+   * URL should access.
+   */
   cacheKey: string;
+  /**
+   * @param {CacheOperation} cacheOperation - The operation the presigned
+   * URL should perform on the item in the cache.
+   */
+  cacheOperation: CacheOperation;
+  /**
+   * @param {number} [ttlSeconds] - The time to live for cache items when
+   * performing a SET. Required when operation is CacheOperation.Set.
+   */
+  ttlSeconds?: number;
 }
 
 interface headers extends JWTHeaderParameters {
@@ -25,10 +59,21 @@ interface headers extends JWTHeaderParameters {
   kid: string;
 }
 
-// the jose.ImportJWK returns Promise<KeyLike | Uint8Array>. It doesn't really
+// `jose.importJWK` returns Promise<KeyLike | Uint8Array>. It doesn't really
 // matter to us which, so this is just an alias for convenience purposes.
 type joseKey = KeyLike | Uint8Array;
 
+/**
+ * Momento Signer
+ *
+ * Features include:
+ * - Create signed jwts for use with HTTP requests to the Momento Simple Cache
+ * Service
+ * - Create presigned URLs to use for HTTP requests to the Momento Simple
+ * Cache Service
+ * @export
+ * @class MomentoSigner
+ */
 export class MomentoSigner {
   private readonly key: joseKey;
   private readonly headers: headers;
@@ -38,6 +83,15 @@ export class MomentoSigner {
     this.headers = headers;
   }
 
+  /**
+   * Creates a new MomentoSigner instance with the specified private key
+   *
+   * @param {string} jwkJsonString - the JSON string of the JWK key. Can be
+   * optained from `createSigningKey` response.
+   * @returns {Promise<MomentoSigner>} - Promise containing the MomentoSigner
+   * instance
+   * @memberof MomentoSigner
+   */
   static async init(jwkJsonString: string): Promise<MomentoSigner> {
     let jwk: JWK;
     try {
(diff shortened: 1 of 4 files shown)


rlinehan:
This is failing with

```
npm ERR! code E401
npm ERR! Unable to authenticate, your authentication token seems to be invalid.
npm ERR! To correct this please trying logging in again with:
npm ERR!     npm login
```

🤔

bruuuuuuuce:
> This is failing with
> 
> ```
> npm ERR! code E401
> npm ERR! Unable to authenticate, your authentication token seems to be invalid.
> npm ERR! To correct this please trying logging in again with:
> npm ERR!     npm login
> ```
> 
> 🤔

@rlinehan you need to run

```
rm -rf ~/.npmrc
rm -rf package-lock.json node_modules
npm i
```

bruuuuuuuce (approved):
:shipit: love the jsdocs!!!

src/momento-signer.ts
```diff
@@ -0,0 +1,213 @@
...
+    const headers = buildHeaders(jwk);
+
+    return new MomentoSigner(key, headers);
+  }
+
+  /**
+   * Creates a new presigned URL
+   *
+   * @param {string} hostname - hostname of Simple Cache Service. Can be
+   * optained from `createSigningKey` response.
```
danielamiao, line 120: NIT: typo `optained` => `obtained`

src/momento-signer.ts
```diff
@@ -0,0 +1,213 @@
...
+  /**
+   * @param {CacheOperation} [cacheOperation] - The operation the signed token
+   * should be valid for. If not present the signed token is allowed to be used
+   * for any operation.
+   */
+  cacheOperation?: CacheOperation;
+  /**
+   * @param {number} [ttlSeconds] - The time to live for cache items that this
+   * signed token can set. If not present the signed token is allowed to be
+   * used for set operations with any ttl.
```
danielamiao, line 33: This is more of a comment to keep in mind for our service in the future: it's probably not very helpful to have an exact TTL constraint. We should really allow the signer to specify a `max_ttl` constraint, and requests must only set TTLs _up to that value_, instead of enforcing the exact TTL value on the backend service (again, not relevant to this PR, so just a drive-by comment).

rlinehan: Yeah that makes a lot of sense to me, but agreed its outside the scope of this PR. I can open an issue for it though.

src/momento-signer.ts
rlinehan, line 120:
```suggestion
   * obtained from `createSigningKey` response.
```

src/momento-signer.ts
```diff
@@ -0,0 +1,213 @@
+import {InvalidArgumentError} from '.';
+import {importJWK, JWK, JWTHeaderParameters, KeyLike, SignJWT} from 'jose';
+
+export enum CacheOperation {
+  Get = 'GET',
+  Set = 'SET',
+}
+
+export interface signingRequest {
```
gautamomento, line 9: I am not very sure what style guide we use but is `lowerCamelCase` the right way to name interfaces. We do `UpperCamelCase` for classes. Just seems a bit inconsistent.

src/momento-signer.ts
```diff
@@ -0,0 +1,213 @@
...
+   * @memberof MomentoSigner
+   */
+  public async createPresignedUrl(
+    hostname: string,
+    presignedUrlRequest: presignedUrlRequest
+  ): Promise<string> {
+    const jwtToken = await this.signAccessToken(presignedUrlRequest);
+    const cacheName = presignedUrlRequest.cacheName;
+    const cacheKey = presignedUrlRequest.cacheKey;
+    if (presignedUrlRequest.cacheOperation === CacheOperation.Get) {
```
gautamomento, line 133: nit: Would a switch read better here instead of if-else-if-else. Just thinking if in future we want to extend this for other methods.

danielamiao (dismissed):

gautamomento (commented):
This is looking great, left some comments.

It would be worthwhile to have @malandis or Chris take a pass at this too. They have been deeply thinking about conventions, style etc.

src/momento-signer.ts
rlinehan: Oh good call... I think you're probably right that `SigningRequest` would be better here and same for `PresignedUrlRequest` below.

src/momento-signer.ts
rlinehan: Oh, now I remember... I had tried this but it resulted in
```
/Users/ruth/momento/client-sdk-javascript/src/momento-signer.ts
  137:9  error  Unexpected lexical declaration in case block  no-case-declarations
```

I think because of the assignment of `const ttlSeconds = presignedUrlRequest.ttlSeconds` in the `Set` case? But thanks to stack overflow I just learned you can get around this by putting in `{}` 🤦‍♀️  TIL.

src/momento-signer.ts
rlinehan: Done in 771fd7f

src/momento-signer.ts
rlinehan: Done in 301d044

src/momento-signer.ts
```diff
@@ -0,0 +1,215 @@
...
+ * Momento Signer
+ *
+ * Features include:
+ * - Create signed jwts for use with HTTP requests to the Momento Simple Cache
+ * Service
+ * - Create presigned URLs to use for HTTP requests to the Momento Simple
+ * Cache Service
+ * @export
+ * @class MomentoSigner
+ */
```
malandis, line 76: Could you include a minimum viable example of the workflow in this header? Would be great in-IDE documentation for users.

malandis (changes requested):

src/momento-signer.ts
rlinehan: Sure, I can add that! There's also a PR up in the [examples repo](https://github.com/momentohq/client-sdk-examples/pull/102) with a more extensive example.

malandis: That's great. As long as we have an example ready to go to demo the usage, you can reference the examples repo one too.

src/momento-signer.ts
```diff
@@ -0,0 +1,233 @@
...
+ *   const keyString = signingKeyResponse.getKey();
+ *   const endpoint = signingKeyResponse.getEndpoint();
+ *   const signer = await MomentoSigner.init(keyString)
+ *   const getUrl = await signer.createPresignedUrl(endpoint, {
+ *     cacheName: 'mycache',
+ *     cacheKey: 'mykey',
+ *     cacheOperation: CacheOperation.GET,
+ *     expiryEpochSeconds: Math.floor(Date.now() / 1000) + 10 * 60;
+ *   });
+ *   fetch(getUrl)
```
malandis, line 86: Markdown around code block:

```suggestion
 *   ```const signingKeyResponse = await momentoClient.createSigningKey(60 * 24)
 *   const keyString = signingKeyResponse.getKey();
 *   const endpoint = signingKeyResponse.getEndpoint();
 *   const signer = await MomentoSigner.init(keyString)
 *   const getUrl = await signer.createPresignedUrl(endpoint, {
 *     cacheName: 'mycache',
 *     cacheKey: 'mykey',
 *     cacheOperation: CacheOperation.GET,
 *     expiryEpochSeconds: Math.floor(Date.now() / 1000) + 10 * 60;
 *   });
 *   fetch(getUrl)```
```

malandis (changes requested):
Thank you. This looks great. Just a couple minor style issues.

malandis (approved):
:shipit:

PR #110, opened 2022-11-11
chore: adds code scanning workflow

eaddingtonwhite:
Want to give githubs `codeql` tool a try to see if it can catch anything and also will make our public repos look a bit more legit I think. I already enabled on dev docs repo to try it out and looked pretty good. Wanted to start with this SDK see how things went. If looks good here will work to roll out to our other repos over time.


eaddingtonwhite:
Looks like were good here. 👍 
Screenshot 2022-11-11 at 2 07 06 PM

cprice404 (approved):

PR #112, opened 2022-11-17
fix: bug in debug logging

eaddingtonwhite:
Fixes seeing debug logs like this every time.
```bash
{"level":20,"time":1668709906140,"pid":78770,"name":"MomentoCache","msg":"Creating cache client using endpoint: 'cache.cell-ap-northeast-1-1.prod.a.momentohq.com'"}
{"level":20,"time":1668709906154,"pid":78770,"name":"MomentoCache","msg":"Request timeout: undefined"}
```

Also lowers concurrency since we made limits more aggressive for GA launch and ups timeout to be more inline for what we are defaulting in SDK.


cprice404 (approved):
:ty:

tylerburdsall (approved):

PR #114, opened 2022-12-01
fix: Introduce idle wrapper around gRPC clients

cprice404:
In some environments such as AWS Lambda, it is possible for
a cache client to be instantiated and stored in a long-lived
(global) variable, which then may not be accessed for long
periods of time (e.g. between lambda invocations).

The lambda environment may suspend the runtime in such a way
that no code will be executing between invocations (i.e., no
keepalive pings).  If there are more than 5 minutes between
lambda invocations, then the server side of the cache client
(AWS NLB) will detect the connection as idle and close it.

On the next lambda invocation, the grpc-js library does not
recognize that the connection has been closed so it will
continue to try to send data to it, resulting in a client-side
timeout (DEADLINE_EXCEEDED).

In this commit we add a wrapper around the client that ensures
that if it has been idle for too long, we will explicitly
refresh it to make sure that the connection is usable before
we issue requests against it.

### DIFF
--- a/src/grpc/idle-grpc-client-wrapper.ts
+++ b/src/grpc/idle-grpc-client-wrapper.ts
@@ -1,0 +1,61 @@
+import {getLogger, Logger} from '../utils/logging';
+import {CloseableGrpcClient, GrpcClientWrapper} from './grpc-client-wrapper';
+
+// TODO: This should not be defined here, it should be part of the Configuration object when we
+// introduce that.
+const MAX_IDLE_MILLIS = 4 * 60 * 1_000; // 4 minutes.  We want to remain comfortably underneath the idle timeout for AWS NLB, which is 350s.
+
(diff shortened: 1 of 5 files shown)


src/grpc/idle-grpc-client-wrapper.ts
```diff
@@ -0,0 +1,61 @@
+import {getLogger, Logger} from '../utils/logging';
+import {CloseableGrpcClient, GrpcClientWrapper} from './grpc-client-wrapper';
+
+// TODO: This should not be defined here, it should be part of the Configuration object when we
```
eaddingtonwhite, line 4: + good call out

eaddingtonwhite (approved):
Great work @cprice404 Thanks for jumping on this.

kvcache (approved):
This makes sense to me.

PR #128, opened 2022-12-15
feat: More load-gen example options.

### LINKED ISSUE #123: update loadgen to match .net
Specifically, copy over the pattern for client-based rate limiting, and also implement Ellery's suggestions from this ticket:

https://github.com/momentohq/client-sdk-dotnet/issues/229

Sync up with the .net configuration, plus add a run duration.

- [x] printStatsEveryNRequests
  - Using the existing hard coded default of 1000.
- [x] cacheItemPayloadBytes: 100
- [x] numberOfConcurrentRequests: 50
- [x] maxRequestsPerSecond: 100
- [x] totalNumberOfOperationsToExecute: 500_000
- [x] run duration

schwern:
* maxRequestsPerSecond
  * prevents throttling
  * log the limit so the user realizes it's the script, not the service
* Change to time based instead of request based limits (see momentohq/client-sdk-dotnet#229)
  * <strike>totalNumberOfOperationsToExecute</strike> totalMsToRun
  * printStatsEveryNMs
* Always print stats at the end.

This probably diverges from the other SDK's load generators.

Closes #123

### DIFF
--- a/examples/load-gen.ts
+++ b/examples/load-gen.ts
@@ -104,10 +104,13 @@ class BasicJavaScriptLoadGen {
 
     const delayMillisBetweenRequests =
       (1000.0 * this.numberOfConcurrentRequests) / this.maxRequestsPerSecond;
-    this.logger.info(`Limiting to ${this.maxRequestsPerSecond} tps`);
     this.logger.trace(
       `delayMillisBetweenRequests: ${delayMillisBetweenRequests}`
     );
+
+    this.logger.info(`Limiting to ${this.maxRequestsPerSecond} tps`);
+    this.logger.info(`Running ${this.numberOfConcurrentRequests} concurrent requests`)
+    this.logger.info(`Running for ${this.totalSecondsToRun} seconds`);
 
     const loadGenContext: BasicJavasScriptLoadGenContext = {
       startTime: process.hrtime(),
--- a/examples/load-gen.ts
+++ b/examples/load-gen.ts
@@ -21,7 +21,7 @@ interface BasicJavaScriptLoadGenOptions {
   requestTimeoutMs: number;
   cacheItemPayloadBytes: number;
   numberOfConcurrentRequests: number;
-  printStatsEveryNRequests: number;
+  printStatsEveryNMs: number;
   maxRequestsPerSecond: number;
   totalMsToRun: number;
 }
@@ -56,7 +56,7 @@ class BasicJavaScriptLoadGen {
   private readonly loggerOptions: LoggerOptions;
   private readonly requestTimeoutMs: number;
   private readonly numberOfConcurrentRequests: number;
-  private readonly printStatsEveryNRequests: number;
+  private readonly printStatsEveryNMs: number;
   private readonly maxRequestsPerSecond: number;
   private readonly totalMsToRun: number;
   private readonly cacheValue: string;
@@ -76,7 +76,7 @@ class BasicJavaScriptLoadGen {
     this.authToken = authToken;
     this.requestTimeoutMs = options.requestTimeoutMs;
     this.numberOfConcurrentRequests = options.numberOfConcurrentRequests;
-    this.printStatsEveryNRequests = options.printStatsEveryNRequests;
+    this.printStatsEveryNMs = options.printStatsEveryNMs;
     this.maxRequestsPerSecond = options.maxRequestsPerSecond;
     this.totalMsToRun = options.totalMsToRun;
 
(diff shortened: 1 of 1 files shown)


schwern:
@eaddingtonwhite What do you think of this implementation of momentohq/client-sdk-dotnet#229?

examples/load-gen.ts
```diff
@@ -133,69 +142,93 @@ class BasicJavaScriptLoadGen {
...
-           } (${BasicJavaScriptLoadGen.percentRequests(
-          loadGenContext,
-          loadGenContext.globalSuccessCount
-        )}%) (${BasicJavaScriptLoadGen.tps(
-          loadGenContext,
-          loadGenContext.globalSuccessCount
-        )} tps)
-       unavailable: ${
-         loadGenContext.globalUnavailableCount
+total requests: ${
```
cprice404, line 178: not a huge deal but the indentation here was intentional, so that we were nesting several subsections underneath 'cumulative stats' header in the output

examples/load-gen.ts
```diff
@@ -383,17 +413,21 @@ const loadGeneratorOptions: BasicJavaScriptLoadGenOptions = {
    */
   loggerOptions: {
     /**
-     * Available log levels are TRACE, DEBUG, INFO, WARN, and ERROR.  DEBUG
+     * Available log levels are TRACE, DEBUG, INFO, WARN, and ERROR.  INFO
      * is a reasonable choice for this load generator program.
      */
-    level: LogLevel.DEBUG,
```
cprice404, line 389: Was there something you were trying to filter out by changing this to INFO?  My memory is rusty but I believe that I had set it to DEBUG because there were some retry messages that were useful to see in certain situations.

examples/load-gen.ts
```diff
@@ -383,17 +413,21 @@ const loadGeneratorOptions: BasicJavaScriptLoadGenOptions = {
...
      * Allows you to choose between formatting your log output as JSON (a good
      * choice for production environments) or CONSOLE (a better choice for
      * development environments).
      */
     format: LogFormat.CONSOLE,
   },
+  /** Print some statistics about throughput and latency every time this many
+   *  milliseconds have passed. Default is 5 seconds.
+   */
+  printStatsEveryNMs: 5 * 1000,
```
cprice404, line 430: nit: I don't think anyone would ever want to print these more than once per second, so maybe we change it from `Ms` to `Seconds`?  not a huge deal.

examples/load-gen.ts
```diff
@@ -414,11 +455,10 @@ const loadGeneratorOptions: BasicJavaScriptLoadGenOptions = {
...
   numberOfConcurrentRequests: 10,
   /**
-   * Controls how long the load test will run.  We will execute this many operations
-   * (1 cache 'set' followed immediately by 1 'get') across all of our concurrent
-   * workers before exiting.  Statistics will be logged every 1000 operations.
+   * Controls how long the load test will run, in milliseconds. We will execute operations
+   * for this long and the exit. The default is 60 seconds.
    */
-  totalNumberOfOperationsToExecute: 50_000,
+  totalMsToRun: 60 * 1000,
```
cprice404, line 461: same nit here, i don't think anyone is ever going to target ms granularity for the duration; seconds would probably be fine?

cprice404 (approved):
a few nits but overall looks great, :ty:!

examples/load-gen.ts
schwern: I think that's prettier's fault.

examples/load-gen.ts
schwern: Yes, I added some logs that are quite verbose to test the delay mechanic. There's no other debug logs AFAIK, unless they're coming from something outside the script.

examples/load-gen.ts
schwern: We already had `requestTimeoutMs`. 🤷 Either way is fine, if they want it every 500 ms they can set it to .5. `printStatsEveryNSeconds` is less awkward.

examples/load-gen.ts
schwern: I'm avoiding having some be in seconds and some be in milliseconds. If I change one, I'll change them all.

examples/load-gen.ts
cprice404: they would show up if a retry was triggered.  would it make sense to do your new logs at TRACE?

examples/load-gen.ts
schwern: It doesn't seem that indentation was consistently maintained anyway.

cprice404 (approved):
LGTM!  thank you!

looks like there is a lint failure though.

schwern:
> looks like there is a lint failure though.

argh

cprice404:
FWIW if you run `npm run build` before committing I believe it will catch those.  And `npm run format` will fix them.

cprice404 (approved):
🚢

PR #129, opened 2022-12-15
chore: Add an editorconfig

schwern:
Makes whitespace handling consistent. Honored by many editors, others have a plugin. See https://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+# These are encrypted files. Don't mess with them.
+[*.enc]
+trim_trailing_whitespace = false
+insert_final_newline = false
--- a/examples/load-gen.ts
+++ b/examples/load-gen.ts
@@ -181,7 +181,7 @@ resource exhausted: ${
           loadGenContext.globalRstStreamCount
         )}%)
 
-cumulative set latencies:      
+cumulative set latencies:
 ${BasicJavaScriptLoadGen.outputHistogramSummary(loadGenContext.setLatencies)}
 
 cumulative get latencies:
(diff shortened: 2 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,14 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
```
cprice404, line 9: in theory i have no qualms with an editorconfig, I know that it's becoming fairly common.  But in this repo (and all of our typescript repos) the linting rules are expressed via `eslint` and `prettier`.  Those rules are enforced in CI, recognized by many (most?) editors, and applied via `npm run format`.

So I am concerned about whether having this file here might cause competing opinions where things start getting flipped back and forth by editorconfig battling with eslint/prettier.  And also expressing these rules in 2 places seems like it might require extra care to make sure we keep them in sync.

Maybe you have experience with using both editorconfig and eslint/prettier before?  If so, and my concerns are unfounded, lmk!

cprice404 (commented):
just have a question about editorconfig vs. eslint/prettier.

.editorconfig
schwern: On the contrary, I added it because the linters complained about trailing whitespace.

Think of it this way. EditorConfig does not define the style, it helps editors follow the style. And the style encoded in EditorConfig is so basic the redundancy doesn't matter.

cprice404: okay.  i am willing to trust your judgment on it.  FWIW I have never configured my editor at all, it just understands the eslint and prettier settings; hence my concern that editorconfig might confuse it rather than help.  But you and I are using different editors so I'm fine with trying this.

cprice404 (approved):

.editorconfig
schwern: If it confuses editors, we can remove it.

cprice404: agreed!

PR #131, opened 2022-12-16
feat: update response and error handling

### LINKED ISSUE #93: js sdk improve exception ux

pgautier404:
This commit updates response and error handling to bring them into conformance with [error handling standards and practices](https://github.com/momentohq/standards-and-practices/blob/main/docs/client-specifications/error-handling.md).

I'll update the examples and the README files once this is merged and I can release a new version of the SDK.

Fixes #93


cprice404:
@pgautier404 this looks really nice at first glance!  There's more here than I will be able to take in today.  One thing that would be really helpful is if you could file a second PR that is just a "draft" PR that would show how the example code will change to correspond to these changes.

I'd like to have some combination of @bruuuuuuuce , @eaddingtonwhite , and @poppoerika take a look at the new API and see how it will feel, maybe even pull it down and kick the tires a bit.  But being able to show what the diff of the example code will look like I think will go a long way towards helping folks evaluate UX.

integration/simple-cache-client.test.ts
```diff
@@ -60,87 +67,151 @@ describe('SimpleCacheClient.ts Integration Tests', () => {
     await withCache(momento, cacheName, async () => {
       await momento.set(cacheName, 'key', 'value');
       const res = await momento.get(cacheName, 'key');
-      expect(res.text()).toEqual('value');
+      expect(res instanceof CacheGet.Error).toEqual(false);
+      if (res instanceof CacheGet.Hit) {
+        expect(res.valueString()).toEqual('value');
+      }
```
schwern, line 73: If it isn't a CacheGet.Hit, the test will pass. Same for the rest of the tests.

Add an expectation.

```suggestion
      expect(res instanceof CacheGet.Hit).toEqual(true)
      expect(res.valueString()).toEqual('value');
```

integration/simple-cache-client.test.ts
```diff
@@ -60,87 +67,151 @@ describe('SimpleCacheClient.ts Integration Tests', () => {
...
+  it('should return NotFoundError if deleting a non-existent cache', async () => {
     const cacheName = v4();
     const momento = new SimpleCacheClient(AUTH_TOKEN, 1111);
-    await expect(momento.deleteCache(cacheName)).rejects.toThrow(NotFoundError);
+    const deleteResponse = await momento.deleteCache(cacheName);
+    expect(deleteResponse instanceof DeleteCache.Error).toEqual(true);
+    if (deleteResponse instanceof DeleteCache.Error) {
+      expect(deleteResponse.errorCode()).toEqual(
+        MomentoErrorCode.NOT_FOUND_ERROR
+      );
```
schwern, line 84:
```suggestion
    expect(deleteResponse instanceof DeleteCache.Error).toEqual(true);
    expect(deleteResponse.errorCode()).toEqual(
      MomentoErrorCode.NOT_FOUND_ERROR
    );
```

integration/simple-cache-client.test.ts
```diff
@@ -60,87 +67,151 @@ describe('SimpleCacheClient.ts Integration Tests', () => {
...
-      const caches = (await momento.listCaches()).getCaches();
-      const names = caches.map(c => c.getName());
-      expect(names.includes(cacheName)).toBeTruthy();
+      const listResponse = await momento.listCaches();
+      expect(listResponse instanceof ListCaches.Error).toEqual(false);
+      if (listResponse instanceof ListCaches.Success) {
+        const caches = listResponse.getCaches();
+        const names = caches.map(c => c.getName());
+        expect(names.includes(cacheName)).toBeTruthy();
+      }
```
schwern, line 105:
```suggestion
      expect(listResponse instanceof ListCaches.Error).toEqual(false);
      expect(listResponse instanceof ListCaches.Success).toEqual(true)

      const names = listResponse.getCaches().map(c => c.getName());
      expect(names.includes(cacheName)).toBeTruthy();
```

integration/simple-cache-client.test.ts
```diff
@@ -60,87 +67,151 @@ describe('SimpleCacheClient.ts Integration Tests', () => {
...
+    expect(setResponse instanceof CacheSet.Error).toEqual(false);
+    const getResponse = await momento.get(
+      INTEGRATION_TEST_CACHE_NAME,
+      cacheKey
+    );
+    expect(getResponse instanceof CacheGet.Error).toEqual(false);
+    expect(getResponse instanceof CacheGet.Miss).toEqual(false);
+    if (getResponse instanceof CacheGet.Hit) {
+      expect(getResponse.valueString()).toEqual(cacheValue);
+    }
```
schwern, line 126:
```suggestion
    expect(getResponse instanceof CacheGet.Hit).toEqual(true);
    expect(getResponse.valueString()).toEqual(cacheValue);
```

And so on.

integration/simple-cache-client.test.ts
```diff
@@ -152,42 +223,97 @@ describe('SimpleCacheClient.ts Integration Tests', () => {
...
+    const deleteResponse = await momento.delete(
+      INTEGRATION_TEST_CACHE_NAME,
+      cacheKey
+    );
+    expect(deleteResponse instanceof CacheDelete.Error).toEqual(false);
+    const missResponse = await momento.get(
+      INTEGRATION_TEST_CACHE_NAME,
+      cacheKey
+    );
+    expect(missResponse instanceof CacheGet.Miss).toEqual(true);
```
schwern, line 272: This seems redundant with the previous test.

src/errors/errors.ts
```diff
@@ -1,153 +1,201 @@
...
  */
-export class InvalidArgumentError extends ClientSdkError {
-  constructor(message: string) {
-    super(message);
-    this.name = 'InvalidArgumentError';
+export abstract class SdkError extends Error {
+  public readonly errorCode: MomentoErrorCode;
+  public readonly messageWrapper: string;
+  public readonly transportDetails: MomentoErrorTransportDetails;
+  public readonly stack: string | undefined;
```
schwern, line 58: Why is this undefined instead of null?

src/internal/cache-client.ts
```diff
@@ -208,36 +213,37 @@ export class CacheClient {
...
+                break;
+              case cache.cache_client.ECacheResult.Hit:
+                resolve(new CacheGet.Hit(resp.cache_body));
+                break;
+              case cache.cache_client.ECacheResult.Invalid:
+                resolve(new CacheGet.Error(new UnknownError(resp.message)));
+                break;
+              case cache.cache_client.ECacheResult.Ok:
+                resolve(new CacheGet.Error(new UnknownError(resp.message)));
+                break;
```
schwern, line 228: The two cases seem the same. Use a fall through?
```suggestion
              case cache.cache_client.ECacheResult.Invalid:
              case cache.cache_client.ECacheResult.Ok:
                resolve(new CacheGet.Error(new UnknownError(resp.message)));
                break;
```

schwern (commented):
This is as far in as I got.

src/errors/errors.ts
cprice404: This Is The Way in typescript :)

this is the way

integration/simple-cache-client.test.ts
cprice404: `getCaches` is only available if the type of the response is `Success`.  Putting the code inside of the instanceof guard statement allows the compiler to smart-cast to `Success` and thus makes the `getCaches()` method available when it would not otherwise be available.

My expectation would be that the version of the code you pasted here would not compile, because the compiler will only be able to infer the type of `listResponse` to be `ListCachesResponse`, and that class does not have a method called `getCaches`.

integration/simple-cache-client.test.ts
cprice404: I made a similar comment farther down, but I do not believe this would compile because the parent class does not have a method called `valueString`.

integration/simple-cache-client.test.ts
cprice404: same here.

integration/simple-cache-client.test.ts
pgautier404: The line before this `if` asserts that the response is not an `.Error` type, so it will fail if an error is encountered. However, since this is a `get` test, it will give a false positive if the result is a miss, so I'll update for that.

integration/simple-cache-client.test.ts
pgautier404: If I take the second `expect` out of the `if`, I lose the pattern matching/casting to the subclass. I could add an `else` clause to the `if` statement instead, of using a redundant `instanceof` call, but I don't think that really buys me anything.

integration/simple-cache-client.test.ts
pgautier404: Yes, the `if` is necessary for the cast.

integration/simple-cache-client.test.ts
pgautier404: Good eye, I deleted it.

src/internal/cache-client.ts
pgautier404: Sure, will do.

integration/simple-cache-client.test.ts
cprice404: I do agree with @schwern though that the positive assertion is better than the negative ones.  However unlikely, if we were to ever add more subclasses to the response type, then without the `true` assertion for the `Success` type, the test could be a false positive.

pgautier404: I was totally assuming these responses would always be binary success or failure, but I take your point about there possibly being other response types in the future though. I'll fix these up.

cprice404: 99.9% chance that your assumption is fine :)  but might as well just make the tests 💯

pgautier404:
> @pgautier404 this looks really nice at first glance! There's more here than I will be able to take in today. One thing that would be really helpful is if you could file a second PR that is just a "draft" PR that would show how the example code will change to correspond to these changes.
> 
> I'd like to have some combination of @bruuuuuuuce , @eaddingtonwhite , and @poppoerika take a look at the new API and see how it will feel, maybe even pull it down and kick the tires a bit. But being able to show what the diff of the example code will look like I think will go a long way towards helping folks evaluate UX.

Good idea: https://github.com/momentohq/client-sdk-javascript/pull/132.

integration/simple-cache-client.test.ts
```diff
@@ -61,13 +61,13 @@ async function withCache(
...
 describe('SimpleCacheClient.ts Integration Tests', () => {
-  it('should create and delete a cache', async () => {
+  it('should create and delete a cache, set and get a value', async () => {
     const cacheName = v4();
     const momento = new SimpleCacheClient(AUTH_TOKEN, 1111);
     await withCache(momento, cacheName, async () => {
       await momento.set(cacheName, 'key', 'value');
       const res = await momento.get(cacheName, 'key');
-      expect(res instanceof CacheGet.Error).toEqual(false);
+      expect(res instanceof CacheGet.Hit).toEqual(true);
```
schwern, line 70: Note there is [`toBeInstanceOf`](https://jestjs.io/docs/expect#tobeinstanceofclass). More concise, easier to read.

```suggestion
      expect(res).toBeInstanceOf(CacheGet.Hit)
```

src/messages/responses/create-signing-key/create-signing-key.ts
```diff
@@ -34,3 +37,27 @@ export class CreateSigningKeyResponse {
...
+  }
+
+  public errorCode(): MomentoErrorCode {
+    return this._innerException.errorCode;
+  }
+
+  public override toString(): string {
+    return super.toString() + ': ' + this.message();
+  }
+}
```
schwern, line 63: This appears to be a duplicate of CreateCacheResponse's Error. Can this be DRYed up with a [mixin](https://www.typescriptlang.org/docs/handbook/mixins.html)?

src/messages/responses/list-caches/list-caches.ts
```diff
@@ -0,0 +1,54 @@
...
+    );
+  }
+
+  public getNextToken() {
+    return this.nextToken;
+  }
+
+  public getCaches() {
+    return this.caches;
+  }
```
schwern, line 24: The Error classes use an public attribute() / private _attribute` pattern. This uses a public getAttribute() / private attribute` pattern. Would it be better to make them consistent?

src/messages/responses/list-caches/list-caches.ts
```diff
@@ -0,0 +1,54 @@
...
+export class Success extends ListCachesResponse {
+  private readonly nextToken: string | null;
+  private readonly caches: CacheInfo[];
+  constructor(result?: control.control_client._ListCachesResponse) {
+    super();
+    this.nextToken = result?.next_token || null;
+    this.caches = [];
+    result?.cache.forEach(cache =>
+      this.caches.push(new CacheInfo(cache.cache_name))
+    );
```
schwern, line 15: Looks like a map?

src/messages/responses/set/cache-set.ts
```diff
@@ -0,0 +1,46 @@
...
+    this.value = value;
+  }
+
+  public valueString(): string {
+    return TEXT_DECODER.decode(this.value);
+  }
+
+  public valueBytes(): Uint8Array {
+    return this.value;
+  }
```
schwern, line 21: Mixin to share with CacheGetResponse?

src/messages/responses/create-signing-key/create-signing-key.ts
pgautier404: I'll add an issue to look at this as a followup. I looked into mixins for this but spent too much time wrestling with it to justify adding it in this PR.

src/messages/responses/list-caches/list-caches.ts
pgautier404: Yes, good call. I'll update the errors.

src/messages/responses/list-caches/list-caches.ts
pgautier404: Yeah, it does. This was copied and pasted from the old response class. Kinda seems six one half dozen the other to me but if you have a compelling reason to change it I'll do it.

integration/simple-cache-client.test.ts
pgautier404: Nice, thanks.

src/messages/responses/set/cache-set.ts
pgautier404: See above regarding mixins . . . I'll look into that separately.

src/messages/responses/list-caches/list-caches.ts
schwern: Takes more time to figure out what's going on. And everywhere else uses map.

integration/simple-cache-client.test.ts
```diff
@@ -19,24 +25,32 @@ const deleteCacheIfExists = async (
...
   cacheName: string
 ) => {
-  try {
-    await momento.deleteCache(cacheName);
-  } catch (e) {
-    if (!(e instanceof NotFoundError)) {
-      throw e;
+  const deleteResponse = await momento.deleteCache(cacheName);
+  if (deleteResponse instanceof DeleteCache.Error) {
+    deleteResponse.toString();
```
schwern, line 30: This seems to do nothing.

schwern (approved):

src/messages/responses/list-caches/list-caches.ts
pgautier404: ok, i changed it for you.

schwern (approved):

integration/simple-cache-client.test.ts
```diff
@@ -19,24 +25,31 @@ const deleteCacheIfExists = async (
...
   cacheName: string
 ) => {
-  try {
-    await momento.deleteCache(cacheName);
-  } catch (e) {
-    if (!(e instanceof NotFoundError)) {
-      throw e;
+  const deleteResponse = await momento.deleteCache(cacheName);
+  if (deleteResponse instanceof DeleteCache.Error) {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
```
bruuuuuuuce, line 30: why does it think that `deleteResponse.errorCode()` is potentially unsafe?

src/errors/error-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+    return this._innerException.errorCode();
+  }
+
+  public toString(): string {
+    return this.message();
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function applyMixins(derivedCtor: any, constructors: any[]) {
```
bruuuuuuuce, line 23: It is not immediately clear what this function is doing, might be good to add a comment about exactly what its purpose is. Also, could this function be typed with `export function applyMixins(derivedCtor: ErrorBody, constructors: ErrorBody[])`? Just looking at where its used it looks like it could potentially be typed this way.

src/errors/errors.ts
```diff
@@ -1,153 +1,209 @@
...
+  // Request rate exceeded the limits for the account
+  LIMIT_EXCEEDED_ERROR = 'LIMIT_EXCEEDED_ERROR',
+  // Request was invalid
+  BAD_REQUEST_ERROR = 'BAD_REQUEST_ERROR',
+  // Client's configured timeout was exceeded
+  TIMEOUT_ERROR = 'TIMEOUT_ERROR',
+  // Server was unable to handle the request
+  SERVER_UNAVAILABLE = 'SERVER_UNAVAILABLE',
+  // A client resource (most likely memory) was exhausted
+  CLIENT_RESOURCE_EXHAUSTED = 'CLIENT_RESOURCE_EXHAUSTED',
```
bruuuuuuuce, line 27: What is the difference between `CLIENT_RESOURCE_EXHAUSTED` and `LIMIT_EXCEEDED_ERROR`?

src/errors/errors.ts
```diff
@@ -1,153 +1,209 @@
...
- * Error when calls are throttled due to request limit rate
+ * Insufficient permissions to perform an operation on Cache Service
  */
-export class LimitExceededError extends CacheServiceError {
-  constructor(message: string) {
-    super(message);
-    this.name = 'LimitExceededError';
-  }
+export class PermissionError extends SdkError {
+  _errorCode = MomentoErrorCode.PERMISSION_ERROR;
```
bruuuuuuuce, line 171: Just curious, do we currently ever throw this error? I have never seen it before

src/internal/cache-client.ts
```diff
@@ -156,11 +156,12 @@ export class CacheClient {
...
     cacheName: string,
     key: Uint8Array
-  ): Promise<DeleteResponse> {
+  ): Promise<CacheDelete.Response> {
     const request = new cache.cache_client._DeleteRequest({
       cache_key: key,
     });
     const metadata = this.createMetadata(cacheName);
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     return await new Promise((resolve, reject) => {
```
bruuuuuuuce, line 165: nit: can remove `reject` and then you don't need to disable any eslint rules here

src/internal/cache-client.ts
```diff
@@ -182,23 +183,24 @@ export class CacheClient {
...
     key: Uint8Array
-  ): Promise<GetResponse> {
+  ): Promise<CacheGet.Response> {
     const request = new cache.cache_client._GetRequest({
       cache_key: key,
     });
     const metadata = this.createMetadata(cacheName);
 
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     return await new Promise((resolve, reject) => {
```
bruuuuuuuce, line 204: ditto

src/internal/cache-client.ts
```diff
@@ -208,36 +210,35 @@ export class CacheClient {
...
+            switch (resp.result) {
+              case cache.cache_client.ECacheResult.Miss:
+                resolve(new CacheGet.Miss());
+                break;
+              case cache.cache_client.ECacheResult.Hit:
+                resolve(new CacheGet.Hit(resp.cache_body));
+                break;
+              case cache.cache_client.ECacheResult.Invalid:
+              case cache.cache_client.ECacheResult.Ok:
+                resolve(new CacheGet.Error(new UnknownError(resp.message)));
```
bruuuuuuuce, line 222: this reads a little weird to me. If the result is `OK` we return an error? I am assuming this is because get should only return a hit or a miss?

bruuuuuuuce (commented):
A few nits/questions but overall looks great!

integration/simple-cache-client.test.ts
```diff
@@ -19,24 +25,31 @@ const deleteCacheIfExists = async (
...
+    if (deleteResponse.errorCode() !== MomentoErrorCode.NOT_FOUND_ERROR) {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
```
eaddingtonwhite, line 32: why do we need this eslint disabled line on 32 here but not on 43  or 51?

eaddingtonwhite: O I guess this is test but just want to double check make sure nothing weird happening here vs other API calls

src/errors/error-utils.ts
eaddingtonwhite: Agree maybe also just some JS Doc comments on what its doing and why we need this just for another dev coming in and looking at this?

src/errors/errors.ts
```diff
@@ -1,153 +1,209 @@
...
  */
-export class AuthenticationError extends CacheServiceError {
-  constructor(message: string) {
-    super(message);
-    this.name = 'AuthenticationError';
-  }
+export class LimitExceededError extends SdkError {
+  _errorCode = MomentoErrorCode.LIMIT_EXCEEDED_ERROR;
+  _messageWrapper =
+    'Request rate exceeded the limits for this account.  To resolve this error, reduce your request rate, or contact us at support@momentohq.com to request a limit increase';
```
eaddingtonwhite, line 154: Limit errors can be for bandwidth and object size as well. Maybe reword this message a bit to be more generic vs just 'rate'?

src/internal/control-client.ts
```diff
@@ -58,75 +54,85 @@ export class ControlClient {
...
-  public async deleteCache(name: string): Promise<DeleteCacheResponse> {
+  public async deleteCache(name: string): Promise<DeleteCache.Response> {
+    try {
+      this.validateCacheName(name);
+    } catch (err) {
+      if (err instanceof SdkError) {
+        return new DeleteCache.Error(err);
+      } else if (err instanceof Error) {
+        return new DeleteCache.Error(new UnknownError(err.message));
+      }
```
eaddingtonwhite, line 100: seen this little catch and wrap or not logic a bit now. Could maybe dry up with a basic utility so could do something like:
```js
    } catch (err) {
      return wrapErrorIfSDKError(DeleteCache.Error, err)
    }
```

integration/simple-cache-client.test.ts
pgautier404: I honestly don't know . . . let me know if you'd like to look into it with me.

src/errors/error-utils.ts
pgautier404: Fair point. I'll add some explanation.

src/messages/responses/list-caches/list-caches.ts
cprice404: +1 to map for this, thanks for changing it

src/errors/error-utils.ts
pgautier404: > Also, could this function be typed with `export function applyMixins(derivedCtor: ErrorBody, constructors: ErrorBody[])`? Just looking at where its used it looks like it could potentially be typed this way.

The class passed into `derivedCtor` is not of type `ErrorBody` and the `constructors` argument could be passed anything. This is intended to be a general purpose function for using [mixing](https://www.typescriptlang.org/docs/handbook/mixins.html#alternative-pattern) that we happen to be using for a single specific purpose for now.

src/errors/errors.ts
pgautier404: The former is something like your laptop running out of memory. The latter is something like you've exceeded your request rate limit.

cprice404: client resources are things like memory on the client side

src/internal/cache-client.ts
```diff
@@ -119,13 +118,14 @@ export class CacheClient {
...
     ttl: number
-  ): Promise<SetResponse> {
+  ): Promise<CacheSet.Response> {
     const request = new cache.cache_client._SetRequest({
       cache_body: value,
       cache_key: key,
       ttl_milliseconds: ttl * 1000,
     });
     const metadata = this.createMetadata(cacheName);
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
cprice404, line 128: matt's comment below applies here too i think

src/internal/cache-client.ts
cprice404: yeah the wire format is kind of weird on this one, but i think it would be a bug for the server to return anything but Hit here.  this is a little different from the .net sdk though, which does `if Miss ... else return Hit`.

this one feels probably more correct to me though.

integration/simple-cache-client.test.ts
cprice404: I would like to look into this, I can find some time today.  Not necessarily a blocker for the PR though.

cprice404:
@eaddingtonwhite is there anything here you think is a blocker or should we fast-follow the rest?

src/errors/errors.ts
pgautier404: I imagine so. There is a corresponding gRPC status that gets mapped to this error type.

eaddingtonwhite (approved):
Great work on this @pgautier404!! 🏆  I think GTG can merge with this as is now and then fast follow on some of the tactical feedback still open in here. Overall looks great very excited to get these changes out!

src/errors/errors.ts
schwern: @eaddingtonwhite This error message is copied from the [Standards and Practices Error Handling doc](https://github.com/momentohq/standards-and-practices/blob/main/docs/client-specifications/error-handling.md) so it should be updated there as well.

src/internal/cache-client.ts
pgautier404: Nice, thanks.

PR #132, opened 2022-12-19
chore: update examples for latest SDK version

pgautier404:
Updates examples to work with changes in https://github.com/momentohq/client-sdk-javascript/pull/131.


examples/presigned-url-example.ts
```diff
@@ -47,12 +45,16 @@ const main = async () => {
...
+      signingKey = signingKeyResponse.getKey();
+      process.env.MOMENTO_ENDPOINT = signingKeyResponse.getEndpoint();
+    } else if (signingKeyResponse instanceof CreateSigningKey.Error) {
+      throw signingKeyResponse.innerException();
+    }
   }
   const endpoint = requireEnvVar('MOMENTO_ENDPOINT');
 
-  const signer = await MomentoSigner.init(signingKey);
+  const signer = await MomentoSigner.init(signingKey as string);
```
cprice404, line 57: i'm not sure this should be necessary?  Can't you give it the type hint for the signingKey var on the line where it is initialized via `let`?  Or, I guess, I'm not clear why this cast would be necessary now when it wasn't before?

cprice404:
Thanks @pgautier404 !

@eaddingtonwhite @bruuuuuuuce we would really appreciate your feedback on the direction that this UX is heading before we move too much farther along.

examples/presigned-url-example.ts
pgautier404: @cprice404 The original `let` is inferred as type `string | undefined` because it comes from an environment variable. Because it's now reassigned to a string in a pattern matching `if` that has an `else if` clause, the compiler can no longer be sure that it will end up with a string value from `getKey()` (despite the fact that the else clause throws an exception). The cast to string allows me to leave the rest of that code outside of the if statement.

cprice404: okay.

this is a nit so it's up to you if you want to change it but I consider both the `let` and the `as` to be code smells here.  Probably what I would do is extract a function called `getSigningKey`, and have its return type be `string`.

Then you can use a const to grab the env var, check it, and return the string if it is present.  If not then you can do the if/else and ensure that all code paths return a string or else they throw.

pgautier404: cool, that's fair. i'll clean it up.

pgautier404:
Superseded by https://github.com/momentohq/client-sdk-javascript/pull/134

PR #134, opened 2023-01-04
chore: update examples for SDK v0.16.0

pgautier404:


schwern (dismissed):
Ran all the examples, they work. 👍 

When I run index.ts, there's a mysterious JSON log message and I can't find where it's coming from.

```
$ npm run example

> momento-javascript-example@1.0.0 example
> tsc && node dist/index.js

{"level":30,"time":1672793112957,"pid":25498,"hostname":"Windhund.local","name":"ControlClient","msg":"Creating cache: cache"}
cache already exists
Listing caches:
js-loadgen
cache
schwern-python-test-cache
Storing key=key, value=value, ttl=10
Key stored successfully with value value
cache hit: value
Deleted key from cache
success!!
```

examples/presigned-url-example.ts
```diff
@@ -61,6 +44,29 @@ const main = async () => {
...
+  // created and revoked via the Momento CLI.
+  console.log(
+    'Did not find signing key in environment, creating new signing key'
+  );
+  const signingKeyResponse = await momento.createSigningKey(60 * 24); // valid for 24 hours
+  if (signingKeyResponse instanceof CreateSigningKey.Error) {
+    throw signingKeyResponse.innerException();
+  }
+  process.env.MOMENTO_ENDPOINT = (
+    signingKeyResponse as CreateSigningKey.Success
```
cprice404, line 65: I'd really like to avoid casts in the example code, can we just put this in another branch of the `if` so that we get the smart cast?

cprice404 (changes requested):

examples/presigned-url-example.ts
```diff
@@ -58,13 +58,14 @@ async function getSigningKey(): Promise<string> {
     'Did not find signing key in environment, creating new signing key'
   );
   const signingKeyResponse = await momento.createSigningKey(60 * 24); // valid for 24 hours
+  let returnVal = '';
```
cprice404, line 61: i'm sorry to keep nitpicking about this.  You can avoid this mutable var and the unfortunate code path related to it if you just add a final `else` clause to the condition and then throw from it, I think that's better for example code.

cprice404 (changes requested):

cprice404 (approved):
:ty:

PR #138, opened 2023-01-04
chore: update README files

pgautier404:


eaddingtonwhite (approved):

README.template.md
```diff
@@ -32,7 +32,44 @@ Here is a quickstart you can use in your own project:
...
+if (getResponse instanceof CacheGet.Error) {
+    if (getResponse.errorCode() == MomentoErrorCode.TIMEOUT_ERROR) {
+       // this would represent a client-side timeout, and you could fall back to your original data source
+    }
+}
+```
+
+Note that, outside of `SimpleCacheClient` responses, exceptions can occur and should be handled as usual. For example,
+trying to instantiate a `SimpleCacheClient` with an invalid authentication token will result in an
+`IllegalArgumentException` being thrown.
```
cprice404, line 72: ❤️

examples/README.md
```diff
@@ -45,29 +45,59 @@ npm install @gomomento/sdk
...
-await momento.set(CACHE_NAME, 'key', 'value');
-const res = await momento.get(CACHE_NAME, 'key');
-console.log('result: ', res.text());
-
-// sets key with ttl of 5 seconds
-await momento.set(CACHE_NAME, 'key2', 'value2', 5);
-
-// permanently deletes cache
-await momento.deleteCache(CACHE_NAME);
+import {
```
cprice404, line 48: Thanks for updating this.  We should definitely merge these changes.  but...

This approach has that issue where if the API changes we won't necessarily know that the README code is broken.  I know we solved this for the main READMEs via the template.  I don't remember what we did in the other SDKs to deal with this issue in the example READMEs?

cprice404 (approved):
🚢 but let's maybe have a convo about how to keep code in the example readmes from ending up broken, and see what we are doing in the other repos.

examples/README.md
pgautier404: We don't have a standard approach. The .NET example README only contains small code snippets and the PHP version points to the various examples without including any code at all. I'll make a ticket for defining and implementing a standard we can apply across SDKs.

pgautier404: https://github.com/momentohq/dev-eco-issue-tracker/issues/169

PR #141, opened 2023-01-07
feat: use standard configuration classes

pgautier404:


src/config/configurations.ts
```diff
@@ -0,0 +1,29 @@
...
+  ITransportStrategy,
+  StaticGrpcConfiguration,
+  StaticTransportStrategy,
+} from './transport/transport-strategy';
+import {IGrpcConfiguration} from './transport/grpc-configuration';
+
+export class Laptop extends Configuration {
+  constructor(transportStrategy: ITransportStrategy, maxIdleMillis: number) {
+    super(transportStrategy, maxIdleMillis);
+  }
```
schwern, line 12: Redundant, it's already inherited.

src/config/transport/grpc-configuration.ts
```diff
@@ -0,0 +1,6 @@
+export interface IGrpcConfiguration {
+  getDeadlineMilliseconds(): number;
+  withDeadlineMilliseconds(deadlineMilliseconds: number): IGrpcConfiguration;
+  getMaxSessionMemory(): number;
+  withMaxSessionMemory(maxSessionMemory: number): IGrpcConfiguration;
```
schwern, line 5: Never used.

src/config/transport/transport-strategy.ts
```diff
@@ -0,0 +1,85 @@
+import {IGrpcConfiguration} from './grpc-configuration';
+
+export interface ITransportStrategy {
+  getMaxConcurrentRequests(): number | null;
```
schwern, line 4: Never used.

src/config/transport/transport-strategy.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  getGrpcConfig(): IGrpcConfiguration;
+
+  // TODO: for use in middleware
+  withMaxConcurrentRequests(maxConcurrentRequests: number): ITransportStrategy;
+
+  withGrpcConfig(grpcConfig: IGrpcConfiguration): ITransportStrategy;
```
schwern, line 11: Never used.

src/config/transport/transport-strategy.ts
```diff
@@ -0,0 +1,85 @@
...
+  withDeadlineMilliseconds(
+    deadlineMilliseconds: number
+  ): StaticGrpcConfiguration {
+    return new StaticGrpcConfiguration(
+      deadlineMilliseconds,
+      this.maxSessionMemory
+    );
+  }
+
+  withMaxSessionMemory(maxSessionMemory: number): StaticGrpcConfiguration {
```
schwern, line 41: Never used.

src/config/transport/transport-strategy.ts
```diff
@@ -0,0 +1,85 @@
...
+    maxConcurrentRequests: number
+  ): StaticTransportStrategy {
+    return new StaticTransportStrategy(maxConcurrentRequests, this.grpcConfig);
+  }
+
+  withGrpcConfig(grpcConfig: IGrpcConfiguration): StaticTransportStrategy {
+    return new StaticTransportStrategy(this.maxConcurrentRequests, grpcConfig);
+  }
+
+  withClientTimeout(clientTimeout: number): StaticTransportStrategy {
```
schwern, line 79: It's a little confusing that "clientTimeout" sets "deadlineMilliseconds".

Should deadlineMilliseconds be timeoutMilliseconds? Should clientTimeout be clientTimeoutMilliseconds?

src/internal/cache-client.ts
```diff
@@ -49,42 +50,49 @@ export class CacheClient {
...
+    this.authProvider = props.authProvider;
     this.logger = getLogger(this);
-    this.validateRequestTimeout(props.requestTimeoutMs);
+    const grpcConfig = this.configuration
+      .getTransportStrategy()
+      .getGrpcConfig();
+
+    this.requestTimeoutMs =
+      grpcConfig.getDeadlineMilliseconds() ||
+      CacheClient.DEFAULT_REQUEST_TIMEOUT_MS;
```
schwern, line 62: Should grpc defaults live as constants inside the cache client?

Should users of IGrpcConfiguration have to worry about providing a default if getDeadlineMilliseconds is 0?

Should IGrpcConfiguration allow getDeadlineMilliseconds to be 0?

schwern (changes requested):
* I'd like to see unit tests.

* Rather than a "with" method for every possible config variable, would a single `clone` method be better?

```
// Uses grpcConfig.maxSessionMemory
const copy = grpcConfig.clone(deadlineMilliseconds: 23)
```

It seems Typescript might make this difficult. 🤔

cprice404:
Thanks for the review @schwern !  I told Pete that I would take this over and fast-follow with any changes that came out of the PR review unless there were show-stoppers, so I am planning to review your feedback more closely, create an issue to capture the tasks, and then knock those out in follow-up PRs.  Just haven't quite had time to sink my teeth in yet.  Apologies for the delay.

cprice404:
> * I'd like to see unit tests.
> 
>     * Rather than a "with" method for every possible config variable, would a single `clone` method be better?
> 
> 
> ```
> // Uses grpcConfig.maxSessionMemory
> const copy = grpcConfig.clone(deadlineMilliseconds: 23)
> ```
> 
> It seems Typescript might make this difficult. 🤔

The general idea for the Config objects is that they would always have 3 `with` copy constructors: RetryStrategy, Middlewares, and TransportStrategy.  

https://github.com/momentohq/standards-and-practices/blob/main/docs/client-specifications/configuration.md#configuration-interface

We will occasionally promote a few commonly used features like `ClientTimeout` up to have their own `with`s at this top level even though they are technically part of one of the nested strategies (in this case, transport strategy), as a convenience so that users don't have to deal with as much ceremony in order to override them.

In this case I will probably move this idle timeout setting down into the transport strategy so that might eliminate a bit of what you are getting at here.

I do think the `clone` idea is a cool one, as a possible supplement to the explicit copy constructors.  In TS we would just need to create an interface for the arguments for it and then it will feel like the keyword args that you're after.  I'll file a ticket to capture this idea.

src/auth/credential-provider.ts
```diff
@@ -0,0 +1,63 @@
...
+export interface ICredentialProvider {
+  getAuthToken(): string;
+
+  getControlEndpoint(): string;
+
+  getCacheEndpoint(): string;
+
+  getTrustedControlEndpointCertificateName(): string | null;
+
+  getTrustedCacheEndpointCertificateName(): string | null;
```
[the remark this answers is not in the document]
cprice404: good catch, I will look into wiring those up, or I will remove them for now and file a ticket to come back and add them when we have a customer ask.

src/auth/credential-provider.ts
```diff
@@ -0,0 +1,63 @@
...
+  private readonly trustedControlEndpointCertificateName: string | null;
+  private readonly trustedCacheEndpointCertificateName: string | null;
+  constructor(
+    envVariableName: string,
+    controlEndpoint?: string | null,
+    cacheEndpoint?: string | null,
+    trustedControlEndpointCertificateName?: string | null,
+    trustedCacheEndpointCertificateName?: string | null
+  ) {
+    const authToken = process.env[envVariableName];
```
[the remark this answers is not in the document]
cprice404: good question.  this is just the first implementation of the interface.  We will add more over time.  I will probably add a `String` one in my follow-up PR, like we have in some of the other repos.

src/config/configuration.ts
```diff
@@ -0,0 +1,44 @@
...
+    return this.maxIdleMillis;
+  }
+
+  withTransportStrategy(transportStrategy: ITransportStrategy): IConfiguration {
+    return new Configuration(transportStrategy, this.maxIdleMillis);
+  }
+
+  withMaxIdleMillis(maxIdleMillis: number) {
+    return new Configuration(this.transportStrategy, maxIdleMillis);
+  }
```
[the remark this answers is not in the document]
cprice404: fair callout, i'll add some tests for these in my follow-up PR.

src/config/configurations.ts
cprice404: are you saying that in JS/TS a child class will inherit non-default constructors from the parent class?  If so, that seems surprising to me!  I don't know if I've seen that in a language before.

Even if that's the case I might still be tempted to keep this here just so that it's really clear and explicit when you're looking at this file, but I will play with this in my follow-up PR.

src/config/transport/grpc-configuration.ts
cprice404: will add test coverage.

src/config/transport/transport-strategy.ts
cprice404: good call, we don't actually have a way to limit these in JS for now, so I'll peel these out for the time being.

src/config/transport/transport-strategy.ts
cprice404: will add tests

src/config/transport/transport-strategy.ts
cprice404: will add tests

src/config/transport/transport-strategy.ts
cprice404: this is the layer at which we are trying to abstract away grpc implementation details.

in grpc-speak the correct name for this field is `*Deadline*`.

In customer-facing APIs we usually don't want them to have to think about gRPC unless they really dig down into the guts, so we expose it as `ClientTimeout`.  This would theoretically let us use this config item for other transports as well.

I think you are correct that `clientTimeout` should be `clientTimeoutMilliseconds` though, unless there is a sane duration object we can use in JS-land.  I will change this in one of those two ways.

src/internal/cache-client.ts
cprice404: > Should grpc defaults live as constants inside the cache client?
> 
> Should users of IGrpcConfiguration have to worry about providing a default if getDeadlineMilliseconds is 0?
> 
> Should IGrpcConfiguration allow getDeadlineMilliseconds to be 0?

These are all fair callouts.  I need to look at some of the other SDKs to see what we're doing for the zero case.  And if the semantics of JS/TS are that `(0 || x) == x` (i.e., 0 is not truth-y) then I think we probably need to make that field nullable and do a better check on it.  I will add that to my list.

cprice404:
@schwern @pgautier404 I created this ticket to capture fast follow items; I think I should be able to pick most of them off this week:

https://github.com/momentohq/client-sdk-javascript/issues/146

Please let me know if that list looks sufficient and if so I will approve this PR and we can merge it, and I'll start filing additional PRs to hit that remaining list ASAP.

cprice404 (approved):

src/config/configurations.ts
schwern: Inheriting a constructor seems normal to me, I'd be surprised if it didn't.

> Classes may extend from a base class. A derived class has all the properties and methods of its base class, and also define additional members.

Tests pass if you delete it.

src/config/transport/transport-strategy.ts
schwern: That's fair if the idea is to keep GRPC terms at the GRPC layer.

There doesn't seem to be a built-in Javascript nor Typescript duration object. I found [this popular module](https://www.npmjs.com/package/duration), but it doesn't let you construct Durations out of minutes and seconds. 🤷

We start building up a style guide with things like how to handle time durations in various languages and how to name args that take them.

schwern:
@cprice404 It's not immediately obvious to me if `withX` does a clone with the new value, or if it's a setter that returns itself (a common setter pattern). Maybe it's a TypeScript convention? `cloneWithX` would be obvious. Docs might be enough to solve the issue.

I messed around with using ["destructruing function parameters"](https://www.typescriptlang.org/docs/handbook/variable-declarations.html#function-declarations) to implement this 🤢. It can be done, but it makes Perl's argument processing look positively streamlined. Might be best to avoid it.

PR #148, opened 2023-01-11
chore: fix json5 vulnerability

schwern:
npm audit caught a security vulnerability.
https://vulert.com/vulnerability/npm-json5-47903

Should dependabot have caught this?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4923,9 +4923,9 @@
       "dev": true
     },
     "node_modules/json5": {
-      "version": "2.2.1",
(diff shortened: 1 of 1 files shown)


cprice404 (approved):

cprice404:
> npm audit caught a security vulnerability. https://vulert.com/vulnerability/npm-json5-47903
> 
> Should dependabot have caught this?

I think we are currently only running dependabot checks on the examples dir, and only when a new version of the actual SDK is released.  We should look into expanding that in the future.

PR #149, opened 2023-01-12
feat: Add ListFetch and ListPushFront

schwern:
This adds two list data structures, enough to establish a pattern and test them.

- ListFetch
- ListPushFront

Also:
- Make `CacheGet.Hit.toString()` consistent when the value is truncated.

### DIFF
--- a/src/internal/cache-client.ts
+++ b/src/internal/cache-client.ts
@@ -12,6 +12,8 @@ import * as CacheGet from '../messages/responses/cache-get';
 import * as CacheSet from '../messages/responses/cache-set';
 import * as CacheDelete from '../messages/responses/cache-delete';
 import * as CacheSetFetch from '../messages/responses/cache-set-fetch';
+import * as CacheListFetch from '../messages/responses/cache-list-fetch';
+import * as CacheListPushFront from '../messages/responses/cache-list-push-front';
 import {version} from '../../package.json';
 import {getLogger, Logger} from '../utils/logging';
 import {IdleGrpcClientWrapper} from '../grpc/idle-grpc-client-wrapper';
@@ -21,11 +23,13 @@ import {
   ensureValidKey,
   ensureValidSetRequest,
   validateCacheName,
+  validateListName,
   validateSetName,
 } from '../utils/validators';
 import {CredentialProvider} from '../auth/credential-provider';
 import {Configuration} from '../config/configuration';
 import {SimpleCacheClientProps} from '../simple-cache-client-props';
+import {CollectionTtl} from '../utils/collection-ttl';
 
 export class CacheClient {
   private readonly clientWrapper: GrpcClientWrapper<grpcCache.ScsClient>;
@@ -289,6 +293,120 @@ export class CacheClient {
             }
           } else {
             resolve(new CacheGet.Error(cacheServiceErrorMapper(err)));
+          }
+        }
+      );
+    });
+  }
+
+  public async listFetch(
+    cacheName: string,
+    listName: string
+  ): Promise<CacheListFetch.Response> {
+    try {
+      validateCacheName(cacheName);
+      validateListName(listName);
+    } catch (err) {
+      return new CacheListFetch.Error(normalizeSdkError(err as Error));
+    }
+    this.logger.trace(`Issuing 'listFetch' request; listName: ${listName}`);
+    const result = await this.sendListFetch(cacheName, this.convert(listName));
+    this.logger.trace(`'listFetch' request result: ${result.toString()}`);
+    return result;
+  }
+
+  private async sendListFetch(
+    cacheName: string,
+    listName: Uint8Array
+  ): Promise<CacheListFetch.Response> {
+    const request = new grpcCache._ListFetchRequest({
+      list_name: listName,
+    });
+    const metadata = this.createMetadata(cacheName);
+
+    return await new Promise(resolve => {
+      this.clientWrapper.getClient().ListFetch(
+        request,
+        metadata,
+        {
+          interceptors: this.interceptors,
+        },
+        (err, resp) => {
+          if (resp?.missing) {
+            resolve(new CacheListFetch.Miss());
+          } else if (resp?.found) {
+            resolve(new CacheListFetch.Hit(resp.found.values));
+          } else {
+            resolve(new CacheListFetch.Error(cacheServiceErrorMapper(err)));
+          }
+        }
+      );
+    });
+  }
+
+  public async listPushFront(
+    cacheName: string,
+    listName: string,
+    value: string | Uint8Array,
+    ttl: CollectionTtl = CollectionTtl.fromCacheTtl(),
+    truncateBackToSize?: number
+  ): Promise<CacheListFetch.Response> {
+    try {
+      validateCacheName(cacheName);
+      validateListName(listName);
+    } catch (err) {
+      return new CacheListFetch.Error(normalizeSdkError(err as Error));
+    }
+
+    this.logger.trace(
+      `Issuing 'listPushFront' request; listName: ${listName}, value length: ${
+        value.length
+      }, ${ttl.toString()}, truncateBackToSize: ${
+        truncateBackToSize?.toString() ?? 'null'
+      }`
+    );
+
+    const result = await this.sendListPushFront(
+      cacheName,
+      this.convert(listName),
+      this.convert(value),
+      ttl.ttlMilliseconds() || this.defaultTtlSeconds * 1000,
+      ttl.refreshTtl(),
+      truncateBackToSize
+    );
+    this.logger.trace(`'listPushFront' request result: ${result.toString()}`);
+    return result;
+  }
+
+  private async sendListPushFront(
+    cacheName: string,
+    listName: Uint8Array,
+    value: Uint8Array,
+    ttlMilliseconds: number,
+    refreshTtl: boolean,
+    truncateBackToSize?: number
+  ): Promise<CacheListPushFront.Response> {
+    const request = new grpcCache._ListPushFrontRequest({
+      list_name: listName,
+      value: value,
+      ttl_milliseconds: ttlMilliseconds,
+      refresh_ttl: refreshTtl,
+      truncate_back_to_size: truncateBackToSize,
+    });
+    const metadata = this.createMetadata(cacheName);
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    return await new Promise(resolve => {
+      this.clientWrapper.getClient().ListPushFront(
+        request,
+        metadata,
+        {
+          interceptors: this.interceptors,
+        },
+        (err, resp) => {
+          if (resp) {
+            resolve(new CacheListPushFront.Success(resp.list_length));
+          } else {
+            resolve(new CacheListPushFront.Error(cacheServiceErrorMapper(err)));
           }
         }
       );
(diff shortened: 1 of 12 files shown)


src/simple-cache-client.ts
```diff
@@ -147,6 +148,33 @@ export class SimpleCacheClient {
     return await client.delete(cacheName, key);
   }
 
+  public async listFetch(
+    cacheName: string,
+    listName: string | Uint8Array
```
poppoerika, line 153: I don't think we let users to pass dictionary/set/list names in bytes?

src/internal/cache-client.ts
```diff
@@ -258,6 +261,127 @@ export class CacheClient {
...
+      );
+    });
+  }
+
+  public async listPushFront(
+    cacheName: string,
+    listName: string | Uint8Array,
+    value: string | Uint8Array,
+    refreshTtl: boolean,
+    ttl?: number,
```
poppoerika, line 320: Do we want to implement `CollectionTtl`? 
https://github.com/momentohq/client-sdk-dotnet-incubating/blob/main/src/Momento.Sdk.Incubating/Requests/CollectionTtl.cs

src/messages/responses/cache-list-fetch.ts
```diff
@@ -0,0 +1,33 @@
+import {SdkError} from '../../errors/errors';
+import {ResponseBase} from './response-base';
+import {applyMixins, ErrorBody} from '../../errors/error-utils';
+
+export abstract class Response extends ResponseBase {}
+
+export class Hit extends Response {
+  private readonly _values: Uint8Array[];
```
poppoerika, line 8: I know this is wip PR but are you planning to add an array of string values?

src/internal/cache-client.ts
```diff
@@ -258,6 +261,127 @@ export class CacheClient {
...
+          if (resp) {
+            if (resp.missing) {
+              resolve(new CacheListFetch.Miss());
+            } else if (resp.found) {
+              resolve(new CacheListFetch.Hit(resp.found.values));
+            } else {
+              resolve(new CacheListFetch.Error(cacheServiceErrorMapper(err)));
+            }
+          } else {
+            resolve(new CacheListFetch.Error(cacheServiceErrorMapper(err)));
```
poppoerika, line 308: I think we can use switch like [this](https://github.com/momentohq/client-sdk-javascript/blob/main/src/internal/cache-client.ts#L232)?

src/messages/responses/cache-list-push-front.ts
```diff
@@ -0,0 +1,26 @@
...
+import {ResponseBase} from './response-base';
+import {applyMixins, ErrorBody} from '../../errors/error-utils';
+
+export abstract class Response extends ResponseBase {}
+
+export class Success extends Response {
+  private readonly _list_length: number;
+  constructor(list_length: number) {
+    super();
+    this._list_length = list_length;
```
poppoerika, line 11: Do we want to make sure if `list_length` is within the range of `number` type like [this](https://github.com/momentohq/client-sdk-dotnet-incubating/blob/main/src/Momento.Sdk.Incubating/Responses/CacheListPushFrontResponse.cs#L17)?

src/simple-cache-client.ts
virratanasangpunth: +1 to just `listName: string`

src/messages/responses/cache-list-fetch.ts
virratanasangpunth: The type here (`Uint8Array[]`) seems right. 

style nit: I see we sometimes use `_` as the prefix and sometimes we don't do that.

src/simple-cache-client.ts
schwern: list_name in the protobuf is bytes. Are we only taking strings in the SDK?

virratanasangpunth: Oh hmm, looking at the other methods we do allow key to be `Uint8Array`. So I'm actually supportive of `string | Uint8Array`. In csharp-sdk we don't allow this, but I think it's better here to align with the the other method signatures in this sdk.

src/internal/cache-client.ts
schwern: Yes, the protobuf args are very confusing. I'm happy to take it on and coordinate between our branches.

src/simple-cache-client.ts
schwern: I'm happy to take only strings and add bytes later if it's necessary.

src/messages/responses/cache-list-fetch.ts
schwern: @virratanasangpunth Seems one can't have a property and a method with the same name, so I went with _private style.

@poppoerika PHP doesn't. dotnet-incubating does. Sure, I'll add it.

src/internal/cache-client.ts
poppoerika: 🙇‍♀️

src/messages/responses/cache-list-fetch.ts
poppoerika: We're using the .NET SDKs (incubating soon to be merged into main) as our SDK foundation, so I think we want to follow the .NET SDK rather than the PHP SDK.
Thank you!

src/internal/cache-client.ts
schwern: I couldn't figure out how to make it work. The other response has a single property to check against an enum. This one has multiple boolean properties.

poppoerika: How about like this?
```
// enum needs to be outside of CacheClient class
enum DataStructureResponse {
  None = 'none',
  Found = 'found',
  Missing = 'missing',
}

(err, resp) => {
          if (resp) {
            switch (resp.list) {
              case DataStructureResponse.Missing:
                resolve(new CacheListFetch.Miss());
                break;
              case DataStructureResponse.Found:
                resolve(new CacheListFetch.Hit(resp.found.values));
                break;
              default:
                resolve(new CacheListFetch.Error(cacheServiceErrorMapper(err)));
                break;
            }
          } else {
            resolve(new CacheListFetch.Error(cacheServiceErrorMapper(err)));
          }
        }
```

schwern: That'll work. However, it's a lot of extra scaffolding for the same code. It turns the original protobuf number into a string and back into a number just to use a switch.

poppoerika: Do we need to do that?
> It turns the original protobuf number into a string and back into a number just to use a switch.

I thought `resp.list` returns string (`missing`, `found`, `none`).
Sorry if I misunderstood your concern.

poppoerika: At least that's what my VS code tells me 🤷‍♀️ 
image

poppoerika: But I agree that if it's too much work to make this a switch statement, then it's fine to use a if statement here 👍

schwern: I tried this.

```
        (err, resp) => {
          switch (resp) {
            case resp.missing instanceof cache_client._ListFetchResponse._Missing:
              resolve(new CacheListFetch.Miss());
              break;
            case resp.found instanceof cache_client._ListFetchResponse._Found:
              resolve(new CacheListFetch.Hit(resp.found.values));
              break;
            default:
              resolve(new CacheListFetch.Error(cacheServiceErrorMapper(err)));
              break;
          }
        }
```

Which *should* work, but eslint doesn't recognize that `resp.found` is a _Found object.

```
  303:46  error  Unsafe argument of type `any` assigned to a parameter of type `Uint8Array[]`  @typescript-eslint/no-unsafe-argument
  303:46  error  Unsafe member access .values on an `any` value                                @typescript-eslint/no-unsafe-member-access
```

schwern:
```
        (err, resp) => {
          switch (true) {
            case typeof resp === 'undefined':
            default:
              resolve(new CacheListFetch.Error(cacheServiceErrorMapper(err)));
              break;
            case resp.missing instanceof
              cache_client._ListFetchResponse._Missing:
              resolve(new CacheListFetch.Miss());
              break;
            case resp.found instanceof cache_client._ListFetchResponse._Found:
              resolve(new CacheListFetch.Hit(resp.found.values));
              break;
          }
        }
```

Doesn't work either, eslint thinks resp can be undefined.

I have to stop obsessing over this, so I asked StackOverflow.
https://stackoverflow.com/questions/75102169/how-do-i-narrow-a-value-to-not-be-undefined

schwern: They came through. https://stackoverflow.com/a/75102198/14660

```
        (err, resp) => {
          if (resp?.missing) {
            resolve(new CacheListFetch.Miss());
          } else if (resp?.found) {
            resolve(new CacheListFetch.Hit(resp.found.values));
          } else {
            resolve(new CacheListFetch.Error(cacheServiceErrorMapper(err)));
          }
        }
```

For some reason this doesn't work as a switch. 🤷

src/messages/responses/cache-list-fetch.ts
```diff
@@ -0,0 +1,33 @@
...
+  constructor(values: Uint8Array[]) {
+    super();
+    this._values = values;
+  }
+
+  public values(): Uint8Array[] {
```
poppoerika, line 14: Should we rename this to `valueListUint8Array`? 
That's how it's named in the .NET. 
https://github.com/momentohq/client-sdk-dotnet-incubating/blob/main/src/Momento.Sdk.Incubating/Responses/CacheListFetchResponse.cs#L34

Also that way, we don't need to use `_` [here](https://github.com/momentohq/client-sdk-javascript/pull/149/files#diff-d6754ba5098f5e6fb3b8604bc5ceaeca7fdbe136912365868476fe9cccdaf768R8).

src/internal/cache-client.ts
schwern: That works. We can check more directly...

```
          switch (resp?.list) {
            case 'missing':
              resolve(new CacheListFetch.Miss());
              break;
            case 'found':
              resolve(new CacheListFetch.Hit(resp.found.values));
              break;
            default:
              resolve(new CacheListFetch.Error(cacheServiceErrorMapper(err)));
              break;
          }
```

poppoerika: Can we use enum so that we can use missing/found in other data structure APIs uniformly?
Something like this?
```
enum DataStructureResponse {
  None = 'none',
  Found = 'found',
  Missing = 'missing',
}
```

schwern: Then we're back to adding a lot of extra code just to use a switch.

I think the if/elseif/else here is fine. As long as its DRY.

virratanasangpunth (approved):
LGTM

src/utils/validators.ts
```diff
@@ -1,17 +1,23 @@
 import {InvalidArgumentError} from '../errors/errors';
 
 export function validateCacheName(name: string) {
-  if (!name.trim()) {
+  if (isEmpty(name)) {
```
poppoerika, line 4: I believe `isEmpty` returns `false` even if a passed value is an empty string, which we don't want?
i.e. `isEmpty("   ")` returns `false`.

poppoerika (dismissed):

poppoerika:
@schwern It looks good to me too! 
One question I have is: do we want to create API-based test files instead of grouping APIs and make one test? (for example, create one test file for list APIs and one for set APIs, etc?)
If we want to do that, then I need to update my PR based on this.

schwern:
@virratanasangpunth 

> do we want to create API-based test files instead of grouping APIs and make one test? (for example, create one test file for list APIs and one for set APIs, etc?)

Yes, if you can avoid copying a bunch of scaffolding code. Otherwise, lump the integration tests together for now. I plan on doing a refactoring pass to DRY up the code and split up that monster test.

virratanasangpunth (approved):

poppoerika (approved):

PR #155, opened 2023-01-12
feat: add SetFetch operation

virratanasangpunth:
### Info

Since this is my first PR I'm creating a small PR to get my foot in the door/get the feet wet.

This PR adds support for SetFetch operation, which will return the entire Set back for a given
cacheName/setName.

In a subsequent PR once SetUnion/SetDifference operations are added, more integ-test cases will be added as well.

### Testing
```
npm run integration

 PASS  integration/simple-cache-client.test.ts
  SimpleCacheClient.ts Integration Tests
    ✓ should create and delete a cache, set and get a value (555 ms)
    ✓ should return NotFoundError if deleting a non-existent cache (20 ms)
    ✓ should return AlreadyExists response if trying to create a cache that already exists (388 ms)
    ✓ should create 1 cache and list the created cache (408 ms)
    ✓ should set and get string from cache (91 ms)
    ✓ should set and get bytes from cache (91 ms)
    ✓ should set string key with bytes value (29 ms)
    ✓ should set byte key with string value (29 ms)
    ✓ should set and get string from cache and returned set value matches string cacheValue (14 ms)
    ✓ should set string key with bytes value and returned set value matches byte cacheValue (16 ms)
    ✓ should timeout on a request that exceeds specified timeout (424 ms)
    ✓ should set and then delete a value in cache (56 ms)
    ✓ should return InvalidArgument response for set, get, and delete with empty key
    ✓ should return InvalidArgument response for set, get, and delete with invalid cache name (1 ms)
    ✓ should return InvalidArgument response for set request with empty key or value
    ✓ should return InvalidArgument response for get request with empty key (1 ms)
    ✓ should create, list, and revoke a signing key (486 ms)
  Integration Tests for operations on sets datastructure
    ✓ should return MISSING if set does not exist (17 ms)
```


src/messages/responses/set-fetch.ts
```diff
@@ -0,0 +1,37 @@
...
+export class Found extends Response {
+  private readonly elements: Uint8Array[];
+
+  constructor(elements: Uint8Array[]) {
+    super();
+    this.elements = elements;
+  }
+
+  public valueSetString(): Set<string> {
+    return new Set(this.elements.map(e => TEXT_DECODER.decode(e)));
```
schwern, line 21: Hopefully `TextDecoder.decode` is copy-on-write.

integration/simple-cache-client.test.ts
```diff
@@ -375,3 +376,13 @@ describe('SimpleCacheClient.ts Integration Tests', () => {
     ).toEqual(false);
   });
 });
+
+describe('Integration Tests for operations on sets datastructure', () => {
```
schwern, line 380: I guess you need SetAdd to do more testing. Would you add a test for an invalid set name?

src/internal/cache-client.ts
```diff
@@ -152,6 +154,47 @@ export class CacheClient {
...
+              resolve(new SetFetch.Missing());
+            } else if (resp.found) {
+              resolve(new SetFetch.Found(resp.found.elements));
+            } else {
+              resolve(new SetFetch.Error(cacheServiceErrorMapper(err)));
+            }
+          } else {
+            resolve(new SetFetch.Error(cacheServiceErrorMapper(err)));
+          }
+        }
```
schwern, line 193: I ([meaning Stack Overflow](https://stackoverflow.com/a/75102198/14660)) found a better pattern.
```suggestion
        (err, resp) => {
          if (resp?.missing) {
            resolve(new CacheListFetch.Miss());
          } else if (resp?.found) {
            resolve(new CacheListFetch.Hit(resp.found.values));
          } else {
            resolve(new CacheListFetch.Error(cacheServiceErrorMapper(err)));
          }
        }
```

src/messages/responses/set-fetch.ts
```diff
@@ -0,0 +1,37 @@
...
+import {applyMixins, ErrorBody} from '../../errors/error-utils';
+import {TextDecoder} from 'util';
+
+const TEXT_DECODER = new TextDecoder();
+
+export abstract class Response extends ResponseBase {}
+
+export class Missing extends Response {}
+
+export class Found extends Response {
```
schwern, line 12: Call this Miss and Hit to match Get.

schwern (changes requested):

src/internal/cache-client.ts
```diff
@@ -179,16 +179,12 @@ export class CacheClient {
...
-          if (resp) {
-            if (resp.missing) {
-              resolve(new SetFetch.Missing());
-            } else if (resp.found) {
-              resolve(new SetFetch.Found(resp.found.elements));
-            } else {
-              resolve(new SetFetch.Error(cacheServiceErrorMapper(err)));
-            }
+          if (resp?.missing) {
+            resolve(new CacheListFetch.Miss());
```
schwern: Sorry, I didn't get my suggestion fixed in time.

SetFetch.Miss() and you'll have to change SetFetch.Missing to SetFetch.Miss

src/internal/cache-client.ts
```diff
@@ -179,16 +179,12 @@ export class CacheClient {
...
+          } else if (resp?.found) {
+            resolve(new CacheListFetch.Hit(resp.found.values));
```
schwern: SetFetch.Hit and SetFetch.Found -> Hit.

src/internal/cache-client.ts
```diff
@@ -179,16 +179,12 @@ export class CacheClient {
...
           } else {
-            resolve(new SetFetch.Error(cacheServiceErrorMapper(err)));
+            resolve(new CacheListFetch.Error(cacheServiceErrorMapper(err)));
```
schwern: SetFetch.Error

schwern (changes requested):

schwern (dismissed):

src/internal/cache-client.ts
```diff
@@ -152,6 +154,43 @@ export class CacheClient {
...
+    setName: string
+  ): Promise<SetFetch.Response> {
+    validateCacheName(cacheName);
+    validateSetName(setName);
+    return await this.sendSetFetch(cacheName, this.convert(setName));
+  }
+
+  private async sendSetFetch(
+    cacheName: string,
+    setName: Uint8Array
```
poppoerika, line 168: I thought we decided that we only allow string for `setName`?

src/internal/cache-client.ts
```diff
@@ -152,6 +154,43 @@ export class CacheClient {
...
+    const metadata = this.createMetadata(cacheName);
+    return await new Promise(resolve => {
+      this.clientWrapper.getClient().SetFetch(
+        request,
+        metadata,
+        {
+          interceptors: this.interceptors,
+        },
+        (err, resp) => {
+          if (resp?.missing) {
```
poppoerika, line 182: Yay, optional chaining operator.

integration/simple-cache-client.test.ts
virratanasangpunth: Yea there'll be more test cases once I add the other operations.

src/messages/responses/set-fetch.ts
poppoerika: @schwern what does it mean? and why do we want copy-on-write here? (I'm not familiar with the concept...)

virratanasangpunth: If someone tries to tamper with the underlying byte array (which I think they can get via `valueSetByteArray()`) ,then will that tampering change the values returned from `valueSetString()`?

I think that's what he's trying to get at.

virratanasangpunth: Gonna write some unit tests to verify this.

src/internal/cache-client.ts
schwern: We did, but the protobuf takes bytes. SetFetch takes a string and passes bytes to sendSetFetch to pass to the request.

poppoerika: Ah I mistook the above `setFetch` as a client-facing API. Sorry.
Got it!

src/messages/responses/set-fetch.ts
poppoerika: Ohhh, I see. Thank you for the explanation, Vir!

src/messages/responses/set-fetch.ts
```diff
@@ -0,0 +1,37 @@
+import {ResponseBase} from './response-base';
```
poppoerika, line 1: Do we want to name this file `cache-set-fetch` to align with the .NET SDK?

src/messages/responses/set-fetch.ts
```diff
@@ -0,0 +1,37 @@
...
+import {applyMixins, ErrorBody} from '../../errors/error-utils';
+import {TextDecoder} from 'util';
+
+const TEXT_DECODER = new TextDecoder();
+
+export abstract class Response extends ResponseBase {}
+
+export class Miss extends Response {}
+
+export class Hit extends Response {
```
poppoerika, line 12: nit: the order we use in the .NET SDK is Hit, Miss, Error.

src/messages/responses/set-fetch.ts
virratanasangpunth: Ummmm.... yes... good callout.

src/messages/responses/set-fetch.ts
schwern: Copy-on-write (COW) is where multiple variables share the same underlying memory until one of them changes.

The primary purpose is to save memory. In this case, TextDecoder.decode shouldn't need to copy the values just to change their encoding. `this.elements` and the returned `Set<string>` should share the same memory.

@virratanasangpunth It's an optimization, it shouldn't be detectable in tests.

```
str1 = "test"

# without COW the string is copied here
# with COW, the string is not copied. Both str1 and str2 share the same underlying memory for the string.
str2 = str1

# with COW, now str2 makes a copy and changes it.
str2.concat(" more")
```

schwern (approved):

src/messages/responses/set-fetch.ts
virratanasangpunth: Hmmm I'm still exploring this topic in my local branch.

I think I"m gonna merge the PR as-is and I can come back to optimize/fix (if needed) once I'm done exploring.

schwern: I wouldn't worry about it. It's hidden from the user. We can always optimize later.

virratanasangpunth: > It's hidden from the user

I think that's the part where I'm not 100% convinced :) but you're probably right.

PR #167, opened 2023-01-13
refactor: Make it clearer what cache.cache_client is.

schwern:
It was unclear to me that cache.cache_client and control.control_client are GRPC types. They can be confused with the actual client objects.

### DIFF
--- a/src/internal/cache-client.ts
+++ b/src/internal/cache-client.ts
@@ -1,4 +1,5 @@
 import {cache} from '@gomomento/generated-types';
+import grpcCache = cache.cache_client;
 // older versions of node don't have the global util variables https://github.com/nodejs/node/issues/20365
 import {TextEncoder} from 'util';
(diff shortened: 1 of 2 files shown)


src/internal/cache-client.ts
```diff
@@ -1,4 +1,5 @@
 import {cache} from '@gomomento/generated-types';
+import cache_types = cache.cache_client;
```
virratanasangpunth, line 2: Why is it called `cache_types`? I think something like `grpc_cache_client` is clearer?

schwern: Because it's from "generated-types", but yeah, they aren't all types.

I want to avoid "client". `grpc_cache` and `grpc_control`?

virratanasangpunth: yup that works.

virratanasangpunth (approved):

PR #171, opened 2023-01-17
feat: basic implementation of SetAddElements and SetRemoveElements

virratanasangpunth:
There is a few small TODOs left that can be fast-followed after this
PR is merged.

Integ tests are not added yet, that will be done in the next PR.

### DIFF
--- a/src/internal/cache-client.ts
+++ b/src/internal/cache-client.ts
@@ -12,6 +12,8 @@ import * as CacheGet from '../messages/responses/cache-get';
 import * as CacheSet from '../messages/responses/cache-set';
 import * as CacheDelete from '../messages/responses/cache-delete';
 import * as CacheSetFetch from '../messages/responses/cache-set-fetch';
+import * as CacheSetAddElements from '../messages/responses/cache-set-add-elements';
+import * as CacheSetRemoveElements from '../messages/responses/cache-set-remove-elements';
 import {version} from '../../package.json';
 import {getLogger, Logger} from '../utils/logging';
 import {IdleGrpcClientWrapper} from '../grpc/idle-grpc-client-wrapper';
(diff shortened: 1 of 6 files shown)


src/internal/cache-client.ts
```diff
@@ -196,6 +198,107 @@ export class CacheClient {
...
+    const metadata = this.createMetadata(cacheName);
+    return await new Promise(resolve => {
+      this.clientWrapper.getClient().SetUnion(
+        request,
+        metadata,
+        {
+          interceptors: this.interceptors,
+        },
+        (err, resp) => {
+          if (err) {
```
cprice404, line 240: nit - lint warnings above can be avoided by simply omitting `resp` from your argument list here.  (I think this is an annoying language design choice and I think your way is more clear though so if you wanted to add the one-line eslint-ignore comment on that arg list that would be fine too.

src/messages/responses/cache-set-add-elements.ts
```diff
@@ -0,0 +1,22 @@
+import {ResponseBase} from './response-base';
+import {SdkError} from '../../errors/errors';
+import {applyMixins, ErrorBody} from '../../errors/error-utils';
+import {TextDecoder} from 'util';
+
```
cprice404, line 5: legit warning

cprice404 (approved):
couple nits re: lint warnings but overall lgtm, and i don't care if you fast-follow those lint warnings personally.  but i guess we need to align on merge order convo this morning.

src/internal/cache-client.ts
virratanasangpunth: kk gonna tackle these as fast-follows: https://github.com/momentohq/client-sdk-javascript/issues/178

PR #173, opened 2023-01-17
feat: add unit and integ tests for sets operations

virratanasangpunth:
you probably want to review/approve this guy first: https://github.com/momentohq/client-sdk-javascript/pull/171

### DIFF
--- a/integration/simple-cache-client.test.ts
+++ b/integration/simple-cache-client.test.ts
@@ -33,9 +33,6 @@ const cacheClientProps: SimpleCacheClientProps = {
 const INTEGRATION_TEST_CACHE_NAME =
   process.env.TEST_CACHE_NAME || 'js-integration-test-default';
 
-const LOL_BYTE_ARRAY = Uint8Array.of(108, 111, 108);
-const FOO_BYTE_ARRAY = Uint8Array.of(102, 111, 111);
-
 const deleteCacheIfExists = async (
   momento: SimpleCacheClient,
   cacheName: string
@@ -576,6 +573,8 @@ describe('lists', () => {
 });
 
 describe('Integration Tests for operations on sets datastructure', () => {
+  const LOL_BYTE_ARRAY = Uint8Array.of(108, 111, 108);
+  const FOO_BYTE_ARRAY = Uint8Array.of(102, 111, 111);
   it('should succeed for addElements with byte arrays happy path', async () => {
     const setName = v4();
     const addResponse = await momento.setAddElements(
--- a/integration/simple-cache-client.test.ts
+++ b/integration/simple-cache-client.test.ts
@@ -17,6 +17,8 @@ import {
   RevokeSigningKey,
   CacheSetFetch,
   SimpleCacheClient,
+  CacheSetAddElements,
+  CacheSetRemoveElements,
 } from '../src';
 import {TextEncoder} from 'util';
 import {SimpleCacheClientProps} from '../src/simple-cache-client-props';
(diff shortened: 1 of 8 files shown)


integration/simple-cache-client.test.ts
```diff
@@ -31,6 +33,9 @@ const cacheClientProps: SimpleCacheClientProps = {
 const INTEGRATION_TEST_CACHE_NAME =
   process.env.TEST_CACHE_NAME || 'js-integration-test-default';
 
+const LOL_BYTE_ARRAY = Uint8Array.of(108, 111, 108);
+const FOO_BYTE_ARRAY = Uint8Array.of(102, 111, 111);
```
schwern, line 37: Put these inside the describe block where they're used?

integration/simple-cache-client.test.ts
```diff
@@ -571,6 +576,221 @@ describe('lists', () => {
...
 
 describe('Integration Tests for operations on sets datastructure', () => {
+  it('should succeed for addElements with byte arrays happy path', async () => {
+    const setName = v4();
+    const addResponse = await momento.setAddElements(
+      INTEGRATION_TEST_CACHE_NAME,
+      setName,
+      [LOL_BYTE_ARRAY, FOO_BYTE_ARRAY]
+    );
+    expect(addResponse).toBeInstanceOf(CacheSetAddElements.Success);
```
schwern, line 586: I wonder if we could DRY up these tests by putting some of these really common operations into functions.

Something like this in the `describe` block.

```
const addToSet = function(setName: string, values: string[] | Uint8Array[], ttl: CollectionTtl?) {
  const addResponse = await momento.setAddElements(
    INTEGRATION_TEST_CACHE_NAME,
    setName,
    values,
    ttl
  );
  expect(addResponse).toBeInstanceOf(CacheSetAddElements.Success);
};
```

Then use it like so:

```suggestion
    const setName = v4();
    addToSet(setName, [LOL_BYTE_ARRAY, FOO_BYTE_ARRAY]);
```

virratanasangpunth: gonna timebox this: I see more compilation errors than I wanna see atm.

integration/simple-cache-client.test.ts
virratanasangpunth: sure.

integration/simple-cache-client.test.ts
virratanasangpunth: I propose that we punt on this, unless you think this is a blocker.

Also won't this code have to change with your refactor PR anyway?

integration/simple-cache-client.test.ts
```diff
@@ -571,6 +573,223 @@ describe('lists', () => {
...
+      setName
+    );
+    expect(fetchResponse).toBeInstanceOf(CacheSetFetch.Hit);
+    expect((fetchResponse as CacheSetFetch.Hit).valueSetUint8Array()).toEqual(
+      new Set([LOL_BYTE_ARRAY, FOO_BYTE_ARRAY])
+    );
+  });
+  it('should succeed for addElements with byte arrays happy path with no refresh ttl', async () => {
+    const setName = v4();
+    let addResponse = await momento.setAddElements(
```
cprice404, line 598: I don't like that we are perpetuating this pattern of using mutable variables in all of these tests when they are not necessary, but not a blocker.

integration/simple-cache-client.test.ts
cprice404: I think the suggestion is a good one but I would be fine punting or fast-following it if we are happy with this PR otherwise.

cprice404 (approved):
🙈

Issue #146, opened 2023-01-11
fast follows for JS Configuration API


cprice404:
Based on review of the original PR https://github.com/momentohq/client-sdk-javascript/pull/141 , we have the following outstanding tasks to take care of:

Before publicly sharing with customers:

- [x] add InRegion pre-built configurations
- [x] look at constructors for pre-built configurations, decide if they can be / should be inherited
- [x] rename Config interface based on Vir's feedback
- [x] re-introduce logger options
- [x] consider renaming copy constructors from `with` to `clone`
- [x] do a release
- [x] update examples

Not blocking sharing with customers but need to be addressed now:

- [x] docstrings for public config APIs and pre-built configurations (note that at least SimpleCacheClientProps docstrings are out of date)
- [x] wire up RetryStrategy
- [x] move idle timeout setting into transport strategy, remove from top-level configuration for now
- [x] remove CertificateName settings for now, they can be added later if requested by a customer
- [x] add `String`-backed implementation of credential provider interface
- [x] add unit tests for config copy constructors
- [x] remove max concurrent requests setting for now
- [x] withClientTimeout either needs to use a duration object or be renamed to withClientTimeoutMillis
- [x] evaluate behavior user were to configure client side timeout to zero, and see how we are handling this in other SDKs

PR #176, opened 2023-01-18
chore: move max idle millis setting into transport strategy

cprice404:
This addresses one of the outstanding items from the list
of fast-follows for the Configuration API:

https://github.com/momentohq/client-sdk-javascript/issues/146

Prior to this commit we were exposing 'maxIdleMillis' as a
top-level setting on the Configuration.  However, we do not
expect users to need to know about this setting in the 90%
use case, so this commit moves it down into the TransportStrategy.

### DIFF
--- a/src/config/configuration.ts
+++ b/src/config/configuration.ts
@@ -1,6 +1,17 @@
 import {TransportStrategy} from './transport/transport-strategy';
 import {LoggerOptions} from '../utils/logging';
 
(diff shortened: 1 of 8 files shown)


pgautier404 (approved):

schwern (approved):

schwern (approved):

schwern:
Looks like I merged the PRs in the wrong order.

Issue #170, opened 2023-01-17
js-sdk: allows SetAdd/SetRemove on a single Element


virratanasangpunth:

virratanasangpunth:
In the dotnet-sdk we have two flavors of SetAdd: `SetAddElements` and `SetAddElement`: https://github.com/momentohq/client-sdk-dotnet-incubating/blob/main/src/Momento.Sdk.Incubating/ISimpleCacheClient.cs#L177 (same here for SetRemove).

In the js-sdk I've only implemented `SetAddElements`. We should consider adding `SetAddElement` too.

virratanasangpunth:
first PR is up: https://github.com/momentohq/client-sdk-javascript/pull/182

virratanasangpunth:
second PR: https://github.com/momentohq/client-sdk-javascript/pull/190

PR #182, opened 2023-01-18
feat: add CacheSetAddElement that accepts a single element

virratanasangpunth:
For integ-testing, since CacheSetAddElement simply delegates
to CacheSetAddElements, the integ test cases are basic smoke
test cases and are not as exhaustive as CacheSetAddElements
integ test cases.

https://github.com/momentohq/client-sdk-javascript/issues/170

### DIFF
--- a/src/simple-cache-client.ts
+++ b/src/simple-cache-client.ts
@@ -182,12 +182,14 @@ export class SimpleCacheClient {
     element: string | Uint8Array,
     ttl?: CollectionTtl
   ): Promise<CacheSetAddElement.Response> {
-    return await this.setAddElements(
-      cacheName,
-      setName,
-      [element] as string[] | Uint8Array[],
-      ttl
-    );
+    return (
+      await this.setAddElements(
+        cacheName,
+        setName,
+        [element] as string[] | Uint8Array[],
+        ttl
+      )
+    ).toSingularResponse();
   }
 
   /**
--- a/src/simple-cache-client.ts
+++ b/src/simple-cache-client.ts
@@ -18,6 +18,7 @@ import * as CacheDictionaryGetFields from './messages/responses/cache-dictionary
 import * as CacheDictionaryIncrement from './messages/responses/cache-dictionary-increment';
 import * as CacheSetFetch from './messages/responses/cache-set-fetch';
 import * as CacheSetAddElements from './messages/responses/cache-set-add-elements';
+import * as CacheSetAddElement from './messages/responses/cache-set-add-element';
 import * as CacheSetRemoveElements from './messages/responses/cache-set-remove-elements';
 import {getLogger, initializeMomentoLogging, Logger} from './utils/logging';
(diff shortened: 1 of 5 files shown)


src/simple-cache-client.ts
```diff
@@ -165,6 +166,30 @@ export class SimpleCacheClient {
...
+  public async setAddElement(
+    cacheName: string,
+    setName: string,
+    element: string | Uint8Array,
+    ttl?: CollectionTtl
+  ): Promise<CacheSetAddElement.Response> {
+    return await this.setAddElements(
+      cacheName,
+      setName,
+      [element] as string[] | Uint8Array[],
```
virratanasangpunth, line 188: uggggh this is an ugly casting that I can't figure out how to remove.

src/simple-cache-client.ts
```diff
@@ -165,6 +166,32 @@ export class SimpleCacheClient {
...
+    ttl?: CollectionTtl
+  ): Promise<CacheSetAddElement.Response> {
+    return (
+      await this.setAddElements(
+        cacheName,
+        setName,
+        [element] as string[] | Uint8Array[],
+        ttl
+      )
+    ).toSingularResponse();
```
virratanasangpunth, line 192: LMK if someone can figure out how to get rid of this ugly cast `as string[] | Uint8Array[]`

src/simple-cache-client.ts
```diff
@@ -165,6 +166,32 @@ export class SimpleCacheClient {
...
+    cacheName: string,
+    setName: string,
+    element: string | Uint8Array,
+    ttl?: CollectionTtl
+  ): Promise<CacheSetAddElement.Response> {
+    return (
+      await this.setAddElements(
+        cacheName,
+        setName,
+        [element] as string[] | Uint8Array[],
```
virratanasangpunth, line 189: LMK if someone can figure out how to get rid of this ugly cast as string[] | Uint8Array[]

cprice404: you could just define the var as an explicitly typed const first:

```
const elements: string[] | Uint8Array[] = [element];
```

virratanasangpunth: The compiler won't allow that... unless I do the same ugly cast:
```
const elements: string[] | Uint8Array[] = [element] as string[] | Uint8Array[]
```

schwern: If it makes you feel better, I ran into the same problem and did something even worse.

```
      if (typeof props.value === 'string') {
        return Momento.listConcatenateFront(
          props.cacheName,
          props.listName,
          [props.value],
          props.ttl,
          props.truncateBackToSize
        );
      } else {
        return Momento.listConcatenateFront(
          props.cacheName,
          props.listName,
          [props.value],
          props.ttl,
          props.truncateBackToSize
        );
      }
```

schwern (dismissed):

schwern (approved):

PR #183, opened 2023-01-18
fix: Small fixes to data structures.

schwern:
A few small fixes I noticed.

* ListPushFront was using ListFetch types.
* Use convertArray() everywhere to map arrays to bytes.

### DIFF
--- a/src/internal/cache-client.ts
+++ b/src/internal/cache-client.ts
@@ -818,11 +818,10 @@ export class CacheClient {
     this.logger.trace(
       `Issuing 'dictionaryGetFields' request; fields: ${fields.toString()}`
     );
-    const encodedFields = fields.map(field => this.convert(field));
     const result = await this.sendDictionaryGetFields(
       cacheName,
       this.convert(dictionaryName),
-      encodedFields
+      this.convertArray(fields)
     );
     this.logger.trace(
       `'dictionaryGetFields' request result: ${result.toString()}`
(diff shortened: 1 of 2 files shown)


src/internal/cache-client.ts
```diff
@@ -462,12 +462,12 @@ export class CacheClient {
     value: string | Uint8Array,
     ttl: CollectionTtl = CollectionTtl.fromCacheTtl(),
     truncateBackToSize?: number
-  ): Promise<CacheListFetch.Response> {
+  ): Promise<CacheListPushFront.Response> {
```
virratanasangpunth, line 465: seems like this should've caused an integ test failure?

virratanasangpunth (approved):

src/internal/cache-client.ts
schwern: Sure does!

I think typescript's type checking is a lot more lax than we'd like. `CacheListFetch.Response` and `CacheListPushFront.Response` implement the same methods, so maybe it saw them as equivalent?

virratanasangpunth: I took a quick glance at the integ tests for `ListPushFront` and I think the integ-test didn't fail because we're not asserting the actual type of the response, e.g. here https://github.com/momentohq/client-sdk-javascript/blob/77628ff08b9dbbed825e893f98a06387a09b2635/integration/list.test.ts#L56

schwern: I'll add a test.

PR #184, opened 2023-01-18
refactor: Reduce the number of places we need to add a response import.

schwern:
A quality of life refactor.

* UnknownError was not being exported. Seems it should be available to the user.


virratanasangpunth (approved):
LGTM. But this will result in git merge conflict with my PR that I've got out.

PR #188, opened 2023-01-19
refactor: DRY the simple success responses.

schwern:
A lot of responses are the same: an empty Success and a simple Error. This avoids having to copy them.


virratanasangpunth (approved):
Sure.

PR #190, opened 2023-01-19
feat: add cacheSetRemoveElement method that accepts a single element

virratanasangpunth:


src/messages/responses/cache-set-remove-elements.ts
```diff
@@ -1 +1,25 @@
...
+
+export abstract class Response extends ResponseBase {
+  abstract toSingularResponse(): CacheSetRemoveElement.Response;
+}
+
+class _Success extends Response {
+  toSingularResponse(): CacheSetRemoveElement.Response {
+    return new CacheSetRemoveElement.Success();
+  }
+}
```
virratanasangpunth, line 13: @schwern curious to hear if you have an opinion on how to extend the `./response-simple-success` construct to include  `toSingularResponse()`

schwern: @virratanasangpunth Something like this maybe?

```
import { default as SimpleSuccess } from './response-simple-success'

export class Success extends SimpleSuccess.Success {
  toSingularResponse(): CacheSetRemoveElement.Response {
    return new CacheSetRemoveElement.Success();
  }
}
```

schwern: @virratanasangpunth This works.

```
import * as SimpleSuccess from './response-simple-success';
import * as CacheSetAddElement from './cache-set-add-element';

export abstract class Response extends SimpleSuccess.Response {
  abstract toSingularResponse(): CacheSetAddElement.Response;
}

export class Success extends SimpleSuccess.Success {
  toSingularResponse(): CacheSetAddElement.Response {
    return new CacheSetAddElement.Success();
  }
}

export class Error extends SimpleSuccess.Error {
  toSingularResponse(): CacheSetAddElement.Response {
    return new CacheSetAddElement.Error(this._innerException);
  }
}
```

schwern (dismissed):

schwern (approved):

PR #191, opened 2023-01-19
feat: Add list ops

schwern:
Adds all the list ops except listErase. We're punting on that for now.

Also:
* Fixed CacheDictionaryGetField.Hit.toString()
* DRY the "get value" responses.


virratanasangpunth (dismissed):

src/internal/cache-client.ts
```diff
@@ -413,6 +420,156 @@ export class CacheClient {
...
+  ): Promise<CacheListConcatenateBack.Response> {
+    const request = new grpcCache._ListConcatenateBackRequest({
+      list_name: listName,
+      values: values,
+      ttl_milliseconds: ttlMilliseconds,
+      refresh_ttl: refreshTtl,
+      truncate_front_to_size: truncateFrontToSize,
+    });
+    const metadata = this.createMetadata(cacheName);
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
eaddingtonwhite, line 477: weird u need this lint ignore? I see you using `resolve`?  Or is it because you don't have `reject` defined?

eaddingtonwhite: also why `GrpcClientWrapper` doesnt support async natively? Isnt that something we wrote? Just us doing all these Promise wrappers seems like something off? Maybe missing some context

eaddingtonwhite: Just looked some more at class def. @schwern wdut? Couldn't we have a helper function wrap this in promise down in wrapper client vs make all call sites do it?  Fine if follow up PR as well.

schwern: @eaddingtonwhite That `eslint-disable-next-line` was just cut & code from an existing method. We used to check `if(err)` to `resp` was unused. It's no longer necessary, I've removed it.

As for the rest, I just learned TypeScript last week. 🤷 I've been doing a lot of refactoring of this SDK myself. If you see something that could be done better, its likely it can be.

Anything to reduce the amount of copypasta.

virratanasangpunth (approved):

PR #192, opened 2023-01-19
fix: Set ops return INVALID_ARGUMENT_ERROR, not raise.

schwern:
While DRYing up the tests, I discovered a bug in how Set is treating cache and set name validations.

Also
* DRY'd up the name validaton tests.
* Added name validation tests to Set.

### DIFF
--- a/src/internal/cache-client.ts
+++ b/src/internal/cache-client.ts
@@ -62,7 +62,7 @@ export class CacheClient {
   private readonly interceptors: Interceptor[];
 
   /**
-   * @param {MomentoCacheProps} props
+   * @param {MomentoValidateCacheProps} props
    */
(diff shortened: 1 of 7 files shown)


src/internal/cache-client.ts
```diff
@@ -184,8 +184,12 @@ export class CacheClient {
...
     setName: string
   ): Promise<CacheSetFetch.Response> {
-    validateCacheName(cacheName);
-    validateSetName(setName);
+    try {
+      validateCacheName(cacheName);
+      validateSetName(setName);
+    } catch (err) {
+      return new CacheSetFetch.Error(normalizeSdkError(err as Error));
+    }
```
virratanasangpunth, line 192: whoops ty.

virratanasangpunth (approved):

PR #193, opened 2023-01-20
refactor: DRY the response classes

schwern:
I did some reading on Typescript modules and I don't think the mixins are necessary, a simple class hierarchy will do. This reduces a lot of cut & code; responses can make subclasses or re-export the basic classes. It also potentially lets us reduce a lot of copypasta by giving us superclasses to refer to in shared method signatures.

* Replace mixins with subclassing.
* Remove ResponseBase class, use Response class.
* Remove IResponseError interface, use Error class.


virratanasangpunth (approved):
LGTM

cprice404 (approved):

virratanasangpunth (approved):

PR #195, opened 2023-01-20
refactor: DRY and speed up dictionary tests

schwern:
* Extract shared tests
* Speed up the TTL tests by reducing the timeouts (60s -> 20s)
* Move method tests into individual describe blocks for easier reading.


cprice404 (approved):
this diff is large and makes it tricky to verify that we ended up with 1-1 test coverage before/after, but I am content to take your word for it :)

schwern:
> this diff is large and makes it tricky to verify that we ended up with 1-1 test coverage before/after, but I am content to take your word for it :)

We're better than 1-to-1 now. 😁 

Big changes like this can sometimes be more easily reviewed commit-by-commit.

PR #196, opened 2023-01-20
refactor: Turn on noUnusedLocals, fix errors.

### LINKED ISSUE #116: audit eslint / prettier configs - not catching uninitialized class members
Check and make sure that the eslint/prettier configs are up to date with what we've been using internally.

In particular - I noticed that if I was writing a class that had some private readonly member variables, and those variables were not initialized in the constructor, I didn't get a build/lint error.  In our CDK repos I believe we would catch that at compile time.

schwern:
For some reason, the linter was configured to allow unused local properties. The check is off on nearly all our Typescript repos. There doesn't seem to be a reason not to turn the check on.

Closes #116

### DIFF
--- a/src/errors/errors.ts
+++ b/src/errors/errors.ts
@@ -55,7 +55,6 @@ export abstract class SdkError extends Error {
   protected readonly _errorCode: MomentoErrorCode;
   protected readonly _messageWrapper: string;
   private readonly _transportDetails: MomentoErrorTransportDetails;
-  private readonly _stack: string | undefined;
   constructor(
     message: string,
     code = 0,
--- a/src/simple-cache-client.ts
+++ b/src/simple-cache-client.ts
@@ -34,7 +34,7 @@ import {
   CacheSetRemoveElements,
   CacheSetRemoveElement,
 } from '.';
-import {getLogger, initializeMomentoLogging, Logger} from './utils/logging';
+import {initializeMomentoLogging} from './utils/logging';
 import {range} from './utils/collections';
 import {SimpleCacheClientProps} from './simple-cache-client-props';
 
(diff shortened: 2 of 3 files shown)


src/simple-cache-client.ts
```diff
@@ -54,14 +54,12 @@ export class SimpleCacheClient {
...
   private nextDataClientIndex: number;
   private readonly controlClient: ControlClient;
-  private readonly logger: Logger;
 
   /**
    * Creates an instance of SimpleCacheClient.
    */
   constructor(props: SimpleCacheClientProps) {
     initializeMomentoLogging(props.configuration.getLoggerOptions());
-    this.logger = getLogger(this);
```
cprice404, line 64: Instead of removing this can you just add a debug log message to the constructor or something?  We will be coming back in here and adding logging at some point and I want to make sure it's really easy for devs to figure out how to do it.

cprice404 (commented):
thanks for doing this, I agree with the spirit of it.  just a few questions about the couple of impacted variables.

src/errors/errors.ts
```diff
@@ -55,7 +55,6 @@ export abstract class SdkError extends Error {
   protected readonly _errorCode: MomentoErrorCode;
   protected readonly _messageWrapper: string;
   private readonly _transportDetails: MomentoErrorTransportDetails;
-  private readonly _stack: string | undefined;
```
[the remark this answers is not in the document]
schwern: eslint and a search for `_stack` say we're not.

I know we discussed storing the stack in error messages, but we're not consistent about that.

src/simple-cache-client.ts
schwern: I went to do that, and there's very little to log in SimpleCacheClient. All the work is in CacheClient and ControlClient. I think this is at the wrong level.

cprice404 (approved):

PR #197, opened 2023-01-20
chore: Update all our off-the-shelf actions.

### LINKED ISSUE #165: Update Github actions.
We're getting this from the test action.

```
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, actions/setup-node@v1. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
```

I think it's just a matter of updating the versions of those actions.

schwern:
This should quiet the "Node.js 12 actions are deprecated" warnings.

Closes #165

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
 
     steps:
       - name: Setup repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Verify README generation
         uses: momentohq/standards-and-practices/github-actions/oss-readme-template@gh-actions-v1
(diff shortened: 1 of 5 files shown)


poppoerika (approved):
Thank you for updating that!

It made me wonder if we should update the required Node version in our [README](https://github.com/momentohq/client-sdk-javascript#requirements) to 16 since we only test it on this version?

schwern:
@poppoerika I believe they still support older versions of node, it was just something about using outdated versions of the actions.

You're right that we don't test them. I'll look into adding a test matrix. Thx!

PR #198, opened 2023-01-20
fix: Support minimum of node v14

schwern:
* Increase node to >= v14.
  * v10 and v12 are EOL and Jest dropped support for them.
* Matrix test node v14, v16, and v18
  * Only do the README and commit lint once.
* Fix the examples for v14.


examples/index.ts
```diff
@@ -92,6 +92,7 @@ main()
   .then(() => {
     console.log('success!!');
   })
-  .catch(e => {
-    console.error('failed to get from cache', e);
+  .catch((e: Error) => {
+    console.error(`failed to get from cache ${e.message}`);
+    throw e;
```
cprice404, line 97: thank you for fixing this!  I have noticed that we are missing the throws for these in some places and it makes it really confusing to debug :)

test/cache-get.test.ts
```diff
@@ -1,19 +1,21 @@
...
+import {TextEncoder} from 'util';
+
+const TEXT_ENCODER = new TextEncoder();
 
 describe('CacheGet', () => {
   describe('#toString()', () => {
     it('shows a short value', () => {
       const value = 'Napoleon';
-      const hit = new CacheGet.Hit(Buffer.from(value));
+      const hit = new CacheGet.Hit(TEXT_ENCODER.encode(value));
```
cprice404, line 10: Why the switch from buffer to text encoder?

cprice404 (commented):
thanks for doing this!  It's a really important change.  I just had one question about buffer vs. textencoder.

examples/index.ts
schwern: I should make them consistent and all throw?

test/cache-get.test.ts
schwern: Node 10 and 12 don't have it. Now that we're dropping Node 10 and 12 maybe we should switch from TextEncoder to Buffer. However, TextEncoder is what we use everywhere else, and I don't know the benefits/drawbacks of Buffer vs TextEncoder, so switching is a bigger change.

examples/index.ts
cprice404: I was talking about in other repos where we use TS for infra, I'm not aware of whether there are any more missing in this repo.  If you audited and they all look good in this repo, great!

test/cache-get.test.ts
cprice404: okay

cprice404 (approved):

PR #203, opened 2023-01-25
chore: demo ubuntu-22.04 (latest) errors

malandis:
Actions fail on 22.04 but not on 20.04 in GitHub. When run locally on
22.04, the tests succeed.


cprice404:
this is to try to debug https://github.com/actions/runner-images/issues/6981

PR #209, opened 2023-01-27
fix: Prevent infinite loop in example

nand4011:
Use loose equality when checking for the presence of a list caches
token so that null can be compared to undefined. Fixes an infinte
loop that is only broken by rate limiting.

Give the token to the list caches call to prevent an infinite loop
that would occur if there were more than one page of caches.

### DIFF
--- a/examples/index.ts
+++ b/examples/index.ts
@@ -39,9 +39,9 @@ const main = async () => {
   }
 
   console.log('Listing caches:');
-  let token;
+  let token:string|undefined;
   do {
-    const listResponse = await momento.listCaches();
+    const listResponse = await momento.listCaches(token);
     if (listResponse instanceof ListCaches.Error) {
       console.log(`Error listing caches: ${listResponse.message()}`);
       break;
@@ -51,7 +51,7 @@ const main = async () => {
       });
       token = listResponse.getNextToken();
     }
-  } while (token !== null);
+  } while (token != null);
 
   const exampleTtlSeconds = 10;
   console.log(
--- a/examples/package-lock.json
+++ b/examples/package-lock.json
@@ -19,6 +19,7 @@
         "@types/node": "^16.11.4",
         "@types/node-fetch": "2.6.2",
         "@typescript-eslint/eslint-plugin": "5.30.5",
+        "@typescript-eslint/parser": "^5.0.0",
         "eslint": "8.19.0",
         "eslint-config-prettier": "8.5.0",
         "eslint-plugin-import": "2.26.0",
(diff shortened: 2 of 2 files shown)


examples/index.ts
```diff
@@ -39,9 +39,9 @@ const main = async () => {
   }
 
   console.log('Listing caches:');
-  let token;
+  let token: string | undefined;
   do {
-    const listResponse = await momento.listCaches();
+    const listResponse = await momento.listCaches(token);
```
cprice404, line 44: lol wow nice catch, thanks!

cprice404 (approved):
lgtm; one nit/TS learning for the future; I wouldn't have expected the `package-lock.json` files to be modified by this change.  If you use `npm ci` to install your deps, it will honor what is captured in the `package-lock.json` files rather than potentially updating them.

nbd but just a thing to know :)

nand4011:
The package-lock.json files were confusing to me as well. I'll try to rebuild with npm ci to see if I can get rid of them.

cprice404 (approved):

examples/index.ts
cprice404: oh wait, so what happens the first time through the loop here?  We are passing token in as `undefined`, but that works and we still get the list successfully?

examples/index.ts
```diff
@@ -39,9 +39,9 @@ const main = async () => {
   }
 
   console.log('Listing caches:');
-  let token;
+  let token: string | undefined;
```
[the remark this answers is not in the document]
nand4011: I don't have the javascript experience to have a well formed opinion here, but I'm seeing a lot of comments saying that explicitly initializing to `undefined` should be avoided because `undefined` means uninitialized.
Does that mean the `getNextToken()` method should return `null` instead of `undefined`? A question for a different pr perhaps.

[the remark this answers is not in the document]
cprice404: okay then this is fine how it is for now

examples/index.ts
nand4011: It works the same as if the variable hadn't been passed in as per example 4 here:
[https://stackoverflow.com/questions/57243477/can-an-optional-parameter-be-null-in-typescript](url)

It does look like the runtime can be configured to disallow this with `exactOptionalPropertyTypes`:
[https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes](url)

The example works in my testing, but I don't know what is idiomatic here.

nand4011: That line of code won't do anything without pagination working. The actual infinte loop that is fixed here is on line 54. This is just future proofing.

cprice404:
🚢

PR #210, opened 2023-01-31
fix: Differenciate response classes.

schwern:
The switch from classes to mixins in 8d8757a98257ccc0ed6344e611026e24d7912240 caused many classes to actually be aliases. For example, `CacheDictionaryRemoveField.Success` and `CacheDictionaryRemoveFields.Success` were the same class. That means a response from DictionaryRemoveFields was an instance of both `CacheDictionaryRemoveField.Success` and `CacheDictionaryRemoveFields.Success`. The dictionary tests had several bugs of this kind.

This also confused IDEs, they would report the incorrect return type for responses.

Screen Shot 2023-01-30 at 17 16 06

### DIFF
--- a/integration/dictionary.test.ts
+++ b/integration/dictionary.test.ts
@@ -142,7 +142,7 @@ describe('Integration tests for dictionary operations', () => {
         value: 'value2',
         ttl: CollectionTtl.of(timeout * 10).withNoRefreshTtlOnUpdates(),
       });
-      expect(changeResponse).toBeInstanceOf(CacheDictionarySetField.Success);
+      expect(changeResponse).toBeInstanceOf(Success);
       await sleep(timeout * 1000);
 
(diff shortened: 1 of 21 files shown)


cprice404:
Thanks!  It looks like you were able to do this without having to revert your test consolidations that had motivated the original change, correct?

cprice404 (dismissed):

nand4011 (approved):

schwern:
@cprice404 

> Thanks! It looks like you were able to do this without having to revert your test consolidations that had motivated the original change, correct?

Yeah. I had to add those bogus IMiss and IHit interfaces with unfortunately public properties to make it work in lieu of being able to use mixins as types. I don't think they're too offensive.

nand4011 (approved):

PR #217, opened 2023-02-02
chore: replace dictionary set fields array with map

nand4011:
Replace the fields array of key value pairs with a Map.

Tested against the integration tests.


schwern (approved):
lgtm

PR #219, opened 2023-02-03
feat: rename dictionary value accessors, add Record accessors

cprice404:
This commit makes a few changes to the Dictionary APIs to help
simplify our in-browser demo code.

Changes include:
* DictionarySetFields can now be called with a Record literal input
  rather than needing to construct a Map.
* DictionaryFetch and DictonaryGetFields: renamed `valueDictionary*`
  value accessor names to `valueMap*` to make it more obvious what
  the return type will be.
* DictionaryFetch and DictonaryGetFields: added `valueRecord*` accessors
  so that users can avoid dealing with `Map`s if their field names
  are strings.
* DictionarySetFields now accepts a Record or a Map as input.

### DIFF
--- a/src/messages/responses/cache-dictionary-fetch.ts
+++ b/src/messages/responses/cache-dictionary-fetch.ts
@@ -14,11 +14,6 @@ export abstract class Response extends ResponseBase {}
 
 class _Hit extends Response {
   private readonly items: cache_client._DictionaryFieldValuePair[];
-  private readonly dictionaryUint8ArrayUint8Array: Map<Uint8Array, Uint8Array> =
-    new Map();
-  private readonly dictionaryStringString: Map<string, string> = new Map();
-  private readonly dictionaryStringArrayBuffer: Map<string, Uint8Array> =
-    new Map();
   private readonly _displayListSizeLimit = 5;
 
   constructor(items: cache_client._DictionaryFieldValuePair[]) {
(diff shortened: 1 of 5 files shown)


src/messages/responses/cache-dictionary-fetch.ts
```diff
@@ -14,47 +14,58 @@ export abstract class Response extends ResponseBase {}
...
+  public valueMapStringUint8Array(): Map<string, Uint8Array> {
+    return this.items.reduce((acc, item) => {
+      acc.set(TEXT_DECODER.decode(item.field), item.value);
+      return acc;
+    }, new Map<string, Uint8Array>());
+  }
+
+  public valueRecord(): Record<string, string> {
+    return this.valueRecordStringString();
+  }
```
cprice404, line 51: Curious if anyone has any objections to this alias, which hopefully would be the most common use case?

src/messages/responses/cache-dictionary-fetch.ts
```diff
@@ -14,47 +14,58 @@ export abstract class Response extends ResponseBase {}
...
   }
 
-  public valueDictionaryUint8ArrayUint8Array(): Map<Uint8Array, Uint8Array> {
-    for (const item of this.items) {
-      this.dictionaryUint8ArrayUint8Array.set(item.field, item.value);
-    }
-    return this.dictionaryUint8ArrayUint8Array;
+  public valueMap(): Map<string, string> {
+    return this.valueMapStringString();
   }
```
cprice404, line 26: Curious if anyone has any objections to this alias, which hopefully would be the most common use case?

nand4011: Will non-text fields be decoded into garbage strings? I'm alright with it as long as there won't be surprise errors thrown.

nand4011 (approved):
LGTM barring objections on the return type/existence of valueMap() and valueRecord().

src/messages/responses/cache-dictionary-fetch.ts
kvcache: `string | Uint8Array` I think would be harmful - there's nothing in the backing data that would hint to the sdk which is appropriate, so it would vary per field 😬.

pgautier404 (approved):
LGTM

src/messages/responses/cache-dictionary-fetch.ts
nand4011: Fair. If you try to mix byte array and string fields in a dictionary then you can't expect intuitive things to happen.

src/messages/responses/cache-dictionary-fetch.ts
kvcache: i like

kvcache (approved):
I think this looks good.

src/messages/responses/cache-dictionary-fetch.ts
schwern: 👍 People can see the return signature in their IDE.

src/messages/responses/cache-dictionary-get-fields.ts
```diff
@@ -22,12 +22,7 @@ export abstract class Response extends ResponseBase {}
 class _Hit extends Response {
   private readonly items: grpcCache._DictionaryGetResponse._DictionaryGetResponsePart[];
   private readonly fields: Uint8Array[];
-  private readonly dictionaryUint8ArrayUint8Array: Map<Uint8Array, Uint8Array> =
-    new Map();
-  private readonly dictionaryStringString: Map<string, string> = new Map();
-  private readonly dictionaryStringUint8Array: Map<string, Uint8Array> =
-    new Map();
-  public responsesList: CacheDictionaryGetFieldResponseType[] = [];
+  public responses: CacheDictionaryGetFieldResponseType[] = [];
```
schwern, line 25: ❤️ `responseList` -> `responses`

src/simple-cache-client.ts
```diff
@@ -500,7 +500,9 @@ export class SimpleCacheClient {
   public async dictionarySetFields(
     cacheName: string,
     dictionaryName: string,
-    items: Map<string | Uint8Array, string | Uint8Array>,
+    items:
+      | Map<string | Uint8Array, string | Uint8Array>
+      | Record<string, string | Uint8Array>,
```
schwern, line 505: That's a mouthful. Could this be a type alias?

schwern (approved):

src/simple-cache-client.ts
cprice404: sure, can do that in a followup PR

PR #221, opened 2023-02-03
Update constructor in docs for EnvMomentoTokenProvider

allenheltondev:
The example in the README passed the env variable as a param. But it actually needs to be in a `props` object.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ const cacheName = 'cache';
 const cacheKey = 'key';
 const cacheValue = 'value';
 
-const credentialsProvider = new EnvMomentoTokenProvider('MOMENTO_AUTH_TOKEN');
+const credentialsProvider = new EnvMomentoTokenProvider({ environmentVariableName: 'MOMENTO_AUTH_TOKEN' });
 
 const loggerOptions: LoggerOptions = {
   level: LogLevel.INFO,


cprice404:
Hey @allenheltondev - thanks for the PR!  This README is actually generated from a template in `README.template.md`.  I think the issue here is that we haven't yet updated the examples to use the latest release of the package.  Let me know if you wanna have a quick zoom so that I can explain this a bit further; it's not our ideal long-term state but there are some reasons we are doing it this way for now.

@nand4011 should be filing a PR to update the examples soon and that should fix this.  If we merge this as-is it will get overwritten by the template engine.

Thanks a bunch for pointing it out though!  It definitely highlights the need for us to update the examples ASAP.

cprice404 (commented):
Left a comment on the PR - thanks for fixing this!  But it will get overwritten by the template engine if we merge it as is.  LMK if you wanna chat so I can explain further!

cprice404:
> Ok I understand about using the template @cprice404, no worries. Should I just close this if @nand4011 is going to update it in his upcoming change?

Yeah I think so!  I'll have him ping you on his PR so you know when it goes in.

allenheltondev:
Closing due to duplicate effort.

PR #224, opened 2023-02-03
feat: re-work Logging API

cprice404:
Prior to this commit, our logging API was just a quick-and-dirty first pass; it was very inflexible and tightly coupled to Pino.

In this commit we re-work the API to be more like the .NET one. Customers provide a MomentoLoggerFactory that knows how to create MomentoLogger objects, and we use those for our logging.  We provide a default initial implementation based on Pino, but we can trivially add new implementations that support any other existing logging framework that a customer may be using, so that they can be assured that Momento log messages will be routed to the same destinations as any other log messages that they are producing in their apps.


src/config/configuration.ts
```diff
@@ -26,16 +26,16 @@ export interface ConfigurationProps {
 export interface Configuration {
   // TODO: add Middlewares
```
nand4011, line 27: Is this TODO affected by the pr? I don't know what middleware means here.

pgautier404 (dismissed):

cprice404:
rebased

nand4011 (approved):

src/config/configuration.ts
cprice404: nope good question though!  in the .net SDK we have a middleware interface.  we don't have it here yet.  i'm not sure whether i will try to get that in before 1.0 or not.

PR #225, opened 2023-02-03
chore: add dictionary examples

nand4011:
Add an example for the dictionary data type. It mirrors the functionality of the example in the c# sdk.

Update the examples to work with the latest sdk release.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -34,6 +34,22 @@ MOMENTO_AUTH_TOKEN=<YOUR AUTH TOKEN> npm run presigned-url-example
 
 Example Code: [presigned-url-example.ts](presigned-url-example.ts)
 
+## Running the Dictionary Example
+
+This example demonstrates how to use the dictionary data type.
+
+- Node version 16 or higher is required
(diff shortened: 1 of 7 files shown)


cprice404:
FYI @allenheltondev this PR should fix that example.

examples/README.md
```diff
@@ -34,6 +34,22 @@ MOMENTO_AUTH_TOKEN=<YOUR AUTH TOKEN> npm run presigned-url-example
 
 Example Code: [presigned-url-example.ts](presigned-url-example.ts)
 
+## Running the Dictionary Example
+
+This example demonstrates how to use the dictionary data type.
+
+- Node version 16 or higher is required
```
cprice404, line 41: nit: this isn't your fault but while you're in here can we consolidate the requirements stuff up to one section at the top?  and we actually only require node 14 or higher.

cprice404 (commented):
one nit about minor cleanup in the README but otherwise looks great, thanks!

examples/README.md
cprice404: also feel free to do this as a follow-up PR now that i think about it.  whichever you prefer.

cprice404 (approved):

cprice404 (approved):

PR #227, opened 2023-02-03
feat: improve logger exports, add DefaultMomentoLoggerFactory

cprice404:
While updating the examples to use the new logging API, I found a
few things that didn't feel right:

1. The pino logger wasn't exported properly so the imports looked weird
2. It felt weird for users to have to be exposed to the name Pino if they
   weren't really interested in digging in to logs.

This commit fixes the imports, and adds DefaultMomentoLoggerFactory
as a wrapper around the Pino factory to make the UX a little better
for users who don't care about the underlying logger details.

### DIFF
--- a/test/config/configuration.test.ts
+++ b/test/config/configuration.test.ts
@@ -4,11 +4,11 @@ import {
   StaticTransportStrategy,
 } from '../../src/config/transport/transport-strategy';
 import {FixedCountRetryStrategy} from '../../src/config/retry/fixed-count-retry-strategy';
-import {PinoMomentoLoggerFactory} from '../../src/config/logging/pino-momento-logger';
 import {NoopMomentoLoggerFactory} from '../../src/config/logging/noop-momento-logger';
+import {DefaultMomentoLoggerFactory} from '../../src';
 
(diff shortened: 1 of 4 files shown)


test/config/configuration.test.ts
```diff
@@ -4,11 +4,11 @@ import {
   StaticTransportStrategy,
 } from '../../src/config/transport/transport-strategy';
 import {FixedCountRetryStrategy} from '../../src/config/retry/fixed-count-retry-strategy';
-import {PinoMomentoLoggerFactory} from '../../src/config/logging/pino-momento-logger';
 import {NoopMomentoLoggerFactory} from '../../src/config/logging/noop-momento-logger';
+import {DefaultMomentoLoggerFactory} from '../../src';
 
 describe('configuration.ts', () => {
-  const testLoggerFactory = new PinoMomentoLoggerFactory();
+  const testLoggerFactory = new DefaultMomentoLoggerFactory();
```
eaddingtonwhite, line 11: nice @cprice404 this is much better.

pgautier404 (approved):

PR #231, opened 2023-02-06
chore: remove key signing dependency

nand4011:
Delete the key signing code and remove the dependency. This is being done to avoid pulling in an extra dependency for users who don't need key signing. The signing code will be published separately.


cprice404 (approved):
This lgtm, @malandis can you take a quick peek and make sure you think this is the correct scope?

malandis:
Looks good!

malandis (approved):

PR #236, opened 2023-02-07
chore: swap truncate and ttl arguments

nand4011:
Create interfaces for the optional api arguments. This forces them to be
named and lets them be supplied in any order. Methods with only one
optional argument still get an interface for consistency.

### DIFF
--- a/src/simple-cache-client.ts
+++ b/src/simple-cache-client.ts
@@ -40,8 +40,20 @@ import {
   BackTruncatableCallOptions,
   CollectionCallOptions,
   FrontTruncatableCallOptions,
-  SingleCallOptions,
+  ScalarCallOptions,
 } from './utils/cache-call-options';
+
+// Type aliases to differentiate the different methods' optional arguments.
+type SetOptions = ScalarCallOptions;
+type ListConcatenateBackOptions = FrontTruncatableCallOptions;
+type ListConcatenateFrontOptions = BackTruncatableCallOptions;
+type ListPushBackOptions = FrontTruncatableCallOptions;
+type ListPushFrontOptions = BackTruncatableCallOptions;
+type SetAddElementOptions = CollectionCallOptions;
+type SetAddElementsOptions = CollectionCallOptions;
+type DictionarySetFieldOptions = CollectionCallOptions;
+type DictionarySetFieldsOptions = CollectionCallOptions;
+type DictionaryIncrementOptions = CollectionCallOptions;
 
 /**
  * Momento Simple Cache Client.
@@ -111,7 +123,7 @@ export class SimpleCacheClient {
    * @param {string} cacheName - Name of the cache to store the item in.
    * @param {(string | Uint8Array)} key - The key to set.
    * @param {(string | Uint8Array)} value - The value to be stored.
-   * @param {SingleCallOptions} [options]
+   * @param {SetOptions} [options]
    * @param {number} [options.ttl] - Time to live (TTL) for the item in Cache.
    * This TTL takes precedence over the TTL used when initializing a cache client.
    * @returns {Promise<CacheSet.Response>} - Result of the set operation.
@@ -121,7 +133,7 @@ export class SimpleCacheClient {
     cacheName: string,
     key: string | Uint8Array,
     value: string | Uint8Array,
-    options?: SingleCallOptions
+    options?: SetOptions
   ): Promise<CacheSet.Response> {
     const client = this.getNextDataClient();
     return await client.set(cacheName, key, value, options?.ttl);
@@ -150,7 +162,7 @@ export class SimpleCacheClient {
    * @param {string} cacheName - Name of the cache to store the list in.
    * @param {string} listName - The list to add to.
    * @param {string[] | Uint8Array[]} values - The values to add to the list.
-   * @param {FrontTruncatableCallOptions} [options]
+   * @param {ListConcatenateBackOptions} [options]
    * @param {number} [options.truncateFrontToSize] - If the list exceeds this
    * length, remove excess from the start of the list. Must be positive.
    * @param {CollectionTtl} [options.ttl] - How the TTL should be managed.
@@ -161,7 +173,7 @@ export class SimpleCacheClient {
     cacheName: string,
     listName: string,
     values: string[] | Uint8Array[],
-    options?: FrontTruncatableCallOptions
+    options?: ListConcatenateBackOptions
   ): Promise<CacheListConcatenateBack.Response> {
     const client = this.getNextDataClient();
     return await client.listConcatenateBack(
@@ -180,7 +192,7 @@ export class SimpleCacheClient {
    * @param {string} cacheName - Name of the cache to store the list in.
    * @param {string} listName - The list to add to.
    * @param {string[] | Uint8Array[]} values - The values to add to the list.
-   * @param {BackTruncatableCallOptions} [options]
+   * @param {ListConcatenateFrontOptions} [options]
    * @param {number} [options.truncateBackToSize] - If the list exceeds this
    * length, remove excess from the end of the list. Must be positive.
    * @param {CollectionTtl} [options.ttl] - How the TTL should be managed.
@@ -191,7 +203,7 @@ export class SimpleCacheClient {
     cacheName: string,
     listName: string,
     values: string[] | Uint8Array[],
-    options?: BackTruncatableCallOptions
+    options?: ListConcatenateFrontOptions
   ): Promise<CacheListConcatenateFront.Response> {
     const client = this.getNextDataClient();
     return await client.listConcatenateFront(
@@ -263,7 +275,7 @@ export class SimpleCacheClient {
    * @param {string} cacheName - Name of the cache with the list.
    * @param {string} listName - The list to push to.
    * @param {string | Uint8Array} value - The value to push.
-   * @param {FrontTruncatableCallOptions} [options]
+   * @param {ListPushBackOptions} [options]
    * @param {number} [options.truncateFrontToSize] - If the list exceeds this
    * length, remove excess from the start of the list. Must be positive.
    * @param {CollectionTtl} [options.ttl] - How the TTL should be managed.
@@ -274,7 +286,7 @@ export class SimpleCacheClient {
     cacheName: string,
     listName: string,
     value: string | Uint8Array,
-    options?: FrontTruncatableCallOptions
+    options?: ListPushBackOptions
   ): Promise<CacheListPushBack.Response> {
     const client = this.getNextDataClient();
     return await client.listPushBack(
@@ -291,7 +303,7 @@ export class SimpleCacheClient {
    * @param {string} cacheName - Name of the cache with the list.
    * @param {string} listName - The list to push to.
    * @param {string | Uint8Array} value - The value to push.
-   * @param {BackTruncatableCallOptions} [options]
+   * @param {ListPushFrontOptions} [options]
    * @param {number} [options.truncateBackToSize] - If the list exceeds this
    * length, remove excess from the end of the list. Must be positive.
    * @param {CollectionTtl} [options.ttl] - How the TTL should be managed.
@@ -302,7 +314,7 @@ export class SimpleCacheClient {
     cacheName: string,
     listName: string,
     value: string | Uint8Array,
-    options?: BackTruncatableCallOptions
+    options?: ListPushFrontOptions
   ): Promise<CacheListPushFront.Response> {
     const client = this.getNextDataClient();
     return await client.listPushFront(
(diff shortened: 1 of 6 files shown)


schwern:
Since one can use named arguments, I don't know that the positions of optional arguments matters enough to drift from the other implementations.

>  Ttl needs to be specified less often than the truncate size

How did you come to that conclusion?

> Change the name of the ttl argument to collectionTtl for all methods that take a CollectionTtl object to avoid confusion with methods where the ttl is a number.

I would argue that instead we unify TTL behavior: everything takes a collectionTtl and honors the refresh behavior. I don't see a reason why people wouldn't want to have the refresh behavior on set.

cprice404:
> Since one can use named arguments, I don't know that the positions of optional arguments matters enough to drift from the other implementations.

We will update the other implementations before we ship them.  You can't use named arguments in JS with the signatures that we have today and I don't want to force all of these functions to have property arguments, it's too much unnecessary boilerplate for end users when these last two arguments are likely to be rarely used anyway.  I agree that we should encourage named args in languages that more naturally support it.

> > Ttl needs to be specified less often than the truncate size
> 
> How did you come to that conclusion?

Educated guess based on feedback from @allenheltondev and observing how our current customers are using the truncate functionality.
 
> > Change the name of the ttl argument to collectionTtl for all methods that take a CollectionTtl object to avoid confusion with methods where the ttl is a number.
> 
> I would argue that instead we unify TTL behavior: everything takes a collectionTtl and honors the refresh behavior. I don't see a reason why people wouldn't want to have the refresh behavior on set.

This refresh logic does not exist on the server side for any APIs other than mutative collection APIs.

cprice404 (dismissed):
lgtm - @allenheltondev would you mind scanning this and confirming this re-ordering matches your expectations of which of those two args customers are more likely to want to use?

cprice404:
i talked to @allenheltondev a bit this morning and he suggested a variant of what @schwern said, where we would introduce a props object *just for the optional args*, and leave the positional args in place.  I have seen this in other JS libs and it makes sense to me so we will go ahead and make that change.  @nand4011 i'll ping you.

src/utils/cache-call-options.ts
```diff
@@ -0,0 +1,29 @@
+import {CollectionTtl} from './collection-ttl';
+
+export interface SingleCallOptions {
```
nand4011, line 3: I'm not attached to any of the names in here if anyone has alternatives.

src/simple-cache-client.ts
```diff
@@ -339,20 +353,22 @@ export class SimpleCacheClient {
    * @param {string} cacheName - Name of the cache to store the set in.
    * @param {string} setName - The set to add elements to.
    * @param {(string | Uint8Array)} element - The data to add to the set.
-   * @param {CollectionTtl} [ttl] - TTL for the set in cache. This TTL takes precedence over the TTL used when initializing a cache client. Defaults to client TTL.
+   * @param {CollectionCallOptions} options
+   * @param {number} [options.ttl] - TTL for the set in cache. This TTL takes
```
schwern, line 357:
```suggestion
   * @param {CollectionTtl} [options.ttl] - TTL for the set in cache. This TTL takes
```

src/simple-cache-client.ts
```diff
@@ -365,16 +381,23 @@ export class SimpleCacheClient {
    * @param {string} cacheName - Name of the cache to store the set in.
    * @param {string} setName - The set to add elements to.
    * @param {(string[] | Uint8Array[])} elements - The data to add to the set.
-   * @param {CollectionTtl} [ttl] - TTL for the set in cache. This TTL takes precedence over the TTL used when initializing a cache client. Defaults to client TTL.
+   * @param {CollectionCallOptions} options
+   * @param {number} [options.ttl] - TTL for the set in cache. This TTL takes
```
schwern, line 385:
```suggestion
   * @param {CollectionTtl} [options.ttl] - TTL for the set in cache. This TTL takes
```

src/simple-cache-client.ts
```diff
@@ -467,9 +490,11 @@ export class SimpleCacheClient {
    * After this operation, the set will contain the union of the element passed in and the elements of the set.
    * @param {string} cacheName - Name of the cache to store the dictionary in.
    * @param {string} dictionaryName - The dictionary to set.
-   * @param {string | Uint8Array} items - The field in the dictionary to set.
+   * @param {string | Uint8Array} field - The field in the dictionary to set.
    * @param {string | Uint8Array} value - The value to be stored.
-   * @param {CollectionTtl} ttl -  TTL for the dictionary in cache. This TTL takes precedence over the TTL used when initializing a cache client. Defaults to client TTL.
+   * @param {CollectionCallOptions} options
+   * @param {number} [options.ttl] - TTL for the dictionary in cache. This TTL takes
```
schwern, line 496:
```suggestion
   * @param {CollectionTtl} [options.ttl] - TTL for the dictionary in cache. This TTL takes
```

src/simple-cache-client.ts
```diff
@@ -495,7 +520,9 @@ export class SimpleCacheClient {
    * @param {string} cacheName - Name of the cache to store the dictionary in.
    * @param {string} dictionaryName - The dictionary to set.
    * @param {Map<string | Uint8Array, string | Uint8Array>} items - The field-value pairs in the dictionary to set.
-   * @param {CollectionTtl} ttl -  TTL for the dictionary in cache. This TTL takes precedence over the TTL used when initializing a cache client. Defaults to client TTL.
+   * @param {CollectionCallOptions} options
+   * @param {number} [options.ttl] - TTL for the dictionary in cache. This TTL takes
```
schwern, line 524:
```suggestion
   * @param {CollectionTtl} [options.ttl] - TTL for the dictionary in cache. This TTL takes
```

src/simple-cache-client.ts
```diff
@@ -595,23 +622,25 @@ export class SimpleCacheClient {
    * @param {string} dictionaryName - The dictionary to set.
    * @param {string | Uint8Array} field - Name of the field to increment from the dictionary.
    * @param {number} amount - The quantity to add to the value. May be positive, negative, or zero. Defaults to 1.
-   * @param {CollectionTtl} ttl - TTL for the dictionary in cache. This TTL takes precedence over the TTL used when initializing a cache client. Defaults to client TTL.
+   * @param {CollectionCallOptions} options
+   * @param {number} [options.ttl] - TTL for the dictionary in cache. This TTL takes
```
schwern, line 626:
```suggestion
   * @param {CollectionTtl} [options.ttl] - TTL for the dictionary in cache. This TTL takes
```

src/utils/cache-call-options.ts
schwern: NumberTtlCallOptions?

schwern (changes requested):

src/simple-cache-client.ts
```diff
@@ -252,51 +262,55 @@ export class SimpleCacheClient {
    * Add a value to the beginning of a list.
    * @param {string} cacheName - Name of the cache with the list.
    * @param {string} listName - The list to push to.
-   * @param {string | Uint8Array} - The value to push.
-   * @param {CollectionTtl} [ttl] - How the ttl should be managed. Defaults to the client's TTL.
-   * @param {number} [truncateFrontToSize] - If the list exceeds this length, remove excess from the start of the list. Must be positive.
+   * @param {string | Uint8Array} value - The value to push.
+   * @param {FrontTruncatableCallOptions} [options]
+   * @param {number} [options.truncateFrontToSize] - If the list exceeds this
```
cprice404, line 267: nice!

src/utils/cache-call-options.ts
cprice404: i was hesitant to put the name `Ttl` in the name of any of these because part of the motivation for including these in the first place is that the options list might expand in the future to include stuff other than TTLs.

So my recommendation for this one would probably just be `ScalarCallOptions` but I'm also okay with your current name.

cprice404 (commented):
other than @schwern 's 🦅 👀 docstring catches and possibly renaming that one interface (up to you IMO), I think I'm good!

cprice404:
oh @nand4011 one more thing - I was expecting to see the get-set-delete test modified too, since I would have thought this would have impacted some of the `set` tests?

src/simple-cache-client.ts
```diff
@@ -252,51 +262,55 @@ export class SimpleCacheClient {
...
+    options?: FrontTruncatableCallOptions
   ): Promise<CacheListPushBack.Response> {
     const client = this.getNextDataClient();
     return await client.listPushBack(
       cacheName,
       listName,
       value,
-      ttl,
-      truncateFrontToSize
+      options?.truncateFrontToSize,
```
allenheltondev, line 284: I'm curious why you didn't pass the options through here. When the options grow, this signature is going to grow as well. Just a thought.

src/simple-cache-client.ts
```diff
@@ -252,51 +262,55 @@ export class SimpleCacheClient {
    * Add a value to the beginning of a list.
    * @param {string} cacheName - Name of the cache with the list.
    * @param {string} listName - The list to push to.
-   * @param {string | Uint8Array} - The value to push.
-   * @param {CollectionTtl} [ttl] - How the ttl should be managed. Defaults to the client's TTL.
-   * @param {number} [truncateFrontToSize] - If the list exceeds this length, remove excess from the start of the list. Must be positive.
+   * @param {string | Uint8Array} value - The value to push.
+   * @param {FrontTruncatableCallOptions} [options]
```
allenheltondev, line 266: I'm typically all for DRY programming, but I'm wondering if reusing this type across multiple calls will be intuitive for users. I might think there was a typo in the intellisense if I was trying to autocomplete the params list for `listPopFront` and got prompted with `FrontTruncatableCallOptions`.

I'm asking this open ended wondering what the consensus is of the dev team. I think we either:

* Leave it as is and make sure we document it very well
* Create types for every call

Pro for leaving it as is is that we have fewer classes to manage. Con is that if we add something specific to one call but not the others, we have to introduce a breaking change because the signature would change.

allenheltondev (commented):
Would like input from the dev team on the comments I added to the PR

src/simple-cache-client.ts
cprice404: 2 thoughts on this:

1. I think we actually could make it a backward compatible change because TS/JS interfaces are really flexible; as long as the new interface in the future is a superset of the old one then it shouldn't be a breaking change.
2. I think we could maybe get the best of both worlds by just doing a per-API type alias?  so like:

```typescript
type ListPushFrontOptions = FrontTruncateCallOptions
```

If we just added one line like that above each API then I think the type hints would hopefully give a less weird-sounding name but there's not really anything extra for us to maintain?

src/simple-cache-client.ts
nand4011: Well caught.

src/simple-cache-client.ts
allenheltondev: Big fan of number 2!

nand4011:
> oh @nand4011 one more thing - I was expecting to see the get-set-delete test modified too, since I would have thought this would have impacted some of the `set` tests?

Looks like those don't use the optional ttl argument so they wouldn't be affected by the change.

src/simple-cache-client.ts
nand4011: I wanted to get the first version up so people could see the shape of the options object. I am not opposed to passing it through if that is cleaner.

cprice404:
> > oh @nand4011 one more thing - I was expecting to see the get-set-delete test modified too, since I would have thought this would have impacted some of the `set` tests?
> 
> Looks like those don't use the optional ttl argument so they wouldn't be affected by the change.

yipes, that sounds like missing test coverage then :(  would you mind filing a ticket to capture that?

src/simple-cache-client.ts
nand4011: Truncatable is a word in a math context at least - [https://en.wikipedia.org/wiki/Truncatable_prime](url). It not being in the dictionary is unfortunate.
We would add a new sub-interface if we had a new method that needed an additional optional argument. We couldn't add a new optional argument to an existing signature without either making it available to other methods that use the same options object or changing which one it takes. Is the convenience of not technically breaking backwards compatibility when adding an optional argument worth the overhead of managing per-call options classes? @cprice404

src/simple-cache-client.ts
nand4011: I'm going to keep the internal api as is for now. It won't be set in stone with the 1.0 release since it is internal. I added a ticket to revisit this in the future. [https://github.com/momentohq/client-sdk-nodejs/issues/240](url)

cprice404 (approved):
This LGTM!  would like to get @allenheltondev to take a quick peek if he's still online :)

allenheltondev (approved):
Looks great! Thank you for these changes :)

cprice404:
@schwern i believe @nand4011 addressed your requested changes?

schwern (approved):

PR #244, opened 2023-02-08
chore: update examples to 0.22.0

cprice404:


nand4011 (approved):
Why was the presigned url example deleted?

cprice404:
> Why was the presigned url example deleted?

because you removed the code that it was exercising :)

PR #245, opened 2023-02-08
feat: add some niceties for working with ListFetch/SetFetch response

cprice404:
In some recent commits we added some niceties for working with Dictionary responses.  In particular we added shorthand-aliases for the `*String` accessors, and we added support for Record types because they are more idiomatic to work with in JS than Map types.

Similarly in this commit we:
* Add shorthand-aliases for the `*String` accessors for Lists
* Add shorthand-aliases for the `*String` accessors for Sets
* Add `valueArray*` accessors alongside the `valueSet*` accessors for Sets, because Arrays are easier to work with in JS and more idiomatic for some workflows.

### DIFF
--- a/src/messages/responses/cache-list-fetch.ts
+++ b/src/messages/responses/cache-list-fetch.ts
@@ -27,6 +27,10 @@ class _Hit extends Response {
     return this._values.map(v => TEXT_DECODER.decode(v));
   }
 
+  public valueList(): string[] {
+    return this.valueListString();
+  }
+
   public override toString(): string {
     const truncatedStringArray = truncateStringArray(this.valueListString());
     return `${super.toString()}: [${truncatedStringArray.toString()}]`;
--- a/integration/list.test.ts
+++ b/integration/list.test.ts
@@ -255,6 +255,7 @@ describe('lists', () => {
         await Momento.listFetch(IntegrationTestCacheName, listName)
       );
       expect(respFetch.valueListString()).toEqual([valueString]);
+      expect(respFetch.valueList()).toEqual([valueString]);
       expect(respFetch.valueListUint8Array()).toEqual([valueBytes]);
     });
   });
--- a/integration/set.test.ts
+++ b/integration/set.test.ts
@@ -218,9 +218,14 @@ describe('Integration Tests for operations on sets datastructure', () => {
       setName
     );
     expect(fetchResponse).toBeInstanceOf(CacheSetFetch.Hit);
-    expect((fetchResponse as CacheSetFetch.Hit).valueSetUint8Array()).toEqual(
+    const hit = fetchResponse as CacheSetFetch.Hit;
+    expect(hit.valueSetUint8Array()).toEqual(
       new Set([LOL_BYTE_ARRAY, FOO_BYTE_ARRAY])
     );
+    expect(hit.valueArrayUint8Array()).toEqual([
+      LOL_BYTE_ARRAY,
+      FOO_BYTE_ARRAY,
+    ]);
   });
   it('should succeed for addElements for string arrays happy path', async () => {
     const setName = v4();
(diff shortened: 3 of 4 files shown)


allenheltondev:
Here's a hot take:

Why don't we come to a consensus and decide what we think would be the most used accessor and put it behind a `.value()` function call.

Example: If we think that `valueArray` is going to be used 60% of the time, let's write the code to do this:

```javascript
/**
   * Returns the value of the set formatted as an array
   * @returns {Set<string>}
   */
  public value(): string[] {
    return this.valueArrayString();
  }
```

cprice404:
> Here's a hot take:
> 
> Why don't we come to a consensus and decide what we think would be the most used accessor and put it behind a `.value()` function call.
> 
> Example: If we think that `valueArray` is going to be used 60% of the time, let's write the code to do this:
> 
> ```js
> /**
>    * Returns the value of the set formatted as an array
>    * @returns {string[]}
>    */
>   public value(): string[] {
>     return this.valueArrayString();
>   }
> ```

I like this idea, but I feel like it might take some time to come to consensus for each collection type, and I don't want to block the 1.0 on it.  We can always add that later, right?

allenheltondev:
Fair enough, I will engage the greater community and see what they think. The rest of the changes in this look great.

allenheltondev (approved):

cprice404:
> Fair enough, I will engage the greater community and see what they think. The rest of the changes in this look great.

awesome, thanks!  and thanks for the suggestion, i am definitely willing to consider adding those in the near future

src/messages/responses/cache-list-fetch.ts
```diff
@@ -27,6 +27,10 @@ class _Hit extends Response {
     return this._values.map(v => TEXT_DECODER.decode(v));
   }
 
+  public valueList(): string[] {
```
nand4011, line 30: This would be a bigger change, but do we want to replace the word 'value' with 'element' if we're standardizing on the element language for 1.0? I don't think it reads as nicely, but we should explicitly chose whether to do it.

schwern:
In Python (and probably other languages) we can use type overloading to have a single value method that takes and returns different types. And it can have a default type.

```
@overload
def value(type: str) -> str:
  self._value_bytes.decode('utf-8')

@overload
def value(type: bytes) -> bytes:
  return self._value_bytes

@overload
def value(type: int) -> int: ...

@overload
def value(type: json) -> json: ...
```

TypeScript doesn't allow passing in types, but we likely do something similar with type guards or generics.

src/messages/responses/cache-list-fetch.ts
schwern: I see two considerations.

1. "Hit" responses should all have the same method to get the thing from the cache, regardless of collection type, then users don't have to use differently named methods that all do the same thing depending on what sort of collection they're getting from.
2. Ideally it should be the same term as what the setters use.

You set/add an X and you get an X. Whatever X is.

cprice404: > This would be a bigger change, but do we want to replace the word 'value' with 'element' if we're standardizing on the element language for 1.0? I don't think it reads as nicely, but we should explicitly chose whether to do it.

nope, `value` has been decided and is already locked in in .NET 1.0.  Also IMO still consistent with @malandis 's terminology definitions (search for name-o-rama on Notion).

cprice404: > I see two considerations.
> 
>     1. "Hit" responses should all have the same method to get the thing from the cache, regardless of collection type, then users don't have to use differently named methods that all do the same thing depending on what sort of collection they're getting from.
> 
>     2. Ideally it should be the same term as what the setters use so it's obvious how they relate.
> 
> 
> You set/add an X and you get an X. Whatever X is.

I think this is consistent with Allen's suggestion and we will consider it for future releases.

cprice404:
> In Python (and probably other languages) we can use type overloading to have a single value method that takes and returns different types. And it can have a default type. Something like this...
> 
> ```
> @overload
> def value(type: str) -> str:
>   self._value_bytes.decode('utf-8')
> 
> @overload
> def value(type: bytes) -> bytes:
>   return self._value_bytes
> 
> @overload
> def value(type: int) -> int: ...
> 
> @overload
> def value(type: json) -> json: ...
> 
> num = response.value(type: int)
> ```
> 
> TypeScript doesn't allow passing in types, but we likely do something similar with type guards or generics.
> 
> ```
> declare function value<Type>(type: Type): Type {
>    ...
> }
> ```

good idea, happy to experiment with generics when we attack @allenheltondev 's suggestion in the future.

nand4011 (approved):

PR #246, opened 2023-02-08
fix: remove withLoggerFactory from Configurations

cprice404:
I had forgotten a hard-fought lesson that I learned in the .NET
SDK (which also needs this fix): the pre-built configurations
will often pass the loggerFactory around to the constructors
of other components that they instantiate, such as the RetryInterceptor.

If this happens, and then we allow the user to override the loggerFactory
on the config with a new one, we can end up in a state where the original
factory is being used for some components and the new one for other
components.  For a short time I attempted to reconcile this by trying
to keep track of all of the places where the loggerFactory had been
referenced and then trying to update those as well, but it was a big
mess that wasn't guaranteed to work so we decided that was a bad path
to go down.

Hence, it is necessary for users to specify the one true loggerFactory
(or use the default one) at the time when they construct their
Configuration object.  To that end this commit removes the function
that would allow them to override it.

### DIFF
--- a/src/config/configuration.ts
+++ b/src/config/configuration.ts
@@ -31,13 +31,6 @@ export interface Configuration {
   getLoggerFactory(): MomentoLoggerFactory;
 
   /**
-   * Copy constructor for overriding LoggerOptions
-   * @param {MomentoLoggerFactory} loggerFactory
-   * @returns {Configuration} a new Configuration object with the specified LoggerOptions
-   */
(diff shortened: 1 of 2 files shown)


nand4011 (approved):
Is there ever a legitimate reason to use multiple logger factories?

eaddingtonwhite (approved):

cprice404:
> Is there ever a legitimate reason to use multiple logger factories?

nah, someone could write a custom implementation of our interface that did composite logging if they liked, but would still just be one factory from our perspective.

PR #247, opened 2023-02-08
fix: fix incorrect return types and logging bug

nand4011:
Return remove field(s) responses instead of get field(s) responses for the remove field and remove fields methods.

Log the correct ttl in the internal set method and fix a couple of warnings.


cprice404 (approved):
nice catches, t

PR #269, opened 2023-02-09
chore: remove the page token from the list caches method

nand4011:
Remove the page token from the list caches method. Paging is unimplemented and the pagination could be hidden in the sdk once it is.

Remove an out-of-date TODO.

### DIFF
--- a/src/internal/control-client.ts
+++ b/src/internal/control-client.ts
@@ -115,9 +115,9 @@ export class ControlClient {
     });
   }
 
-  public async listCaches(nextToken?: string): Promise<ListCaches.Response> {
+  public async listCaches(): Promise<ListCaches.Response> {
     const request = new grpcControl._ListCachesRequest();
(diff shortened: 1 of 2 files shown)


cprice404 (approved):
this lgtm but let's double check that a few other folks think this is sane before we merge

cc: @kvcache @schwern @gautamomento

cprice404:
> Remove the page token from the list caches method. Paging is unimplemented and the pagination could be hidden in the sdk once it is.
> 
> Remove an out-of-date TODO.

For context for other folks - what we decided here is to remove the bleed-through of the pagination token from the public API for now.  When we add in server-side support, we can extend this to expose the pagination token and/or have it handle the pagination behind the scenes with a lazy iterator.

kvcache (approved):
I like this, but I asked for it =)

schwern (approved):

PR #270, opened 2023-02-09
chore: Stop the credential provider toString from including the cred

nand4011:
Override the valueOf method in the credential providers to prevent them from including the credential in the returned value.

### DIFF
--- a/src/auth/credential-provider.ts
+++ b/src/auth/credential-provider.ts
@@ -76,6 +76,10 @@ export class StringMomentoTokenProvider implements CredentialProvider {
   getControlEndpoint(): string {
     return this.controlEndpoint;
   }
+
+  public toString(): string {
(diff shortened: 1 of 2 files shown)


src/auth/credential-provider.ts
```diff
@@ -76,6 +76,10 @@ export class StringMomentoTokenProvider implements CredentialProvider {
   getControlEndpoint(): string {
     return this.controlEndpoint;
   }
+
+  public toString(): string {
+    return `Control endpoint: ${this.getControlEndpoint()} Cache endpoint: ${this.getCacheEndpoint()}`;
```
cprice404, line 81: sketched some thoughts here: https://github.com/cprice404/client-sdk-javascript/commit/29e9e9cbe60979c97d1f61bd490d33bb07c36123

will discuss irl

cprice404 (approved):

PR #276, opened 2023-02-10
feat: provide factory functions for instantiating CredentialsProviders

cprice404:
This commit adds some static factory functions for instantiating
CredentialsProviders, to make them a little more discoverable and
user-friendly.

### DIFF
--- a/src/auth/credential-provider.ts
+++ b/src/auth/credential-provider.ts
@@ -21,21 +21,33 @@ interface CredentialProviderProps {
  * @export
  * @interface CredentialProvider
  */
-export interface CredentialProvider {
+export abstract class CredentialProvider {
   /**
    * @returns {string} Auth token provided by user, required to authenticate with the service
    */
-  getAuthToken(): string;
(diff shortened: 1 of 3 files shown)


src/auth/credential-provider.ts
```diff
@@ -21,21 +21,33 @@ interface CredentialProviderProps {
...
+  }
+
+  static fromString(
+    props: StringMomentoTokenProviderProps
+  ): CredentialProvider {
+    return new StringMomentoTokenProvider(props);
+  }
 }
 
 abstract class CredentialProviderBase implements CredentialProvider {
```
nand4011, line 53: Is there any reason to have a separate CredentialProviderBase now that CredentialProvider is an abstract class?

cprice404: probably not, but we can collapse in a follow-up PR right?

nand4011: True.

nand4011 (approved):

PR #277, opened 2023-02-10
WIP: sketch 'hitOrElse'

cprice404:
I don't think this actually works well because of the combinatorial
types due to byte array vs. string.  It's probably better left
as some code we could include in "advanced examples" or something,
and then let the reader implement it in a utils namespace based on
the types they care about.

### DIFF
--- a/src/messages/responses/cache-dictionary-fetch.ts
+++ b/src/messages/responses/cache-dictionary-fetch.ts
@@ -9,8 +9,25 @@ import {TextDecoder} from 'util';
 import {cache_client} from '@gomomento/generated-types/dist/cacheclient';
 
 const TEXT_DECODER = new TextDecoder();
+const TEXT_ENCODER = new TextEncoder();
 
-export abstract class Response extends ResponseBase {}
+export abstract class Response extends ResponseBase {
+  public hitOrElse(fallback: () => Record<string, string>): Hit {
+    if (this instanceof Hit) {
+      return this;
+    }
+    const value: cache_client._DictionaryFieldValuePair[] = Object.entries(
+      fallback()
+    ).map(
+      entry =>
+        new cache_client._DictionaryFieldValuePair({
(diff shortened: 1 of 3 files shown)


src/messages/responses/cache-dictionary-fetch.ts
```diff
@@ -9,8 +9,25 @@ import {TextDecoder} from 'util';
 import {cache_client} from '@gomomento/generated-types/dist/cacheclient';
 
 const TEXT_DECODER = new TextDecoder();
+const TEXT_ENCODER = new TextEncoder();
 
-export abstract class Response extends ResponseBase {}
+export abstract class Response extends ResponseBase {
+  public hitOrElse(fallback: () => Record<string, string>): Hit {
```
cprice404, line 15: @kvcache this is an example of why this gets really messy.  A few things to note here:

1. Because the Hit exposes several different value accessors (Record vs Map, string vs byte array etc.) we don't actually know what return type they want.  It *might* be possible to do it via generics but they would have to declare an explicit type in their call.  If we don't want to push that on them then the best we can do is return an instance of `Hit`.
2. To return an instance of `Hit`, I need to call its constructor, with the most primitive value type.  Which means that I either need to force the user's callback function to return that value type (cache_client._DictionaryFieldPair[] in this case, gross), or I would have to accept some kind of union type of all of the different types they might want to return.  Then I'd have to do a bunch of really tricky reflection on the result of their callback function in order to understand what type they gave me, and convert it back into the _DictionaryFieldValuePair type.  :/

cprice404: alternate flavor here, which might be better:

https://github.com/momentohq/client-sdk-nodejs/pull/278

kvcache: Yeah I would not advocate for coercing a miss into a Hit

PR #278, opened 2023-02-10
WIP: hit or else2

cprice404:
slight different sketch of what hitOrElse could look like.  This one takes two lambdas as args so the caller can choose the return type.

### DIFF
--- a/integration/dictionary.test.ts
+++ b/integration/dictionary.test.ts
@@ -272,6 +272,24 @@ describe('Integration tests for dictionary operations', () => {
         expectedStringStringRecord
       );
       expect(hitResponse.valueRecord()).toEqual(expectedStringStringRecord);
+
+      const hitOrElseValueForHit = response.hitOrElse(
+        h => h.valueRecord(),
+        () => ({foo: 'FOO'})
+      );
+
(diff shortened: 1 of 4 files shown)


integration/dictionary.test.ts
```diff
@@ -272,6 +272,24 @@ describe('Integration tests for dictionary operations', () => {
         expectedStringStringRecord
       );
       expect(hitResponse.valueRecord()).toEqual(expectedStringStringRecord);
+
```
cprice404, line 275: check out the lines below - in this version the hitOrElse takes 2 lambdas as args, which allows the caller to deal with the return types so we don't have to.

src/messages/responses/cache-dictionary-fetch.ts
```diff
@@ -10,7 +10,14 @@ import {cache_client} from '@gomomento/generated-types/dist/cacheclient';
 
 const TEXT_DECODER = new TextDecoder();
 
-export abstract class Response extends ResponseBase {}
+export abstract class Response extends ResponseBase {
```
cprice404, line 13: and here's the implementation, pretty simple.

PR #279, opened 2023-02-10
feat: Implement `hitOrElse` utility function on all hittable responses

cprice404:
Adds a utility function that makes it easier for users to bail if all they care about is the `Hit` response path.


kvcache (approved):
I like the way this user code looks, thanks!

cprice404:
Discussed this with @malandis , we have had a few different opinions over time about baking this pattern into the SDK, so I'm going to hold off on this until after 1.0 so we have more time to consider it carefully.

PR #286, opened 2023-02-14
chore: better link to CLI for getting auth token

cprice404:
During our recent bug bash it was called out that the link to the
momento-cli for getting an auth token pointed to a fairly verbose
doc.  This commit changes the link to point to our getting started
page, where users can choose between the in-browser approach or
the approach of installing the CLI for their OS.


malandis (approved):
:+1:

nand4011 (approved):

PR #289, opened 2023-02-14
chore: change wording of log message when constructing client

cprice404:
Feedback from the bug bash was that using the word "Creating"
would be better than "Instantiating".


kvcache (approved):
Oh yeah, thank you!

malandis (approved):

nand4011 (approved):

PR #291, opened 2023-02-14
chore: use logger in advanced example, make sure all instanceof conditionals are exhaustive

cprice404:


examples/advanced.ts
```diff
@@ -19,6 +19,7 @@ const cacheValue = 'value';
 // you can customize your log level or provide your own logger factory to
 // integrate with your favorite logging framework
 const loggerFactory: MomentoLoggerFactory = new DefaultMomentoLoggerFactory();
+const logger = loggerFactory.getLogger(`AdvancedExample`);
```
malandis, line 22: GitHub complains about using backticks when not interpolating

```suggestion
const logger = loggerFactory.getLogger('AdvancedExample');
```

examples/advanced.ts
```diff
@@ -47,67 +48,75 @@ async function createCacheExample() {
...
-      console.log(
+      logger.info(
         `Concurrent get for ${key} returned ${response.valueString()}`
       );
     } else {
-      console.log(
+      logger.info(
         `Something went wrong with concurrent get for key ${key}: ${response.toString()}`
       );
     }
```
malandis, line 122: Not a blocker but do we want to handle the error case separately from an unknown response?

cprice404: I kind of wanted to illustrate that the user can choose which level they want to handle it at by including some examples of both.  wdyt?

cprice404:
rebased to fix merge conflict.

nand4011 (approved):

examples/advanced.ts
malandis: Sounds good. Just wanted to make sure it wasn't an oversight

malandis (approved):

PR #295, opened 2023-02-14
feat: introduce v1 stable versions of pre-built configs

cprice404:
Users may want to avoid the possibility of us making changes to
the tuning settings in our pre-built configs between releases.
This commit renames the current versions to `v1()`, and converts
the `latest()` factory fns into aliases to `v1()`.  This way
in the future we can introduce v2's if necessary without worrying
about rolling out unexpected/undesired config changes to users
who have pinned to v1.


nand4011 (approved):

malandis (approved):
:sparkles:

PR #296, opened 2023-02-15
chore: add docstrings for response types

cprice404:


src/messages/responses/cache-dictionary-set-fields.ts
```diff
@@ -1,14 +1,48 @@
...
+ * the following subtypes:
+ *
+ * - {Success}
+ * - {Error}
+ *
+ * `instanceof` type guards can be used to operate on the appropriate subtype.
+ * @example
+ * For example:
+ * ```
+ * if (response instanceof CacheDictionarySetField.Error) {
```
malandis, line 16:
```suggestion
 * if (response instanceof CacheDictionarySetFields.Error) {
```

src/messages/responses/cache-dictionary-set-fields.ts
```diff
@@ -1,14 +1,48 @@
...
+ *   // Handle error as appropriate.  The compiler will smart-cast `response` to type
+ *   // `CacheDictionarySetField.Error` in this block, so you will have access to the properties
```
malandis, line 18:
```suggestion
 *   // `CacheDictionarySetFields.Error` in this block, so you will have access to the properties
```

malandis (changes requested):
This is much nicer! One thing for us to consider is if in each example we all pattern matching branches accounted for. Eg in most examples we only have the error response handled. In some cases the success/hit has information a user may care about.

malandis:
> I agree that would be great, but I think this is a good incremental step for now? We can flesh them out further as time permits?

Absolutely

malandis (approved):

PR #298, opened 2023-02-22
chore: add test coverage for set operation with ttl

### LINKED ISSUE #239: Add test coverage for set methods that use a ttl argument.
currently our integration test coverage for `set` doesn't include any ttls, so we aren't verifying that the ttl works properly.

poppoerika:
Added several tests for `set` operation with `options {ttl: some_ttl}` to make sure ttl works as expected.
I referenced the tests in our .NET SDK: https://github.com/momentohq/client-sdk-dotnet/blob/main/tests/Integration/Momento.Sdk.Tests/SimpleCacheDataTest.cs

Closes #239

### DIFF
--- a/integration/get-set-delete.test.ts
+++ b/integration/get-set-delete.test.ts
@@ -174,21 +174,6 @@ describe('get/set/delete', () => {
     }
   });
 
-  it('should return INVALID_ARGUMENT_ERROR for invalid ttl when set with byte key/value', async () => {
-    const setResponse = await Momento.set(
-      IntegrationTestCacheName,
-      new TextEncoder().encode(v4()),
-      new TextEncoder().encode(v4()),
-      {ttl: -1}
-    );
-    expect(setResponse).toBeInstanceOf(CacheSet.Error);
-    if (setResponse instanceof CacheSet.Error) {
-      expect(setResponse.errorCode()).toEqual(
-        MomentoErrorCode.INVALID_ARGUMENT_ERROR
-      );
-    }
-  });
-
   it('should set string key/value with valid ttl and get successfully', async () => {
     const cacheKey = v4();
     const cacheValue = v4();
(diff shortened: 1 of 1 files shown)


nand4011 (approved):

integration/get-set-delete.test.ts
```diff
@@ -157,4 +158,83 @@ describe('get/set/delete', () => {
...
+      new TextEncoder().encode(v4()),
+      {ttl: -1}
+    );
+    expect(setResponse).toBeInstanceOf(CacheSet.Error);
+    if (setResponse instanceof CacheSet.Error) {
+      expect(setResponse.errorCode()).toEqual(
+        MomentoErrorCode.INVALID_ARGUMENT_ERROR
+      );
+    }
+  });
```
schwern, line 190: Key/value conversion is separate from TTL validation. We shouldn't need to test them in combination.

integration/get-set-delete.test.ts
```diff
@@ -157,4 +158,83 @@ describe('get/set/delete', () => {
...
+      cacheValue,
+      {ttl: 15}
+    );
+    expect(setResponse).toBeInstanceOf(CacheSet.Success);
+
+    const getResponse = await Momento.get(IntegrationTestCacheName, cacheKey);
+    if (getResponse instanceof CacheGet.Hit) {
+      expect(getResponse.valueUint8Array()).toEqual(cacheValue);
+    }
+  });
```
schwern, line 224: Same.

integration/get-set-delete.test.ts
poppoerika: I had them only with string key/value to test TTL but when I checked our .NET SDK, we tested all the combinations (string key/value, byte key/value, string key/byte value, etc).
But I don't mind focusing on a specific test criteria here 👍 (which is TTL for this case)

cprice404 (approved):

schwern (approved):

PR #300, opened 2023-02-28
feat: add configurable middleware

cprice404:
This commit adds a Middleware interface.  Users can implement it
to add code that gets executed for each request; can be used for
logging, client-side metrics, etc.


cprice404:
putting this up for early review.  Before it gets merged I want to experiment with some concrete logging/metrics middlewares in the load generator, but i'm hoping this API is pretty much final.

trying to get this out for a customer in the next day or so.

src/config/middleware/logging-middleware.ts
```diff
@@ -0,0 +1,67 @@
...
+export class LoggingMiddlewareRequestHandler
+  implements MiddlewareRequestHandler
+{
+  private readonly logger: MomentoLogger;
+  constructor(logger: MomentoLogger) {
+    this.logger = logger;
+  }
+
+  onRequestMetadata(metadata: Metadata): Promise<Metadata> {
+    this.logger.info(
```
nand4011, line 15: Is this how we will implement trace level logging?

nand4011 (commented):
It's probably too late, but why is the retry interceptor a separate concept from middleware? Could retry be implemented as a middleware?

cprice404:
> 

That's an astute question!  The answer is that I expect a lot more people to be interested in retry than other middleware, so I wanted it to feel more like a first-class construct.

src/config/middleware/logging-middleware.ts
cprice404: yeah that's what i'm thinking, and i may do it in this pr

cprice404:
I think this is ready for review now.  I added another commit, https://github.com/momentohq/client-sdk-nodejs/pull/300/commits/3bee3d4e3ae9c42a0d9b9374bf67ab2f34a3e445 👍 

    feat: add experimental request logging and csv metrics middlewares

    This commit provides some initial experimental middlewares that
    can help us debug customer issues.  They are:

    ExperimentalRequestLoggingMiddleware: logs information about each
    request at various stages of its life cycle, can be used to gather
    timing information etc.

    ExperimentalMetricsCsvMiddleware: produces a CSV file containing
    metrics about each request.  This can be shared with Momento to
    help us analyze performance and other issues on the client.

    A subsequent PR will update the examples to illustrate usage of
    these new middlewares.

Here is an example of what the CSV data looks like:

```
numActiveRequests,requestType,status,startTime,endTime,duration,requestSize,responseSize
100,_SetRequest,0,1677691443726,1677691444480,754,126,2
97,_SetRequest,0,1677691443727,1677691444483,756,126,2
95,_SetRequest,0,1677691443725,1677691444483,758,126,2
93,_SetRequest,0,1677691443726,1677691444483,757,126,2
94,_SetRequest,0,1677691443725,1677691444483,758,126,2
92,_SetRequest,0,1677691443724,1677691444483,759,125,2
91,_SetRequest,0,1677691443728,1677691444483,755,126,2
90,_SetRequest,0,1677691443727,1677691444483,756,126,2
89,_SetRequest,0,1677691443726,1677691444484,758,126,2
88,_SetRequest,0,1677691443727,1677691444484,757,126,2
96,_SetRequest,0,1677691443727,1677691444483,756,126,2
86,_SetRequest,0,1677691443726,1677691444484,758,126,2
85,_SetRequest,0,1677691443726,1677691444484,758,126,2
99,_SetRequest,0,1677691443723,1677691444481,758,125,2
87,_SetRequest,0,1677691443727,1677691444484,757,126,2
84,_SetRequest,0,1677691443726,1677691444484,758,126,2
98,_SetRequest,0,1677691443727,1677691444481,754,126,2
```

cprice404:
I will file a separate PR that shows what it looks like to update the examples to use this.

Also, I think there is still a minor bug in the metrics middleware that I'm looking into.

cprice404:
https://github.com/momentohq/client-sdk-nodejs/pull/300/commits/25ce274580aaba056bbb9e52116230e58fc5d976 fixes a few bugs, I think this is gtg now, will file corresponding examples PR.

cprice404:
update to examples to illustrate use: https://github.com/momentohq/client-sdk-nodejs/pull/302

src/config/middleware/experimental-metrics-csv-middleware.ts
```diff
@@ -0,0 +1,164 @@
...
+  private responseSize: number;
+
+  private receivedResponseBody: boolean;
+  private receivedResponseStatus: boolean;
+
+  constructor(logger: MomentoLogger, csvPath: string) {
+    this.logger = logger;
+    this.csvPath = csvPath;
+    ExperimentalMetricsCsvMiddleware.numActiveRequests++;
+    this.startTime = new Date().getTime();
```
kvcache, line 49: Would you consider using the nodejs performance monitoring toolkit for measuring latency (and maybe more) instead of re-rolling it? https://nodejs.org/api/perf_hooks.html#performancenow

src/internal/grpc/middlewares-interceptor.ts
```diff
@@ -0,0 +1,128 @@
...
+        next: (metadata: Metadata, listener: Listener) => void
+      ): void {
+        const newListener: Listener = {
+          onReceiveMetadata: function (
+            metadata: Metadata,
+            next: (metadata: Metadata) => void
+          ): void {
+            applyMiddlewareHandlers(
+              'onResponseMetadata',
+              middlewareRequestHandlers.reverse(),
```
kvcache, line 37: what is this reverse() for?

kvcache (commented):
Looks good generally. I'm wondering why it doesn't use the nodejs perf hooks api though?

src/config/middleware/experimental-metrics-csv-middleware.ts
cprice404: the reason is because I want to see the actual epoch timestamps; there is a particular customer that I'm trying to debug and I want to be able to process the data from the csv to see how many requests are overlapping, how bursty the traffic is, etc.  I could certainly use their timer to compute the duration (and maybe I will in a follow-up before we make this "non-Experimental"), but since I am already having to compute the start/end timestamps anyway, it didn't seem worth doing it both ways?

src/internal/grpc/middlewares-interceptor.ts
cprice404: when I'm composing the functions I need to do it in reverse order so that they will get applied in the order the user expects; i.e. the first middleware they pass needs to be the last one that I wrap with in order for it to get executed first.  I will double-check this but that's what I recall from my earlier testing.

kvcache: It seems unnatural to me to invert the order - middlewares conceptually "wrap" successively around one another. I'd expect the last thing I added to be the furthest from my application / closest to the wire. (and conversely, the first thing I added I'd expect to be right next to my application)
Is there some pre-existing pattern you're basing this on or some user feedback that they want reversed-wrap ordering instead of the normal grpc style layered ordering? Or maybe I don't understand and this reverse() is necessary to achieve the normal layered ordering?

src/config/middleware/experimental-metrics-csv-middleware.ts
kvcache: yeah, seems to me like probably the best plan here for measuring performance characteristics would be to integrate with the ecosystem's affordances for measuring performance. https://nodejs.org/api/perf_hooks.html#class-performanceobserver It's pretty good stuff, I was impressed reading through some of the stuff they offer in here just now. Seems difficult to replicate some of the relevant stuff in there, for example, [http2 details](https://nodejs.org/api/perf_hooks.html#http2-http2-details) which would be _awesome_ and richly informative if they are preserved by the underlying grpc library. If you want to make this csv thing too, 🤷 sure okay. It's labeled "experimental" like you say.

src/internal/grpc/headers-interceptor.ts
```diff
@@ -15,7 +15,7 @@ export class Header {
   }
 }
 
-export class HeaderInterceptor {
+export class HeaderInterceptorProvider {
```
malandis, line 18: For my benefit, what was the thought behind the rename?

src/internal/grpc/middlewares-interceptor.ts
cprice404: i'm saying that I believe this is how to get it to work the way you are describing, because of how I am composing the functions later.

if I add two middlewares, one that logs "FOO" before the request and one that logs "BAR" before the request, and I add them as `withMiddlewares[logFoo, logBar]`, then intuitively I expect to see the FOO log before the BAR log.  and i'm saying that this is how the code needs to be structured to make the function composition work that way.

as long as we are aligned that "FOO" should be logged before "BAR", i will doublecheck and make 100% sure that this code is yielding that behavior, but I'm almost certain that it is :)

src/config/middleware/experimental-metrics-csv-middleware.ts
```diff
@@ -0,0 +1,164 @@
...
+    const csvRow = [
+      metrics.numActiveRequests,
+      metrics.requestType,
+      metrics.status,
+      metrics.startTime,
+      metrics.endTime,
+      metrics.duration,
+      metrics.requestSize,
+      metrics.responseSize,
+    ].join(',');
```
malandis, line 113: I prefer using a csv library for this to get the quoting and separator issues to spec.

src/config/middleware/experimental-metrics-csv-middleware.ts
```diff
@@ -0,0 +1,164 @@
...
+    fs.appendFile(this.csvPath, `${csvRow}\n`, err => {
```
malandis, line 114: Is it possible to keep the filehandle open for the lifetime of the middleware?

src/internal/grpc/middlewares-interceptor.ts
kvcache: yes, "before the request" logging, I think it should be foobar. For _response_ event logging, I expect barfoo for that withMiddlewares you pasted. If it's just a symmetry issue somewhere downstream structurally then okay this is fine and I'm just confused by reading .reverse() here.

src/config/middleware/experimental-metrics-csv-middleware.ts
```diff
@@ -0,0 +1,164 @@
...
+    'requestSize',
+    'responseSize',
+  ];
+}
+
+interface RequestMetrics {
+  numActiveRequests: number;
+  requestType: string;
+  status: number;
+  startTime: number;
```
malandis, line 24: Thinking ahead: there are a few timestamps we will care about. (1) when the client request was initiated; (2) when the grpc request is scheduled; (3) when the grpc/http request actually goes out.

There could be delays between (1) and (2) because of CPU contention; and there could be delays between (2) and (3) because of too many concurrent http streams.

It looks like this will cover (1)? Maybe (2) if the delta between (1) and (2) isn't a big deal? Just wondering if we have any way to tell when the request actually goes out vs when it is scheduled.

src/config/middleware/experimental-metrics-csv-middleware.ts
```diff
@@ -0,0 +1,164 @@
...
+  endTime: number;
```
malandis, line 25: Also regarding the times: are these datetimes or unix timestamps?

src/internal/grpc/middlewares-interceptor.ts
cprice404: okay, we are aligned on the expected behavior, and I believe that this PR does it exactly that way, but I will double check one more time before merging it and leave a comment on this line to clarify that point after i confirm.

src/config/middleware/experimental-metrics-csv-middleware.ts
cprice404: but then we would be forcing that csv dependency on all of our users and probably causing some of them issues with transitive dependencies.  in this case since we have really tight control over all of these values, this seems preferable.

src/config/middleware/experimental-metrics-csv-middleware.ts
cprice404: i think it may be doing this behind the scenes, but that seems like an api-compatible future optimization if it matters

src/internal/grpc/headers-interceptor.ts
cprice404: because this class doesn't actually implement Interceptor, it just provides a function that will create one and return it.

src/config/middleware/experimental-metrics-csv-middleware.ts
cprice404: check out the `on*` callbacks that are defined [here](https://github.com/momentohq/client-sdk-nodejs/pull/300/files#diff-a5d135b804012d314a7ce8a6b8c5627c5fb8825c63aad84e7fd97fb0e651e9f8R4).  Those are the hooks that I have available from the Interceptor AFAIK.

I could theoretically add a timestamp that gets captured in `onRequestBody`, but I don't know if we have any guarantee of what time that happens out of (1), (2), and (3) on your list, and I don't think I have any other hooks.

That said, if you think there's a chance that timestamp might be valuable it's easy enough to add.  LMK if you have a suggestion for a name :)  `requestBodyTime`?

malandis: No suggestions on a name 😬

src/config/middleware/experimental-metrics-csv-middleware.ts
cprice404: thanks for bringing this to my attention.  we can definitely play with that for a future iteration.  this version is just to try to get us started on debugging an active customer issue and see what things are helpful and which aren't, but yeah i agree that's worth looking into for the next go-around.

kvcache: Works for me!

src/internal/grpc/middlewares-interceptor.ts
kvcache: And works for me too! Thanks for confirming the expected vs. actual behavior!

cprice404:
Added a commit: https://github.com/momentohq/client-sdk-nodejs/pull/300/commits/b0fb9f0d4f40e1dbae04a851a8f16ab854d087fa

@kvcache this fixes a bug in the way the middleware reverse was working.  I didn't realize that function modified the array in place.  I manually verified that the order of the callbacks matches what you and I both stated as our expectation above after this commit.

@malandis this adds that extra timestamp you asked for.

kvcache (dismissed):
yup, makes sense then if the order's right. I thought `Array::reverse()` did it in place but it was letting you use it as  an expression value so I thought I was maybe just mistaken. Anyway, glad it's sorted.

malandis (approved):

cprice404:
rebased onto latest main

cprice404:
> yup, makes sense then if the order's right. I thought `Array::reverse()` did it in place but it was letting you use it as an expression value so I thought I was maybe just mistaken. Anyway, glad it's sorted.

yeah thanks for calling it out, i totally missed that.

malandis (approved):

PR #314, opened 2023-03-03
feat: sorted set put value and fetch by index implementation

malandis:
Implements `sortedSetPutValue` and `sortedSetFetchByIndex`.


src/internal/cache-client.ts
```diff
@@ -1640,6 +1643,172 @@ export class CacheClient {
     });
   }
 
+  public async sortedSetPutValue(
```
cprice404, line 1646: for similarity with Sets we may wanna go with Element here.

src/internal/cache-client.ts
```diff
@@ -1640,6 +1643,172 @@ export class CacheClient {
...
+    endIndex?: number
+  ): Promise<CacheSortedSetFetch.Response> {
+    try {
+      validateCacheName(cacheName);
+      validateSortedSetName(sortedSetName);
+    } catch (err) {
+      return new CacheSortedSetFetch.Error(normalizeSdkError(err as Error));
+    }
+    //this.logger.trace(
+    //`Issuing 'sortedSetFetchByIndex' request; startIndex: ${startIndex}, endIndex : ${endIndex}`
```
cprice404, line 1727: can use the logger varargs signature here to avoid the linting error

cprice404 (approved):

src/internal/cache-client.ts
malandis: Fixed in https://github.com/momentohq/client-sdk-nodejs/pull/314/commits/3d63768766c96de67c2e103e8705369a6b28a6c2

cprice404 (approved):

src/internal/cache-client.ts
malandis: Fixed in [49daeee](https://github.com/momentohq/client-sdk-nodejs/pull/314/commits/49daeeef75b3f569d3db576b7bf01e5ff5ae2923)

cprice404 (dismissed):

cprice404 (approved):

PR #317, opened 2023-03-03
feat: adds sortedSetIncrementScore function

cprice404:
This commit wires up the request/response logic for sortedSetIncrementScore. No tests yet because I didn't want to cause merge conflicts in the test file.

### DIFF
--- a/src/simple-cache-client.ts
+++ b/src/simple-cache-client.ts
@@ -40,6 +40,7 @@ import {
   CacheSortedSetFetch,
   MomentoLogger,
   CacheSortedSetGetRank,
+  CacheSortedSetIncrementScore,
 } from '.';
 import {range} from './internal/utils/collections';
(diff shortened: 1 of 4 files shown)


src/simple-cache-client.ts
```diff
@@ -913,13 +915,52 @@ export class SimpleCacheClient {
   public async sortedSetGetRank(
     cacheName: string,
     sortedSetName: string,
-    value: string | Uint8Array
+    elementValue: string | Uint8Array
```
malandis, line 918: value vs elementValue?

malandis (commented):
Looks good. Just have ? about the naming we landed on

src/simple-cache-client.ts
cprice404: ah thanks will fix

PR #318, opened 2023-03-03
feat: use call options for sorted set fetch, including order

malandis:
For `sortedSetFetchByInde`: refactors the start and index index to be optional and part of call options. Also adds `order` to the call options.

### DIFF
--- a/src/utils/cache-call-options.ts
+++ b/src/utils/cache-call-options.ts
@@ -27,3 +27,14 @@ export interface BackTruncatableCallOptions extends CollectionCallOptions {
    */
   truncateBackToSize?: number;
 }
+
+export enum SortedSetOrder {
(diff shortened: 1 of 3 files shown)


src/utils/cache-call-options.ts
```diff
@@ -27,3 +27,14 @@ export interface BackTruncatableCallOptions extends CollectionCallOptions {
...
+export enum SortedSetOrder {
+  Ascending = 'ASC',
+  Descending = 'DESC',
+}
+
+export interface SortedSetFetchByIndexCallOptions {
+  startIndex?: number;
+  endIndex?: number;
+  order?: SortedSetOrder;
+}
```
cprice404, line 40: nice

cprice404 (approved):

PR #320, opened 2023-03-03
feat: implement sorted set fetch by score

malandis:
Implements `sortedSetFetchByScore`. A future PR will implement the other features of this API, including min/max inclusive vs exclusive flags, limit, and count; as well as testing.


cprice404:
> Implements `sortedSetFetchByScore`. A future PR will implement the other features of this API, including min/max inclusive vs exclusive flags, limit, and count; as well as testing.

we decided not to expose those inclusive/exclusive flags unless we get a request.

cprice404 (approved):
looks great!  thanks

PR #322, opened 2023-03-04
feat: Add listRetain and allow slices for listFetch

ben-kugler:


src/simple-cache-client.ts
```diff
@@ -403,6 +417,38 @@ export class SimpleCacheClient {
     return await client.listRemoveValue(cacheName, listName, value);
   }
 
+  /**
+   * Retains slice of elements of a given list.
```
cprice404, line 421: I might expand on this a bit further to try to make it really clear that the rest of the list gets removed on the server-side.  Just to make it clear that we're not just retrieving the slice and leaving the server-side list unmodified.

test/integration/list.test.ts
```diff
@@ -450,6 +477,58 @@ describe('lists', () => {
...
+
+      const respFetch = <CacheListFetch.Hit>(
+        await Momento.listFetch(IntegrationTestCacheName, listName)
+      );
+
+      expect(respFetch.valueListString()).toEqual(valueStringExpected);
+      expect(respFetch.valueList()).toEqual(valueStringExpected);
+      expect(respFetch.valueListUint8Array()).toEqual([valueBytesExpected]);
+    });
+  });
```
cprice404, line 530: these look good to me but curious if @malandis has ideas for any other test cases that we need to cover

src/simple-cache-client.ts
```diff
@@ -403,6 +417,38 @@ export class SimpleCacheClient {
...
+   * @param {ListRetainIndexOptions} [options]
+   * @param {number} [options.startIndex] - Start inclusive index for fetch
+   * operation. Defaults to start of array if not given, 0.
+   * @param {number} [options.endIndex] - End exclusive index for fetch
+   * operation. Defaults to end of array if not given.
+   * @param {CollectionTtl} [options.ttl] - How the TTL should be managed.
+   * Refreshes the list's TTL using the client's default if this is not
+   * supplied.
+   * @returns {Promise<CacheListRetain.Response>} -
+   * {@link CacheListRetain.Hit} containing the list elements if the list exists.
```
ben-kugler, line 434: This return type is wrong, will update, should be `Success`

src/internal/cache-client.ts
```diff
@@ -716,6 +743,81 @@ export class CacheClient {
...
+      validateListName(listName);
+    } catch (err) {
+      return new CacheListRetain.Error(normalizeSdkError(err as Error));
+    }
+    this.logger.trace(
+      `Issuing 'listRetain' request;
+      listName: ${listName},
+      startIndex: ${startIndex ?? 'null'},
+      endIndex: ${endIndex ?? 'null'},
+      ${ttl.toString()}`
```
malandis, line 764: Can you use a format string instead of interpolating the values? That is a pattern we've adopted elsewhere to avoid the extra compute when the level isn't trace.

src/internal/cache-client.ts
```diff
@@ -673,27 +676,51 @@ export class CacheClient {
...
       return new CacheListFetch.Error(normalizeSdkError(err as Error));
     }
-    this.logger.trace(`Issuing 'listFetch' request; listName: ${listName}`);
-    const result = await this.sendListFetch(cacheName, this.convert(listName));
+    this.logger.trace(
+      `Issuing 'listFetch' request;
+      listName: ${listName},
+      startIndex: ${startIndex ?? 'null'},
+      endIndex: ${endIndex ?? 'null'}`
+    );
```
malandis, line 694: Same thing here re: string interpolation

src/internal/cache-client.ts
```diff
@@ -673,27 +676,51 @@ export class CacheClient {
 
   public async listFetch(
     cacheName: string,
-    listName: string
+    listName: string,
+    startIndex?: number,
+    endIndex?: number
```
malandis, line 681: Can you also add a call in the try block to validate the start and end indexes? See @cprice404 's sorted set fetch PR as a reference.

src/internal/cache-client.ts
```diff
@@ -716,6 +743,81 @@ export class CacheClient {
     });
   }
 
+  public async listRetain(
+    cacheName: string,
+    listName: string,
+    startIndex?: number,
+    endIndex?: number,
```
malandis, line 750: Same thing here re: validating the start and end indexes

test/integration/list.test.ts
malandis: I left some suggestions in the review.

malandis (commented):
Few minor changes recommended. Looks good otherwise.

cprice404:
@ben-kugler any idea what the test failures are about?

malandis (approved):

PR #325, opened 2023-03-06
feat: implement `sortedSetGetScores`

malandis:
Implements `sortedSetGetScores` and refactors `sortedSetGetScore` to
delegate to the former.

### DIFF
--- a/src/messages/responses/cache-sorted-set-get-scores.ts
+++ b/src/messages/responses/cache-sorted-set-get-scores.ts
@@ -105,7 +105,7 @@ class _Hit extends Response {
    * This can be used in most places where an Object is desired.
    * @returns {Record<string, number>}
    */
-  public valueRecordStringNumber(): Record<string, number> {
+  public valueRecordString(): Record<string, number> {
     return this.responses.reduce<Record<string, number>>((acc, response) => {
       if (response instanceof CacheSortedSetGetScoreResponse.Hit) {
         acc[response.valueString()] = response.score();
@@ -121,7 +121,7 @@ class _Hit extends Response {
    * @returns {Record<string, number>}
    */
   public valueRecord(): Record<string, number> {
-    return this.valueRecordStringNumber();
+    return this.valueRecordString();
   }
 
   public override toString(): string {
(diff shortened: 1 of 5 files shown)


src/messages/responses/cache-sorted-set-get-scores.ts
```diff
@@ -0,0 +1,169 @@
...
+  public valueMap(): Map<string, number> {
+    return this.valueMapString();
+  }
+
+  /**
+   * Returns the data as a Record whose keys and values are utf-8 strings, decoded from the underlying byte arrays.
+   * This can be used in most places where an Object is desired.
+   * @returns {Record<string, number>}
+   */
+  public valueRecordStringNumber(): Record<string, number> {
```
cprice404, line 108: do we wanna remove Number here to be consistent?

src/simple-cache-client.ts
```diff
@@ -984,7 +985,25 @@ export class SimpleCacheClient {
...
+   * @param {(string | Uint8Array)[]} values - The values of the elements whose scores we are retrieving.
+   * @returns {Promise<CacheSortedSetGetScores.Response>}
+   * {@link CacheSortedGetScores.Hit} containing the scores of the requested elements when found.
+   * {@link CacheSortedGetScores.Miss} when the element or collection does not exist.
+   * {@link CacheSortedGetScores.Error} on failure.
+   */
+  public async sortedSetGetScores(
+    cacheName: string,
+    sortedSetName: string,
+    values: (string | Uint8Array)[]
```
cprice404, line 1002: in most places we have been doing `string[] | Uint8Array[]`, which I think is different from this; we should probably stick to the other one?

cprice404:
@malandis if you agree with those suggestions above just lmk and I can update this PR while you work on other tasks.

src/messages/responses/cache-sorted-set-get-scores.ts
malandis: Yes absolutely. This slipped my mind

malandis: Fixed in [358a0c6](https://github.com/momentohq/client-sdk-nodejs/pull/325/commits/358a0c6214074f9eeb4569fbc3e923102b673afb)

src/simple-cache-client.ts
malandis: I prefer the broader one, though it may be confusing unless we change the rest. So we can change to the more restricted one.

malandis:
> @malandis if you agree with those suggestions above just lmk and I can update this PR while you work on other tasks.

👍

cprice404 (approved):

PR #326, opened 2023-03-06
feat: add sortedSetPutElements

cprice404:


malandis:
Looks great. I've been in the habit of re-writing the singulars to delegate to the plurals. Can you make that change here to `sortedSetPutElement`?

cprice404:
> Looks great. I've been in the habit of re-writing the singulars to delegate to the plurals. Can you make that change here to `sortedSetPutElement`?

can we do that as a follow-up so that we don't block getting this API in place for early consumers?

malandis (approved):

PR #327, opened 2023-03-07
fix: tests for sortedSetFetchByIndex, minor naming fixes

cprice404:
Tweaks a few of the accessor names based on recent conversations.
Adds tests for sortedSetFetchByIndex, including validation for
a few possible error cases for startIndex and endIndex.

### DIFF
--- a/src/internal/cache-client.ts
+++ b/src/internal/cache-client.ts
@@ -62,7 +62,8 @@ import {
   validateSetName,
   validateSortedSetName,
   validateSortedSetOffset,
-  validateSortedSetCount, validateSortedSetIndices,
+  validateSortedSetCount,
+  validateSortedSetIndices,
 } from './utils/validators';
 import {SimpleCacheClientProps} from '../simple-cache-client-props';
 import {Middleware} from '../config/middleware/middleware';
--- a/src/internal/cache-client.ts
+++ b/src/internal/cache-client.ts
@@ -62,7 +62,7 @@ import {
   validateSetName,
   validateSortedSetName,
   validateSortedSetOffset,
-  validateSortedSetCount,
+  validateSortedSetCount, validateSortedSetIndices,
 } from './utils/validators';
 import {SimpleCacheClientProps} from '../simple-cache-client-props';
 import {Middleware} from '../config/middleware/middleware';
@@ -1807,6 +1807,7 @@ export class CacheClient {
     try {
       validateCacheName(cacheName);
       validateSortedSetName(sortedSetName);
+      validateSortedSetIndices(startIndex, endIndex);
     } catch (err) {
       return new CacheSortedSetFetch.Error(normalizeSdkError(err as Error));
     }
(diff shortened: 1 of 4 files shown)


test/integration/sorted-set.test.ts
```diff
@@ -187,174 +186,434 @@ describe('Integration tests for sorted set operations', () => {
...
+          sortedSetName,
+          {
+            startIndex: 10,
+          }
+        );
+
+        expect(response).toBeInstanceOf(CacheSortedSetFetch.Hit);
+        const hitResponse = response as CacheSortedSetFetch.Hit;
+        expect(hitResponse.valueArray()).toEqual([]);
+      });
```
cprice404, line 351: @wandaitzuchen - I don't need you to review this whole PR, just this one test.

This test is failing; we are getting back a `.missing` value on the response object.  I feel like that is not the right behavior?  I feel like this should return a Hit with an empty list of values.  WDYT?

wandaitzuchen: Actually in the backend we do return a miss if we got an empty list from redis (redis response does not differentiate between sorted set not exist or fetching empty list from existing sorted set). Unlike sorted set get rank or sorted set get score, where `_SortedSetFound` is returned to include `ECacheResult`, sorted set fetch return found only when there's elements in the response.  

I agree that we should unify the behavior and return a found with empty list if the set does exist.

src/internal/cache-client.ts
```diff
@@ -62,7 +62,7 @@ import {
   validateSetName,
   validateSortedSetName,
   validateSortedSetOffset,
-  validateSortedSetCount,
+  validateSortedSetCount, validateSortedSetIndices,
```
malandis, line 65:
```suggestion
  validateSortedSetCount,
  validateSortedSetIndices,
```

src/internal/utils/validators.ts
```diff
@@ -18,6 +18,23 @@ export function validateSortedSetName(name: string) {
...
+  start_index: number,
+  end_index?: number
+) {
+  if (end_index === undefined) {
+    return;
+  }
+  if (start_index > 0 && end_index > 0 && start_index >= end_index) {
+    throw new InvalidArgumentError('start index must be less than end index');
+  }
+  if (start_index < 0 && end_index < 0 && start_index >= end_index) {
```
malandis, line 31: What if start and end are both 0?

malandis (commented):
Thank you for traversing this ordeal of a test suite.

src/internal/utils/validators.ts
cprice404: fair

cprice404: hrm actually what are you asking for here?  end index is exclusive so 0,0 wouldn't really make a lot of sense would it?

malandis: I thought you were checking for runtime zero-length spans as well. If that's not an issue then we don't need to check for start == end

malandis (approved):

PR #328, opened 2023-03-07
fix: add tests for sortedSetFetchByScore, fix minor issue with 0 min/max scores

cprice404:
Adds tests for sortedSetFetchByScore.  Fixes a minor issue where if min/max score were explicitly set to 0, we weren't passing them through to the request because they evaluated as "false-y" when we checked to see if they had been provided.

### DIFF
--- a/src/internal/cache-client.ts
+++ b/src/internal/cache-client.ts
@@ -64,6 +64,7 @@ import {
   validateSortedSetOffset,
   validateSortedSetCount,
   validateSortedSetIndices,
+  validateSortedSetScores,
 } from './utils/validators';
 import {SimpleCacheClientProps} from '../simple-cache-client-props';
 import {Middleware} from '../config/middleware/middleware';
@@ -1922,6 +1923,7 @@ export class CacheClient {
     try {
       validateCacheName(cacheName);
       validateSortedSetName(sortedSetName);
+      validateSortedSetScores(minScore, maxScore);
       if (offset !== undefined) {
         validateSortedSetOffset(offset);
       }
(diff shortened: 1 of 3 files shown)


test/integration/sorted-set.test.ts
```diff
@@ -616,6 +616,487 @@ describe('Integration tests for sorted set operations', () => {
...
+          sortedSetName,
+          {
+            minScore: 2_000_000,
+          }
+        );
+
+        expect(response).toBeInstanceOf(CacheSortedSetFetch.Hit);
+        const hitResponse = response as CacheSortedSetFetch.Hit;
+        expect(hitResponse.valueArray()).toEqual([]);
+      });
```
cprice404, line 781: @wandaitzuchen - same as previous PR - I don't necessarily need you to review this, but there are 3 tests that I added in this PR that are asserting that we get back a Hit with an empty list of values, but we are actually getting back a Miss.

src/internal/cache-client.ts
```diff
@@ -1967,7 +1968,7 @@ export class CacheClient {
     count?: number
   ): Promise<CacheSortedSetFetch.Response> {
     const by_score = new grpcCache._SortedSetFetchRequest._ByScore();
-    if (minScore) {
+    if (minScore !== undefined) {
```
malandis, line 1971: This is because `0` evaluates to `false` in JS, correct?

test/integration/sorted-set.test.ts
```diff
@@ -619,6 +619,496 @@ describe('Integration tests for sorted set operations', () => {
...
+        const hitResponse = response as CacheSortedSetFetch.Hit;
+        expect(hitResponse.valueArray()).toEqual([
+          {value: 'bam', score: 1000},
+          {value: 'burrito', score: 9000},
+          {value: 'taco', score: 90210},
+        ]);
+      });
+
+      // Skipping this for now; need to re-enable:
+      // https://github.com/momentohq/client-sdk-nodejs/issues/330
```
malandis, line 932: 🎗️

malandis (approved):
Looks good; just need the server side fix to update the tests.

malandis (approved):

PR #336, opened 2023-03-08
feat: support passing in legacy auth tokens and v1 auth tokens

bruuuuuuuce:
we are adding support for new, programmatically generated auth tokens. These tokens will be base64 encoded json strings of the format

```
{ "api_key": "string", "endpoint": "string" }
```

This commit adds logic within our token parser to try and parse these new tokens, and fallback to our older legacy token parsing logic. I have renamed everything `jwt` to `auth`, since we are no longer only dealing with jwt's. Since we aren't only dealing with jwt's anymore, `decodeAuthToken` now returns a set interface, instead of the claims within the jwt. This change should be fully backwards compatible with previous versions of our sdk.

closes https://github.com/momentohq/client-sdk-nodejs/issues/337


cprice404 (dismissed):
this LGTM but this is a very important change so I would like to leave the PR open for a bit and let other folks from dev eco get a chance to look at it, so they know what's coming.

src/internal/utils/auth.ts
```diff
@@ -0,0 +1,73 @@
...
+ * @param {string=} token
+ * @returns TokenAndEndpoints
+ */
+export const decodeAuthToken = (token?: string): TokenAndEndpoints => {
+  if (!token) {
+    throw new InvalidArgumentError('malformed auth token');
+  }
+
+  try {
+    if (isBase64(token)) {
```
rlinehan, line 53: I wonder if a comment here about why we're checking for base64 encoding would be useful?

src/internal/utils/auth.ts
```diff
@@ -0,0 +1,73 @@
...
+  c: string;
+}
+
+interface V1ApiTokenClaims {
+  sub: string;
+  ver: number;
+  exp: number;
+}
+
+export interface V1ApiTokenClaimsWithEndpoint extends V1ApiTokenClaims {
```
rlinehan, line 23: Is this used anywhere?

test/unit/utils/auth.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    expect(res.controlEndpoint).toEqual('control plane endpoint');
+    expect(res.cacheEndpoint).toEqual('data plane endpoint');
+    expect(res.authToken).toEqual(TEST_LEGACY_AUTH_TOKEN_ENDPOINT);
+
+    const noEndpoints = decodeAuthToken(TEST_LEGACY_AUTH_TOKEN_NO_ENDPOINT);
+    expect(noEndpoints.authToken).toEqual(TEST_LEGACY_AUTH_TOKEN_NO_ENDPOINT);
+    expect(noEndpoints.controlEndpoint).toBeUndefined();
+    expect(noEndpoints.cacheEndpoint).toBeUndefined();
+  });
+  it('should parse endpoints from v1 auth token', () => {
```
rlinehan, line 28: What about a test for a base64 encoded token that doesn't include `api_token` and `endpoint`? e.g. ways that a v1 token might be invalid

rlinehan (commented):
I left a couple small comments, but generally this looks good to me. But I agree with Chris - getting other dev eco folks to review would be really good (I haven't been in this code much at all).

test/unit/utils/auth.test.ts
bruuuuuuuce: https://github.com/momentohq/client-sdk-nodejs/pull/336/commits/eeebeddaade055970291a058ee31981e65436625

src/internal/utils/auth.ts
bruuuuuuuce: removed in https://github.com/momentohq/client-sdk-nodejs/pull/336/commits/eeebeddaade055970291a058ee31981e65436625

src/internal/utils/auth.ts
bruuuuuuuce: added one in https://github.com/momentohq/client-sdk-nodejs/pull/336/commits/eeebeddaade055970291a058ee31981e65436625

src/internal/utils/auth.ts
```diff
@@ -0,0 +1,70 @@
...
+ * @param {string=} token
+ * @returns TokenAndEndpoints
+ */
+export const decodeAuthToken = (token?: string): TokenAndEndpoints => {
+  if (!token) {
+    throw new InvalidArgumentError('malformed auth token');
+  }
+
+  try {
+    // our v1 api tokens don't have an endpoint as part of their claims. Instead, when we give them to a customer, we
```
bruuuuuuuce, line 42:
```suggestion
    // v1 api tokens don't have an endpoint as part of their claims. Instead, when we give them to a customer, we
```

cprice404:
@pgautier404 whenever you are back to work would you mind taking a look at this?

test/unit/utils/auth.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    const decodedToken = {
+      api_key: 'key',
+    };
+    const invalidBase64Token = encodeToBase64(JSON.stringify(decodedToken));
+    expect(() => decodeAuthToken(invalidBase64Token)).toThrow(
+      InvalidArgumentError
+    );
+  });
+
+  it('should throw InvalidJwtError when base64 token is missing endpoint', () => {
```
pgautier404, line 60:
```suggestion
  it('should throw InvalidJwtError when base64 token is missing api key', () => {
```

pgautier404:
Looks great to me apart from the one test description I commented on.

pgautier404 (approved):

PR #339, opened 2023-03-09
feat: rename SimpleCacheClient to CacheClient

cprice404:
This commit renames the primary client object from SimpleCacheClient
to CacheClient, after we received feedback from users that the
name was misleading and seemed to imply that there were multiple
clients to choose from.

The legacy name is retained as a type alias with a deprecation
warning so that this change is backward-compatible.

### DIFF
--- a/README.template.md
+++ b/README.template.md
@@ -87,13 +87,13 @@ if (getResponse instanceof CacheGet.Hit) {
 
 ### Error Handling
 
-Errors that occur in calls to `SimpleCacheClient` methods are surfaced to developers as part of the return values of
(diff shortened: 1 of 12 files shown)


malandis (approved):

cprice404:
> We also have to update any README for text references to `SimpleCacheClient`, eg
> 
> https://github.com/momentohq/client-sdk-nodejs/blob/main/README.md#error-handling

done in https://github.com/momentohq/client-sdk-nodejs/pull/339/commits/20db978ac2b4d23657048ae746777f9ab3a90aca

malandis (approved):

PR #342, opened 2023-03-10
fix: improve metrics for num active requests

cprice404:
We now capture the number of active requests both at the start
and the end of a given request, because it may change a lot
while the request is being processed.

### DIFF
--- a/src/config/middleware/experimental-metrics-csv-middleware.ts
+++ b/src/config/middleware/experimental-metrics-csv-middleware.ts
@@ -20,15 +20,45 @@ function headerFields(): Array<string> {
 }
 
 interface RequestMetrics {
+  /**
+   * number of requests active at the start of the request
+   */
   numActiveRequestsAtStart: number;
+  /**
+   * number of requests active at the finish of the request (including the request itself)
+   */
   numActiveRequestsAtFinish: number;
+  /**
+   * The generated grpc object type of the request
+   */
   requestType: string;
+  /**
+   * The grpc status code of the response
+   */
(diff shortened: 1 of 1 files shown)


src/config/middleware/experimental-metrics-csv-middleware.ts
```diff
@@ -95,7 +99,9 @@ class ExperimentalMetricsCsvMiddlewareRequestHandler
   private recordMetrics(): void {
     const endTime = new Date().getTime();
     const metrics: RequestMetrics = {
-      numActiveRequests: ExperimentalMetricsCsvMiddleware.numActiveRequests,
+      numActiveRequestsAtStart: this.numActiveRequestsAtStart,
+      numActiveRequestsAtFinish:
+        ExperimentalMetricsCsvMiddleware.numActiveRequests,
```
malandis, line 104: Are we including the current request as still active at this point?

cprice404: i would say yes but I don't feel strongly, i can change it if you have a different pref

malandis: Can just document that `numActiveRequestsAtFinish` includes it the current request

cprice404: https://github.com/momentohq/client-sdk-nodejs/pull/342/commits/02da4b3b9c0553fdf3b957d636ffd401dea32e61

malandis (approved):

PR #344, opened 2023-03-13
chore: Add request coalescer example

rishtigupta:
### Revision3
- Addressed comment feedback 

### Revision2
1. Add request coalescing class that implements gets and sets from the map (with keys as cacheKeys and values as cache response promises) (Thanks Chris for helping me out!)
2. Add metric to record number of get/set requests coalesced

### Revision1 
1. Add request coalescing example
    - Runs concurrent requests(~200) using the basic load-generator
    - Runs conscurrent requests(~200) using the request-coleascer-load-generator
    - Prints the histogram and some stats for both the cases as they run for 60 sec. 
1. Refactor overlapping pieces of code to utils

## Issue:
https://github.com/momentohq/dev-eco-issue-tracker/issues/204

## Run the example
1. `npm install`
4. `MOMENTO_AUTH_TOKEN=<YOUR AUTH TOKEN> npm run request-coalescing`

## Stats:
For 200 concurrent requests, ran for 60 sec with 100 requests (max) allowed per sec
- Basic load-gen stats:
BasicLoadGen

- Request Coalescer load-gen stats:
RequestCoalescer

More stats when configuration/parameters are tweaked can be found in this Notion wiki - https://momentohq.notion.site/Request-Coalescing-Stats-7c4efeeab9d448538647712e3d7e1dff

## Stats Inference:
Observe how the latencies have significantly reduced when using request-coalescer-load-gen over basic load-gen.

### DIFF
--- a/examples/request-coalescing.ts
+++ b/examples/request-coalescing.ts
@@ -1,5 +1,4 @@
 import {
-  CacheClient,
   DefaultMomentoLoggerFactory,
   DefaultMomentoLoggerLevel,
   MomentoLogger,
@@ -12,14 +11,26 @@ import {
   BasicLoadGenOptions,
   executeRequestAndUpdateContextCounts,
   initiateLoadGenContext,
+  initiateRequestCoalescerContext,
 } from './utils/load-gen';
 import {
   getElapsedMillis,
+  logCoalescingStats,
   logStats,
 } from './utils/load-gen-statistics-calculator';
 import {createCache, getCacheClient} from './utils/cache';
+import {
+  GetAndSetOnlyClient,
+  MomentoClientWrapperWithCoalescing,
+} from './utils/momento-client-with-coalescing';
 
-class BasicLoadGen {
+const cacheKeys: string[] = [];
+
+for (let i = 0; i < 10; i++) {
+  cacheKeys.push(`cacheKey${i}`);
+}
+
+class RequestCoalescerLoadGen {
   private readonly loggerFactory: MomentoLoggerFactory;
   private readonly logger: MomentoLogger;
   private readonly cacheItemTtlSeconds = 60;
@@ -27,11 +38,11 @@ class BasicLoadGen {
   private readonly delayMillisBetweenRequests: number;
   private readonly cacheValue: string;
 
-  private readonly cacheName: string = 'js-loadgen';
+  private readonly cacheName: string = 'js-requestcoalescer';
 
   constructor(options: BasicLoadGenOptions) {
     this.loggerFactory = options.loggerFactory;
-    this.logger = this.loggerFactory.getLogger('load-gen');
+    this.logger = this.loggerFactory.getLogger('request-coalescer-load-gen');
     this.options = options;
     this.cacheValue = 'x'.repeat(options.cacheItemPayloadBytes);
     this.delayMillisBetweenRequests =
@@ -58,18 +69,19 @@ class BasicLoadGen {
     );
     this.logger.info(`Running for ${this.options.totalSecondsToRun} seconds`);
 
-    const loadGenContext = initiateLoadGenContext();
+    console.log(
+      '------------ PROCESSING REQUESTS WITHOUT REQUEST COALESCING ------------'
+    );
+
+    let loadGenContext = initiateLoadGenContext();
+    // Show stats periodically.
+    let logStatsIntervalId = setInterval(() => {
+      logStats(loadGenContext, this.logger, this.options.maxRequestsPerSecond);
+    }, this.options.showStatsIntervalSeconds * 1000);
 
     const asyncGetSetResults = range(
       this.options.numberOfConcurrentRequests
-    ).map(workerId =>
-      this.launchAndRunWorkers(momento, loadGenContext, workerId + 1)
-    );
-
-    // Show stats periodically.
-    const logStatsIntervalId = setInterval(() => {
-      logStats(loadGenContext, this.logger, this.options.maxRequestsPerSecond);
-    }, this.options.showStatsIntervalSeconds * 1000);
+    ).map(_ => this.launchAndRunWorkers(momento, loadGenContext));
 
     await Promise.all(asyncGetSetResults);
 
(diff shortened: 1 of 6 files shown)


poppoerika:
Just a quick observation. Based on the screenshots you shared above, the latencies for the Request Coalescer load-gen haven't really improved?

cprice404:
a few quick observations from looking at the stats you printed out:

1. Based on the request counts, I can't tell if you reset the histograms between the two runs?  I would expect roughly an equal number of requests for the before/after stats, but it appears that it might be cumulative?
2. the TPS number is lower for the second set of stats than for the first.  that seems surprising to me, I would have expected it to be higher in the second one.  we should dig into that.
3. It might be worth adding a stat for the number of requests that were actually coalesced so that we can see how often it's happening.
4. you are hitting RESOURCE_EXHAUSTED throttling limits.  we should increase your throttling limits so you don't hit those.

I can play with the code tomorrow morning, or we can pair up to dig into it further, whatever you like!

rishtigupta:
> a few quick observations from looking at the stats you printed out:
> 
> 1. Based on the request counts, I can't tell if you reset the histograms between the two runs?  I would expect roughly an equal number of requests for the before/after stats, but it appears that it might be cumulative?
> 2. the TPS number is lower for the second set of stats than for the first.  that seems surprising to me, I would have expected it to be higher in the second one.  we should dig into that.
> 3. It might be worth adding a stat for the number of requests that were actually coalesced so that we can see how often it's happening.
> 4. you are hitting RESOURCE_EXHAUSTED throttling limits.  we should increase your throttling limits so you don't hit those.
> 
> I can play with the code tomorrow morning, or we can pair up to dig into it further, whatever you like!

I see. I understood the metrics calculations wrong, will coordinate with you tomorrow morning to correct this.

examples/request-coalescing.ts
```diff
@@ -0,0 +1,431 @@
...
+    this.loggerFactory = options.loggerFactory;
+    this.logger = this.loggerFactory.getLogger('request-coalescer-load-gen');
+    this.options = options;
+    this.cacheValue = 'x'.repeat(options.cacheItemPayloadBytes);
+    this.delayMillisBetweenRequests =
+      (1000.0 * this.options.numberOfConcurrentRequests) /
+      this.options.maxRequestsPerSecond;
+  }
+
+  async run(): Promise<void> {
```
cprice404, line 112: If you made this function take a `GetSetOnlyClient` as an argument then you could get rid of the second copy of it, right?  You might need to move the banner logging statements up into the calling function, as well as the post-run coalescing stats logging, but the rest of this code is identical to the other function, right?

poppoerika:
What a difference in the latencies especially in the set operations!
I like how we are displaying the number of coalesced request.

Great work, @rishtigupta 🍾

examples/request-coalescing.ts
```diff
@@ -0,0 +1,431 @@
...
+interface GetAndSetOnlyClient {
+  get(cacheName: string, key: string): Promise<CacheGet.Response>;
+  set(
+    cacheName: string,
+    key: string,
+    value: string
+  ): Promise<CacheSet.Response>;
+}
+
+class MomentoClientWrapperWithCoalescing {
```
poppoerika, line 44: I like the wrapper here 👍

examples/request-coalescing.ts
```diff
@@ -0,0 +1,431 @@
...
+      globalRequestCount: 0,
+      globalSuccessCount: 0,
+      globalUnavailableCount: 0,
+      globalDeadlineExceededCount: 0,
+      globalResourceExhaustedCount: 0,
+      globalRstStreamCount: 0,
+    };
+    return loadGenContext;
+  }
+  private async executeRequestAndUpdateContextCounts<T>(
```
cprice404, line 290: if this function and the one below it are identical to the ones in load-gen, I think these would be reasonable to pull out into your utils classes.

examples/utils/loadGen.ts
```diff
@@ -0,0 +1,40 @@
+import {MomentoLoggerFactory} from '@gomomento/sdk';
```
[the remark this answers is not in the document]
rishtigupta: Thanks for letting me know. Will make changes in the next rev

examples/utils/loadGenStatisticsCalculator.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+export function logCoalescingStats(
+  context: RequestCoalescerContext,
+  logger: MomentoLogger
+): void {
+  logger.info(`
+For request coalescer:
+Number of set requests coalesced:
+${context.numberOfSetRequestsCoalesced}
```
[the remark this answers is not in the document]
rishtigupta: The run in the PR description shows around 47.5% requests coalesced, each for get/set requests. 

Calculation:
(no. of get/set requests coalesced/ total global requests) * 100 = (2850/6000) * 100 = 47.5%z

Right?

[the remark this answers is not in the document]
cprice404: 👍 i just read it wrong when skimming

examples/utils/momento-client-with-coalescing.ts
```diff
@@ -0,0 +1,59 @@
...
+export interface GetAndSetOnlyClient {
+  get(cacheName: string, key: string): Promise<CacheGet.Response>;
+  set(
+    cacheName: string,
+    key: string,
+    value: string
+  ): Promise<CacheSet.Response>;
+}
+
+export class MomentoClientWrapperWithCoalescing {
```
cprice404, line 13: nice, I think breaking this out to a separate file is a good call

cprice404 (approved):

cprice404 (approved):
amazing

examples/utils/momento-client-with-coalescing.ts
poppoerika: +1

poppoerika (approved):
Looks good to me! And awesome work, @rishtigupta 🥳

PR #351, opened 2023-03-16
feat: allow user to unsubscribe when receiving an error

malandis:
In the prior PR, the user had no mechanism to unsubscribe / cancel the
stream. In this PR we allow the user to unsubscribe via the subscription response.
We return the response from `subscribe` and also pass it as an argument to the error handler.

See [here](https://github.com/momentohq/client-sdk-nodejs/pull/350/commits/6b6d0268087b66a3cf4449c5f11d232501099cba) for the examples.

### DIFF
--- a/src/internal/subscription-state.ts
+++ b/src/internal/subscription-state.ts
@@ -1,0 +1,41 @@
+/**
+ * Encapsulates a topic subscription stream state.
+ */
+export class SubscriptionState {
+  private _unsubscribeFn: () => void;
+  public lastTopicSequenceNumber?: number;
+  private _isSubscribed: boolean;
+  constructor() {
+    this._unsubscribeFn = () => {
+      return;
+    };
+    this._isSubscribed = false;
+  }
+
+  public get resumeAtTopicSequenceNumber(): number {
+    return (this.lastTopicSequenceNumber ?? -1) + 1;
+  }
+
+  public setSubscribed(): void {
+    this._isSubscribed = true;
+  }
+
+  public setUnsubscribed(): void {
+    this._isSubscribed = false;
+  }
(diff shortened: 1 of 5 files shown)


malandis:
In a following PR we will add a top level response object for `subscribe` which also exposes an `unsubscribe` method.

allenheltondev (dismissed):
Love it

src/messages/responses/topic-subscribe.ts
```diff
@@ -45,14 +45,52 @@ export class Item extends Response {
...
+  public unsubscribe(): void {
+    if (this.isSubscribed()) {
+      this._unsubscribeFn();
+      this._isSubscribed = false;
+    }
+  }
+
+  public setUnsubscribeFn(unsubscribeFn: () => void): void {
+    this._unsubscribeFn = unsubscribeFn;
+  }
```
cprice404, line 79: this is the only part of this that gives me slight pause.  I feel like we should be able to pass the constructor to this class a reference to an object, rather than passing it a function.  Then we can mutate that (private, internal) object from outside of this class, without having to expose this function.

Another option would be to make `Subscription` just be an interface rather than a class, and just don't include this mutator on the interface.  But I'm not sure how that would fit into the response class hierarchy; so probably better to just explore the first option that I suggested.

src/internal/subscription-state.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Encapsulates a topic subscription stream state.
+ */
+export class SubscriptionState {
```
cprice404, line 4: noice

cprice404 (approved):
unsubscribe

PR #354, opened 2023-03-17
fix: orphaned UUID caches

cprice404:
This commit fixes a few places where we were still creating
caches in tests that only had UUIDs for their names.  Also fixes
a few tests where we were orphaning those caches after the
tests completed.


pgautier404 (approved):
LGTM!

rishtigupta (approved):

PR #357, opened 2023-03-20
chore: add default handlers for subscribe

malandis:
- docs: tweak documentation
- test: test the topic client
- test: validate request timeout
- test: topic client request timeout
- fix: increase wait time for CI server
- chore: add default handlers for pubsub subscribe

### DIFF
--- a/src/topic-client.ts
+++ b/src/topic-client.ts
@@ -45,7 +45,7 @@ export class TopicClient {
    *
    * @param {string} cacheName - The name of the cache to containing the topic to subscribe to.
    * @param {string} topicName - The name of the topic to subscribe to.
-   * @param {SubscribeCallOptions} options - The options for the subscription.
+   * @param {SubscribeCallOptions} options - The options for the subscription. Defaults to no-op handlers.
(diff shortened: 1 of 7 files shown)


allenheltondev (dismissed):
This is a great update. Love it

src/topic-client.ts
```diff
@@ -45,17 +45,17 @@ export class TopicClient {
...
+   * @param {function} options.onError - The callback to invoke when an error is received. Defaults to no-op.
    * @returns {Promise<TopicSubscribe.Response>} -
    * {@link TopicSubscribe.Subscription} on success.
    * {@link TopicSubscribe.Error} on failure.
    */
   public async subscribe(
     cacheName: string,
     topicName: string,
-    options: SubscribeCallOptions
+    options: SubscribeCallOptions = {}
```
cprice404, line 58: is there a use case where a user can do something useful with a call to subscribe without passing one of the handlers?  I think I might keep this option as required, thoughts?

malandis: That's reasonable!

cprice404 (approved):

cprice404 (approved):

PR #359, opened 2023-03-21
fix: remove all grpc leaking into response objects

bruuuuuuuce:
This pr is the first step in allowing shared response types between our nodejs sdk and the new grpc web sdk. Since these 2 sdks have different grpc generated types, we need to remove all references to these types within the response objects. This responses folder will eventually get pulled into the `common` or `shared` folder and imported into both sdk projects.


cprice404 (approved):
hero

Issue #343, opened 2023-03-13
add support for topics/pubsub


cprice404:
- [x] MVP
- [x] Unsubscribe
- [x] Tests
- [x] Cancel when cache client doesn't exist
- [x] UI/UX spike and review

malandis:
MVP: https://github.com/momentohq/client-sdk-nodejs/pull/347

malandis:
Unsubscribe functionality: https://github.com/momentohq/client-sdk-nodejs/pull/351

malandis:
Cancel when cache client doesn't exist: https://github.com/momentohq/client-sdk-nodejs/pull/358

PR #360, opened 2023-03-21
fix: pubsub UX minor fixes and refactor

malandis:
- fix: subscribecalloptions should be required
- chore: rename argument to item handler
- refactor: move TopicSubscribe.Item to TopicItem

Addresses #343

### DIFF
--- a/src/messages/responses/topic-item.ts
+++ b/src/messages/responses/topic-item.ts
@@ -1,0 +1,43 @@
+import {truncateString} from '../../internal/utils/display';
+
+/**
+ * Represents the data received from a topic subscription.
+ *
+ * @remarks A subscription is created by calling {@link TopicClient.subscribe}.
+ * The value is guaranteed to be either a {@link string} or a {@link Uint8Array}.
+ * Call the appropriate accessor if you know the type of the value.
+ */
+export class TopicItem {
+  private readonly _value: string | Uint8Array;
+  constructor(_value: string | Uint8Array) {
+    this._value = _value;
+  }
+  /**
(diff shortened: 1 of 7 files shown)


src/messages/responses/topic-item.ts
```diff
@@ -0,0 +1,43 @@
+import {truncateString} from '../../internal/utils/display';
+
+/**
+ * Represents the data received from a topic subscription.
+ *
+ * @remarks A subscription is created by calling {@link TopicClient.subscribe}.
+ * The value is guaranteed to be either a {@link string} or a {@link Uint8Array}.
+ * Call the appropriate accessor if you know the type of the value.
+ */
+export class TopicItem {
```
allenheltondev, line 10: Thank you for this!

src/messages/responses/topic-item.ts
```diff
@@ -0,0 +1,43 @@
...
+  private readonly _value: string | Uint8Array;
+  constructor(_value: string | Uint8Array) {
+    this._value = _value;
+  }
+  /**
+   * Returns the data read from the stream.
+   * @returns string | Uint8Array
+   */
+  public value(): string | Uint8Array {
```
allenheltondev, line 19: I'm going to submit an issue on this repo to start adding a `json()` option for values that automatically parses the value and returns a json object. It's a nice to have feature I've seen in other places like Postman that stands out as great DX

malandis: Sounds interesting. Can you share a link?

cprice404 (approved):

PR #364, opened 2023-03-21
feat: split out codebase into multiple packages

bruuuuuuuce:
This pr begins the process of splitting out components from the nodejs sdk that can be shared with the grpc-web sdk. The 2 sdks should be able to share almost all code, with the exception of the generated grpc code, which is different if we are generating for a web client vs a nodejs client.

I have added a new `packages/` folder, which will eventually contain 3 packages (currently only has 2, still need to add the `grpc-web-sdk` project). The `client-sdk-nodejs`, `common`, and `grpc-web-sdk`.  The folder structure looks like

```
examples/
README.md
README.template.md
packages/
  client-sdk-nodejs/
    src/
    package.json
    ...
  common/
    src/
    package.json
    ...
  grpc-web-sdk/
    src/
    package.json
    ...
```

I have taken an initial stab at pulling shared utils and response objects into the common package, and removed them from the nodejs-sdk package. There are more things that need to be pulled into the common package, such as client interfaces, however I didn't want this pr to be too large. 

When developing this sdk in the future, when making changes to files inside of the `common` package, the `common` package will need to be rebuilt before changes will take affect in the sdk package. I have added a dev helper script `./scripts/dev-build-nodejs-sdk.sh` that should hopefully help with this process.

This pr shouldn't cause any functionality changes from the clients perspective.

Reviewing this pr, it might be easiest to look at the branch as a whole https://github.com/momentohq/client-sdk-nodejs/tree/fix/multiPackage2


packages/client-sdk-nodejs/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "ts-node": "^10.3.0",
+    "typescript": "^4.4.3",
+    "uuid": "8.3.2"
+  },
+  "dependencies": {
+    "@gomomento/generated-types": "0.51.1",
+    "@grpc/grpc-js": "1.7.3",
+    "google-protobuf": "^3.21.2",
+    "jwt-decode": "3.1.2",
+    "@gomomento/common": "file:../common"
```
cprice404, line 54: Do you know how this will work with the packaging?  Or maybe I'll find that out when I look at your build script.

cprice404 (approved):
This looks like the right first step to me.  But I strongly suspect that the build output dir of the nodejs package is not going to include the code from common, which means that if we publish to npm, it will be broken.  We are going to need to file some follow-up PRs to fix that and and then test the npm packages from a local build before we will be able to attempt our next npmjs release.

But we can do that in subsequent PRs.  Probably good to announce to dev eco channel that we may be in an un-release-able state for a day or two.

PR #368, opened 2023-03-22
fix: npm packaging was broken due to module structure; use symlink for common

cprice404:
In the previous commit we introduced the 'common' dir as a new
npm package named `@gomomento/common`.  We would never publish this
package anywhere, so the files that it contains weren't actually
accessible from the new @gomomento/sdk package.  Even if we bundled
those files into the package by hand, the exports didn't work properly
because they reference `@gomomento/common`, and consumers would never
be able to resolve it.

This commit changes the setup so that we use a symlink to bring in
the common deps rather than a separate npm package; this way we never
need to reference the non-existent `@gomomento/common` package in
the code, and the end result is still a single package that contains
all of the required files.


cprice404:
I have built a package from this locally, and tested that I can consume that package successfully from the examples project.  All of the examples run fine.

If we merge this, I can attempt to do a manual release tomorrow morning and test it with the examples too.  We will want to be ready to revert the last few commits and do another release from the previously successful version if anything is broken.

bruuuuuuuce (approved):
:shipit:

PR #369, opened 2023-03-22
feat: add initial commit for web sdk

bruuuuuuuce:
This pr contains a few new changes

- creates a new ping client within the nodejs sdk
- creates a new web sdk to be used inside of a web browser
- sets up testing environment for web sdk, since its different than that of the nodejs sdk
- moves the `CredentialsProvider` interface into the common directory, since it has no dependencies on node grpc


packages/client-sdk-web/src/config/configuration.ts
```diff
@@ -0,0 +1,46 @@
+import {MomentoLoggerFactory} from '../common/config/logging';
+
+export interface ConfigurationProps {
```
cprice404, line 3: so did you decide you do need this in order for the abstract classes to work?

cprice404 (commented):
just one q, otherwise lgtm

packages/client-sdk-web/src/config/configuration.ts
bruuuuuuuce: this is independent of the abstract class, it can be extended without needing these props. This is just here so we can pass the logger to the web-sdk. I left the other attributes commented out that I felt like we might use it the future.

cprice404: okay but the logger/config object isn't actually required to get it to compiler for now?

bruuuuuuuce: the web-sdk is using the logger, so it is required for compiling rn https://github.com/momentohq/client-sdk-nodejs/pull/369/files#diff-7a6be8180d030dba51f316bc567e3546e1924e68f5cc8bcc8ac98097e14c96f6R43

cprice404: got it.

cprice404 (approved):

PR #370, opened 2023-03-23
chore: tweaks to CI jobs to account for symlink refactor

cprice404:


bruuuuuuuce (dismissed):
:shipit:

scripts/build-package.sh
```diff
@@ -7,6 +7,10 @@ echo "building package: ${1}"
 
 pushd ./packages/${1}
     npm ci
+    # this is gross, but npm will fail to import things
+    # during local builds of files in common if it doesn't
+    # find a node_modules directory there.
+    ln -s $(pwd)/node_modules ../common/node_modules
```
cprice404, line 13: @bruuuuuuuce this is a hack that we might be able to find a better solution for in the future, but it should work for now.  You will need to make sure you run this script to create this symlink for grpc-web sdk when you are using that path.

bruuuuuuuce (approved):
:shipit:

PR #377, opened 2023-03-24
chore: switch back to treating common as a package

cprice404:
After a great deal of experimentation I was unable to find a setup where we could handle the common dir via symlink and also satisfy all of the tools (jest, eslint, etc.)  It seems like the path of least resistance will be to just go ahead and release the common code as its own npm.js package.

We will need to be careful to re-export all of the public APIs from that module so that users don't need to express an explicit dependency on the common package.


poppoerika (approved):

bruuuuuuuce (approved):
:shipit:

PR #378, opened 2023-03-24
chore: shared integration tests and beginning of ICacheClient interface

cprice404:
This commit introduces the beginning of the ICacheClient interface and a pattern for sharing integration tests between the nodejs and grpc-web SDKs.


cprice404:
> we should probably also update the build steps in this pr to also build the new web sdk, looks like its just doing a dependency review

yeah i agree but depending on how the day goes we may just want to do that as a separate pr

bruuuuuuuce (approved):
:shipit:

PR #382, opened 2023-03-24
fix: web control plane endpoint and auth

bruuuuuuuce:
- fixes build so that build gets run on each pr
- fixes control plane client to use correct endpoint

Integration tests now pass in my dev cell using the updated envoy config from https://github.com/momentohq/control-plane-service/pull/306


packages/client-sdk-web/src/internal/control-client.ts
```diff
@@ -76,10 +74,12 @@ export class ControlClient<
     //   configuration: props.configuration,
     // });
     console.log(
-      `\n\n\nCreating control client with endpoint: ${props.credentialProvider.getCacheEndpoint()}\n\n\n`
+      `\n\n\nCreating control client with endpoint: ${props.credentialProvider.getControlEndpoint()}\n\n\n`
```
bruuuuuuuce, line 77: 😄

packages/client-sdk-web/src/internal/control-client.ts
```diff
@@ -76,10 +74,12 @@ export class ControlClient<
...
     );
+
+    this.authToken = props.credentialProvider.getAuthToken();
     this.clientWrapper = new control.ScsControlClient(
-      `https://${props.credentialProvider.getCacheEndpoint()}`,
+      `https://${props.credentialProvider.getControlEndpoint()}`,
```
bruuuuuuuce, line 82: 😄

packages/client-sdk-web/src/internal/control-client.ts
```diff
@@ -130,7 +130,7 @@ export class ControlClient<
     return await new Promise<DeleteCache.Response>(resolve => {
       this.clientWrapper.deleteCache(
         request,
-        null,
+        {authorization: this.authToken},
```
bruuuuuuuce, line 133: this seems to be a special header than needs to get passed at the request level

packages/client-sdk-web/src/internal/control-client.ts
cprice404: lol

packages/client-sdk-web/src/internal/control-client.ts
```diff
@@ -100,7 +100,7 @@ export class ControlClient<
     return await new Promise<CreateCache.Response>(resolve => {
       this.clientWrapper.createCache(
         request,
-        null,
+        this.authHeaders,
```
cprice404, line 103: just curious why you're doing the agent header in an interceptor and the auth header on the request?

scripts/build-package.sh
```diff
@@ -6,7 +6,7 @@ set -e
 echo "building package: ${1}"
 
 pushd ./packages/${1}
-    npm ci
+    npm ci -f
```
cprice404, line 9: what does that do?

cprice404 (commented):
:nice:

scripts/build-package.sh
bruuuuuuuce: it does a force install, I kept getting flaky github errors around `npm ERR! code EEXIST`. We can fix this before we merge back to main, this is just what fixed it and unblocked me for building the 2 sdks on pr

cprice404: okay maybe leave a comment there so it's obvious, then i'm fine with it for now

packages/client-sdk-web/src/internal/control-client.ts
bruuuuuuuce: honestly not sure, maybe our interceptors just aren't working correctly? I can change it if you want.

cprice404: it's not urgent but would be nice if they were consistent.  could just file a follow-up ticket.  not a blocker.

bruuuuuuuce:
integration tests should pass once this pr https://github.com/momentohq/control-plane-service/pull/306 is merged and deployed to alpha

.github/workflows/build.yml
```diff
@@ -2,7 +2,7 @@ name: Build
 
 on:
   pull_request:
-    branches: [main]
+    branches: [main, console-mvp]
```
bruuuuuuuce, line 5: I am seeing some strange behavior from earlier versions of nodejs with the web sdk. Removing them for now to unblock us. Will have to come back to this when we merge it into main

poppoerika (approved):
Looks good to me.

PR #385, opened 2023-03-28
fix: `setIfNotExists` should accept a `value` not `field`

malandis:
The API for `setIfNotExists` has a typo: the parameter for `value` was mis-named `field`.


bruuuuuuuce (approved):
:shipit:

PR #386, opened 2023-03-28
chore: remove text encoder; rename common package from core to common

poppoerika:
Ticket: https://github.com/momentohq/dev-eco-issue-tracker/issues/255


packages/common/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@gomomento/core",
+  "name": "@gomomento/common",
```
poppoerika, line 2: @bruuuuuuuce 
I wasn't sure it was intentionally named `core` although the folder is named `common`?

bruuuuuuuce:
I was having problems running matrix tests against this branch, so for now I removed them and am only running tests against node 18. Once node 18 passes, can we change this to node 14 https://github.com/momentohq/client-sdk-nodejs/blob/chore/remove-text-encoder/.github/workflows/build.yml#L31, the oldest version of node we support, and make sure that this is still passing. Ideally we could run all node versions on every pr, however I was running into some strange problems with how we are packaging things

PR #388, opened 2023-03-29
fix: pass other errors to user's subscribe error handler

malandis:
Prior to this we only passed `NOT_FOUND` gRPC errors to the user's error handler. We should invoke the error handlers on other errors but not subscribe.


bruuuuuuuce (approved):
:shipit:

PR #390, opened 2023-03-29
chore: finish interface for control client

bruuuuuuuce:


packages/client-sdk-web/src/cache-client.ts
```diff
@@ -1,17 +1,21 @@
...
 import {CredentialProvider} from '@gomomento/core';
+import {ICacheClient} from '../../common/dist/src/internal/clients/cache/ICacheClient';
+import {DataClient} from './internal/data-client';
 
 export interface CacheClientProps {
   credentialProvider: CredentialProvider;
 }
 
-export class CacheClient extends AbstractCacheClient {
+export class CacheClient extends AbstractCacheClient implements ICacheClient {
```
bruuuuuuuce, line 14:
```suggestion
export class CacheClient extends AbstractCacheClient {
```

packages/client-sdk-web/src/internal/data-client.ts
```diff
@@ -0,0 +1,208 @@
...
+        value.length
+      }, ttl: ${ttlToUse.toString()}`
+    );
+
+    return await this.sendSet(cacheName, key, value, ttlToUse);
+  }
+
+  private async sendSet(
+    cacheName: string,
+    key: string | Uint8Array,
```
bruuuuuuuce, line 178: grpc web is nice cause the generated request objects can take in either string or Uint8Array, so no need for a `convert` function

packages/common/src/internal/clients/cache/AbstractCacheClient.ts
```diff
@@ -44,4 +61,1062 @@ export abstract class AbstractCacheClient implements ICacheClient {
...
+   * Removes the given key from the cache. The key can represent a single value
+   * or a collection.
+   *
+   * @param {string} cacheName - The cache to delete from.
+   * @param {string | Uint8Array} key - The key to delete.
+   * @returns {Promise<CacheDelete.Response>} -
+   * {@link CacheDelete.Success} on success.
+   * {@link CacheDelete.Error} on failure.
+   */
+  // public async delete(
```
bruuuuuuuce, line 116: I copied all the signatures over from our nodejs client, and just commented out the ones that we havent implemented yet. We will implement these soon so felt like it was fine to leave here for now

packages/common/src/internal/clients/cache/IDataClient.ts
```diff
@@ -0,0 +1,11 @@
+import {CacheGet, CacheSet} from '../../../index';
+
+export interface IDataClient {
+  get(cacheName: string, key: string | Uint8Array): Promise<CacheGet.Response>;
+  set(
+    cacheName: string,
+    key: string | Uint8Array,
+    value: string | Uint8Array,
+    ttl?: number
```
bruuuuuuuce, line 9: im not sure if we want the internal data clients to also use the custom option types `SetOptions`? They arent currently, but then we wouldnt need separate interfaces for the public data clients and internal data clients

bruuuuuuuce: I am inclined to leave it as is for now, and then maybe we can come back it it once we merge in the feature branch to main

packages/client-sdk-web/src/internal/data-client.ts
malandis: Nice!

packages/client-sdk-web/src/internal/data-client.ts
```diff
@@ -0,0 +1,206 @@
...
+    value: string | Uint8Array,
+    ttl: number
+  ): Promise<CacheSet.Response> {
+    const request = new _SetRequest();
+    request.setCacheKey(key);
+    request.setCacheBody(value);
+    request.setTtlMilliseconds(ttl * 1000);
```
malandis, line 183: Let's extract `ttl * 1000` to a method since this will be repeated

packages/common/src/internal/clients/cache/IDataClient.ts
malandis: Can loop back to this later

packages/client-sdk-web/src/internal/data-client.ts
bruuuuuuuce: this actually should not be `*1000` right? user passes in milliseconds correct?

packages/client-sdk-web/src/cache-client.ts
poppoerika: Do we not need to update to 
```
export class CacheClient extends AbstractCacheClient {
```

poppoerika (approved):
Looks good to me!

packages/client-sdk-nodejs/src/cache-client.ts
```diff
@@ -1191,11 +1185,11 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
     );
   }
 
-  private getNextDataClient(): DataClient {
+  protected getNextDataClient(): DataClient {
     const client = this.dataClients[this.nextDataClientIndex];
     this.nextDataClientIndex =
       (this.nextDataClientIndex + 1) % this.dataClients.length;
-    return client;
+    return client as DataClient;
```
bruuuuuuuce, line 1192: once the pair of nodejs/web sdks are at parity, this type cast will no longer be needed

malandis (approved):

PR #391, opened 2023-04-04
fix: encode grpc web requests using text encoder

bruuuuuuuce:
Before, we were letting the grpc web generate clients encode string requests into uint8 arrays. This was causing us problems deserializing requests from the ui, so we are changing to encoding everything using a `TextEncoder` like we are doing inside of the nodejs sdk

### DIFF
--- a/packages/client-sdk-web/src/internal/data-client.ts
+++ b/packages/client-sdk-web/src/internal/data-client.ts
@@ -43,11 +43,13 @@ export class DataClient<
   private readonly logger: MomentoLogger;
   private readonly authHeaders: {authorization: string};
   private readonly defaultTtlSeconds: number;
+  private readonly textEncoder: TextEncoder;
 
   /**
    * @param {DataClientProps} props
    */
   constructor(props: DataClientProps) {
+    this.textEncoder = new TextEncoder();
     this.logger = props.configuration.getLoggerFactory().getLogger(this);
     const headers = [new Header('Agent', `nodejs:${version}`)];
     this.interceptors = [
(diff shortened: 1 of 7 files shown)


packages/client-sdk-web/src/internal/data-client.ts
```diff
@@ -168,16 +170,23 @@ export class DataClient<
...
     cacheName: string,
-    key: string | Uint8Array,
-    value: string | Uint8Array,
+    key: Uint8Array,
+    value: Uint8Array,
     ttlSeconds: number
   ): Promise<CacheSet.Response> {
     const request = new _SetRequest();
+    console.log('send set key', key);
+    console.log('send set value', value);
```
poppoerika, line 189: Do you meant to leave these?

packages/common-integration-tests/src/create-delete-list-cache.ts
```diff
@@ -91,7 +91,13 @@ export function runCreateDeleteListCacheTests(Momento: ICacheClient) {
         const value1 = v4();
         const value2 = v4();
         await WithCache(Momento, cacheName, async () => {
+          console.log(
+            `setting cache: ${cacheName}, key: ${key1}, value: ${value1}`
+          );
```
poppoerika, line 96: Do you meant to leave this as well?

packages/common-integration-tests/src/create-delete-list-cache.ts
```diff
@@ -91,7 +91,13 @@ export function runCreateDeleteListCacheTests(Momento: ICacheClient) {
...
           await Momento.set(cacheName, key1, value1);
+          console.log(
```
poppoerika, line 98: Do you meant to leave this as well?

PR #392, opened 2023-04-05
fix: base64 data for grpc web client

bruuuuuuuce:
Completes the unary apis for the grpc web sdk

- fixes bug where we were not base64 encoding grpc web requests
- implements increment, delete, and setIfNotExists for grpc web sdk
- pulls out `get-set-delete` test into shared common integration tests folder

The grpc web sdk doesnt have custom configurations yet (Laptop.Latest()  etc..), so I have left those tests testing different timeouts in the nodejs package.


packages/client-sdk-web/src/internal/data-client.ts
```diff
@@ -200,11 +212,199 @@ export class DataClient<
...
     return {cache: cacheName};
   }
 
   private convertSecondsToMilliseconds(ttlSeconds: number): number {
     return ttlSeconds * 1000;
   }
+
+  private convert(v: string | Uint8Array): string {
+    if (typeof v === 'string') {
+      return btoa(v);
```
poppoerika, line 404: This was the key!

packages/client-sdk-web/src/internal/data-client.ts
```diff
@@ -200,11 +212,199 @@ export class DataClient<
     });
   }
 
+  public async setIfNotExists(
```
poppoerika, line 215: Thank you for adding those.

poppoerika (approved):
Looks good to me!

PR #393, opened 2023-04-06
feat: add `GenerateApiToken`

poppoerika:
Tested on the client proto `0.56.0` without passing a session token generated by our preview console as a session token parameter. But instead, I passed it to the auth header in place of an auth token.

For this PR, I already updated to use the client proto `0.57.0` where the auth proto is introduced. 
But this should not impact the result from any testing I did with `0.56.0`. (@rlinehan am I understanding this correct?)

To pass the build workflow, `TEST_SESSION_TOKEN` needs to be added to the GH secret. Also [this PR](https://github.com/momentohq/control-plane-service/pull/310/files) needs to be merged and be deployed to preprod.

Ticket: https://github.com/momentohq/dev-eco-issue-tracker/issues/276


.github/workflows/build.yml
```diff
@@ -34,6 +34,7 @@ jobs:
     env:
       TEST_AUTH_TOKEN: ${{ secrets.ALPHA_TEST_AUTH_TOKEN }}
       TEST_CACHE_NAME: js-integration-test-ci
+      TEST_SESSION_TOKEN: ${{ secrets.ALPHA_TEST_SESSION_TOKEN }}
```
bruuuuuuuce, line 37: These session tokens expire after a few hours, we would need a way to programatically regenerate them in order to have these integration tests run consistantly

packages/client-sdk-nodejs/src/internal/control-client.ts
```diff
@@ -39,6 +42,7 @@ export interface ControlClientProps {
 
 export class ControlClient {
   private readonly clientWrapper: GrpcClientWrapper<grpcControl.ScsControlClient>;
+  private readonly clientAuthWrapper: GrpcClientWrapper<grpcAuth.AuthClient>;
```
bruuuuuuuce, line 45: Do we want to pull this out into its own client? The pattern that we currently have set with our clients is that they are 1:1 with a grpc service. 

https://github.com/momentohq/client-sdk-nodejs/tree/console-mvp/packages/client-sdk-nodejs/src/internal

If you look here, we have separated out `ping-service`, `data-client`, `control-client`, and `pubsub-client` into their own client files. I think we should maintain this pattern with this new auth service.

bruuuuuuuce: The other thing is, do we want to have the auth client initilaize with all the other clients, or be totally separate? I am leaning towards it being totally separate because of the auth situation. We shouldn't need an auth token to initialize this client, and I don't _think_ that alot of customers will be using these endpoints directly themselves. Would be interested in what @cprice404 thinks of this

packages/client-sdk-nodejs/src/internal/control-client.ts
```diff
@@ -182,6 +195,36 @@ export class ControlClient {
     });
   }
 
+  public async generateApiToken(
+    sessionToken: string
```
bruuuuuuuce, line 199: when we start allowing generation of tokens to expire, what will the method signatures look like? Want to make sure we don't have any breaking changes.

packages/client-sdk-web/src/internal/control-client.ts
```diff
@@ -36,6 +40,7 @@ export class ControlClient<
 > implements IControlClient
 {
   private readonly clientWrapper: control.ScsControlClient;
+  private readonly clientAuthWrapper: auth.AuthClient;
```
bruuuuuuuce, line 43: ditto about the clients

packages/common/src/messages/responses/generate-api-token.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export abstract class Response extends ResponseBase {}
+
+class _Success extends Response {
+  readonly apiToken: string;
+  readonly refreshToken: string;
+  readonly endpoint: string;
+  readonly validUntil: number;
+
+  constructor(result?: _GenerateApiTokenResponse) {
```
bruuuuuuuce, line 13: should this be required? If this is optional, then your getters will return optional strings

.github/workflows/build.yml
bruuuuuuuce: if we _really_ wanted to have session token integration tests, we could have selenium log into our console, grab a session token and use that. But we can probably set that up later.

bruuuuuuuce (commented):
The main concern I have here is putting multiple grpc clients as part of a single `MomentoClient`. This seems to be breaking the paradigm that we have currently set. The auth service is a bit of a snowflake tho, would like to hear @cprice404 thoughts on it

.github/workflows/build.yml
bruuuuuuuce: for sure, its a really good test to have!

poppoerika:
I need to update this PR to pass a session token as a parameter and not pass anything to the auth header with the change happened [here](https://github.com/momentohq/control-plane-service/pull/310#discussion_r1159120265).

packages/client-sdk-nodejs/src/internal/control-client.ts
poppoerika: After discussing with @cprice404 and @bruuuuuuuce, we are taking this out and create an Auth client.

.github/workflows/build.yml
poppoerika: Leave the test and skip it for now.

packages/client-sdk-nodejs/src/internal/control-client.ts
poppoerika: That's a good call. I can add an optional `validUntilSeconds` as a parameter.

packages/client-sdk-web/src/internal/control-client.ts
poppoerika: https://github.com/momentohq/client-sdk-nodejs/pull/393#discussion_r1160137539

.github/workflows/build.yml
cprice404: we should remove this env var from the github thing for now right?

packages/client-sdk-nodejs/src/internal/auth-client.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * @param {AuthClientProps} props
+   */
+  constructor(props: AuthClientProps) {
+    this.logger = props.configuration.getLoggerFactory().getLogger(this);
+    const headers = [
+      new Header('Authorization', props.credentialProvider.getAuthToken()),
+      new Header('Agent', `nodejs:${version}`),
+    ];
```
cprice404, line 33: do we need the auth header?

packages/client-sdk-web/src/auth-client.ts
```diff
@@ -0,0 +1,20 @@
...
+import {
+  AuthClientProps,
+  InternalWebGrpcAuthClient,
+} from './internal/auth-client';
+
+// TODO
+// TODO
+// TODO - before we do a 1.0 release this needs to be removed; we will just have a ping method on the CacheClient.
+// TODO
+// TODO
```
cprice404, line 11: probably don't need these comments?

packages/client-sdk-web/src/internal/auth-client.ts
```diff
@@ -0,0 +1,91 @@
...
+  credentialProvider: CredentialProvider;
+}
+
+export class InternalWebGrpcAuthClient<
+  REQ extends Request<REQ, RESP>,
+  RESP extends UnaryResponse<REQ, RESP>
+> {
+  private readonly clientAuthWrapper: auth.AuthClient;
+  private readonly interceptors: UnaryInterceptor<REQ, RESP>[];
+  // private static readonly REQUEST_TIMEOUT_MS: number = 60 * 1000;
```
cprice404, line 24: need this?

packages/client-sdk-web/src/internal/auth-client.ts
```diff
@@ -0,0 +1,91 @@
...
+   * @param {AuthClientProps} props
+   */
+  constructor(props: AuthClientProps) {
+    this.logger = props.configuration.getLoggerFactory().getLogger(this);
+    const headers = [new Header('Agent', `nodejs:${version}`)];
+    this.interceptors = [
+      new HeaderInterceptorProvider<REQ, RESP>(
+        headers
+      ).createHeadersInterceptor(),
+      // ClientTimeoutInterceptor(ControlClient.REQUEST_TIMEOUT_MS),
```
cprice404, line 37:
```suggestion
```

packages/client-sdk-web/src/internal/auth-client.ts
```diff
@@ -0,0 +1,91 @@
...
+      `Creating control client using endpoint: '${props.credentialProvider.getControlEndpoint()}`
+    );
+    // this.clientWrapper = new IdleGrpcClientWrapper({
+    //   clientFactoryFn: () =>
+    //     new grpcControl.ScsControlClient(
+    //       props.credentialProvider.getControlEndpoint(),
+    //       ChannelCredentials.createSsl()
+    //     ),
+    //   configuration: props.configuration,
+    // });
```
cprice404, line 49:
```suggestion
```

packages/common/src/messages/responses/grpc-response-types.ts
```diff
@@ -57,6 +57,24 @@ export class _ListCachesResponse {
...
+export class _GenerateApiTokenResponse {
+  readonly apiToken: string;
+  readonly refreshToken: string;
+  readonly endpoint: string;
+  readonly validUntil: number;
+  constructor(
+    apiToken?: string,
+    refreshToken?: string,
+    endpoint?: string,
+    validUntil?: number
```
cprice404, line 69: what's the motivation for making these nullable?

cprice404 (commented):
looking good!  handful of nits

packages/client-sdk-nodejs/src/internal/auth-client.ts
poppoerika: Ah we don't. I'll remove it.

packages/common/src/messages/responses/grpc-response-types.ts
poppoerika: I forgot to update those to be required. Thank you!

packages/client-sdk-web/src/internal/auth-client.ts
```diff
@@ -0,0 +1,81 @@
...
+import {Header, HeaderInterceptorProvider} from './grpc/headers-interceptor';
+import {_GenerateApiTokenRequest} from '@gomomento/generated-types-webtext/dist/auth_pb';
+import {cacheServiceErrorMapper} from '../errors/cache-service-error-mapper';
+import {_GenerateApiTokenResponse} from '@gomomento/common';
+import Never = _GenerateApiTokenRequest.Never;
+import Expires = _GenerateApiTokenRequest.Expires;
+
+export interface AuthClientProps {
+  configuration: Configuration;
+  credentialProvider: CredentialProvider;
```
bruuuuuuuce, line 15: Do we want a credentials provider as props? This means that this client would require an auth token. I thought we wanted to be able to initialize it without a token

packages/client-sdk-web/src/internal/auth-client.ts
```diff
@@ -0,0 +1,81 @@
...
+  constructor(props: AuthClientProps) {
+    this.logger = props.configuration.getLoggerFactory().getLogger(this);
+    const headers = [new Header('Agent', `nodejs:${version}`)];
+    this.interceptors = [
+      new HeaderInterceptorProvider<REQ, RESP>(
+        headers
+      ).createHeadersInterceptor(),
+    ];
+    this.logger.debug(
+      `Creating control client using endpoint: '${props.credentialProvider.getControlEndpoint()}`
```
bruuuuuuuce, line 38:
```suggestion
      `Creating auth client using endpoint: '${props.credentialProvider.getControlEndpoint()}`
```

packages/client-sdk-web/src/internal/auth-client.ts
```diff
@@ -0,0 +1,81 @@
...
+    );
+    console.log(
```
bruuuuuuuce, line 40: this can be deleted

packages/client-sdk-web/src/internal/auth-client.ts
```diff
@@ -0,0 +1,81 @@
...
+      null,
+      {
+        unaryInterceptors: this.interceptors,
+      }
+    );
+  }
+
+  public async generateApiToken(
+    sessionToken: string,
+    validUntilSeconds?: number
```
bruuuuuuuce, line 55: I don't know of the best way to do this, but it would be nice if this could be something like

```
expiresAt: SomeSeconds | Never
```

But I am not sure the best way to expose this to customers other than having 2 separate methods, which is not ideal

packages/common/src/messages/responses/generate-api-token.ts
```diff
@@ -0,0 +1,61 @@
...
+export class Success extends ResponseSuccess(_Success) {}
+
+class _Error extends Response {
+  constructor(protected _innerException: SdkError) {
+    super();
+  }
+}
+
+/**
+ * Indicates that an error occurred during the list caches request.
```
bruuuuuuuce, line 52:
```suggestion
 * Indicates that an error occurred during the generate api token request.
```

packages/common/src/messages/responses/generate-api-token.ts
```diff
@@ -0,0 +1,61 @@
...
+    return this.endpoint;
+  }
+
+  public getValidUntil() {
+    return this.validUntil;
+  }
+}
+
+/**
+ * Indicates a Successful list caches request.
```
bruuuuuuuce, line 41:
```suggestion
 * Indicates a Successful generate api token request
```

packages/client-sdk-web/src/internal/auth-client.ts
poppoerika: Good call.

poppoerika: @bruuuuuuuce hmmm but I still need an auth token to get a control endpoint.
So we don't need to pass an auth token to the auth header when calling the generate api token api, but I still need an auth token to get a control endpoint, right?

poppoerika: Ah, never mind. We have our own end pointer creater.

packages/client-sdk-web/src/internal/auth-client.ts
poppoerika: Created: https://github.com/momentohq/client-sdk-nodejs/issues/394

packages/client-sdk-web/src/internal/auth-client.ts
```diff
@@ -0,0 +1,73 @@
...
+import {Configuration} from '../config/configuration';
+import {Request, UnaryInterceptor, UnaryResponse} from 'grpc-web';
+import {Header, HeaderInterceptorProvider} from './grpc/headers-interceptor';
+import {_GenerateApiTokenRequest} from '@gomomento/generated-types-webtext/dist/auth_pb';
+import {cacheServiceErrorMapper} from '../errors/cache-service-error-mapper';
+import {_GenerateApiTokenResponse} from '@gomomento/common';
+import Never = _GenerateApiTokenRequest.Never;
+import Expires = _GenerateApiTokenRequest.Expires;
+
+export interface AuthClientProps {
```
cprice404, line 13: This is a public interface, which users will have to instantiate in order to create the client.  So it shouldn't live in an `internal` file or else they would have to import it from there.  We will either need to move it to the main auth-client.ts file, or put it in its own file to avoid circular dependencies between the two files.

Also - you currently have the `controlEndpoint` specified as a required parameter.  In the 90% use case if a customer instantiates this client we don't want them to have to know that endpoint, and we should be able to use a default one, right?

Neither of these have to be addressed in this PR as long as we are not touching the nodejs SDK (i.e. remove the file I commented on above), but we should create tickets for them if we aren't going to address them now.  I do want to get this merged to unblock you to finish the console work.

cprice404 (commented):
just a few nits, I think if you delete the file you added to nodejs SDK and create a ticket for the other two things i mentioned then I will be +1 on this PR

poppoerika:
@cprice404 @bruuuuuuuce
I'm testing a new auth token generated by the generate api token api. Report back soon 🙇‍♀️

packages/client-sdk-web/src/internal/auth-client.ts
poppoerika: Removed `AuthClient` from `client-sdk-nodejs` and created the tickets to address your feedback:
- https://github.com/momentohq/client-sdk-nodejs/issues/395
- https://github.com/momentohq/client-sdk-nodejs/issues/396

packages/common/src/messages/responses/generate-api-token.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (result) {
+      this.apiToken = result.apiToken;
+      this.refreshToken = result.refreshToken;
+      this.endpoint = result.endpoint;
+      this.validUntil = result.validUntil;
+    }
+  }
+
+  public getApiToken() {
+    return encodeToBase64(
```
bruuuuuuuce, line 25: 🦾

cprice404 (approved):
i'm good with this when @bruuuuuuuce is.  thanks for creating those tickets.

bruuuuuuuce (approved):
:shipit:

PR #403, opened 2023-04-11
fix: support decoding environment variable tokens as v1 tokens

bruuuuuuuce:


src/internal/utils/auth.ts
```diff
@@ -53,7 +53,7 @@ export const decodeAuthToken = (token?: string): TokenAndEndpoints => {
       }
       return {
         controlEndpoint: `control.${base64DecodedToken.endpoint}`,
-        cacheEndpoint: `data.${base64DecodedToken.endpoint}`,
+        cacheEndpoint: `cache.${base64DecodedToken.endpoint}`,
```
bruuuuuuuce, line 56: this was bug that we fixed in `console-mvp` branch

cprice404 (approved):

PR #426, opened 2023-04-14
chore: add all public types to interface

cprice404:
Prior to this commit, there were quite a few types that were
not exported from the main index.  This meant that if you needed
to reference them in your code your import statement would include
the `dist/src` dir explicitly, which is leaking implementation
details.  This is also problematic for the case where we want
to split out a common/core package because those paths would
change.

To remedy that, this commit adds explicit exports for all of the
types that were missing.  Once this is merged we can do a release
and then start working on merging the feature branch back in.


rishtigupta (approved):

bruuuuuuuce (approved):
:shipit:

PR #428, opened 2023-04-18
chore: merge recent changes from main (2/2)

cprice404:
This PR contains the last commits that we were missing from `main`; it should get us back into sync with the `main` branch.  Then we just need to sort out the tasks to do releases to npm.js, and then we can close the feature branch.

This commit brings in the `index.ts` updates from `main`, but also retrofits them onto this branch.  i.e., it updates the indices in common and client-sdk-web to match what we were setting out to do in the nodejs index.  Also updates the nodejs index to import things from common, and updates all of the imports in various other files in all the packages to import from the new locations.

And - removes a few things from grpc-response-types.ts; it was feeling like we were starting to put a lot of stuff in there that didn't need to be there.


packages/client-sdk-web/package.json
```diff
@@ -48,7 +48,7 @@
     "xhr2": "^0.2.1"
   },
   "dependencies": {
-    "@gomomento/common": "file:../common",
+    "@gomomento/core": "file:../common",
```
bruuuuuuuce, line 51: lol I feel like we keep changing this. `core` makes more sense to me if we are publishing it to npm

packages/client-sdk-web/src/cache-client.ts
```diff
@@ -1,13 +1,12 @@
...
+import {NoopMomentoLoggerFactory, CredentialProvider} from '@gomomento/core';
 import {
   AbstractCacheClient,
   IControlClient,
-  NoopMomentoLoggerFactory,
-  CredentialProvider,
   IDataClient,
   ICacheClient,
-} from '@gomomento/common';
+} from '@gomomento/core/dist/src/internal/clients/index';
```
bruuuuuuuce, line 9: I thought we didn't want to have `/dist/` imports?

packages/client-sdk-web/src/index.ts
```diff
@@ -1,61 +1,71 @@
 import {CacheClient} from './cache-client';
-import {AuthClient} from './auth-client';
+
+// Cache Client Response Types
+import * as CacheGet from '@gomomento/core/dist/src/messages/responses/cache-get';
```
bruuuuuuuce, line 4: ditto with the `dist` imports for all of these

bruuuuuuuce (commented):
a few questions about the imports but mostly looks good. Thanks for doing this :fistbump:

pgautier404 (approved):

packages/client-sdk-web/src/cache-client.ts
cprice404: discussed on the call - we don't want **users** to ever need to do imports from `dist`; in this case these are private internal classes that users shouldn't be using, so it's fine.

PR #432, opened 2023-04-20
chore: migrate common scalar integration tests

pgautier404:
This commit is just some cleanup to finalize the sharing of scalar integration tests between the nodejs and web sdks. It retains and renames some client timeout tests that are not applicable to the web client.


packages/client-sdk-nodejs/test/integration/client-timeout.test.ts
```diff
@@ -1,23 +1,22 @@
...
   WithCache,
-  testCacheName,
 } from './integration-setup';
 
-const {Momento, IntegrationTestCacheName} = SetupIntegrationTest();
+const {Momento} = SetupIntegrationTest();
 
-describe('get/set/delete', () => {
-  it('should timeout on a request that exceeds specified timeout', async () => {
+describe('client timeout tests', () => {
```
cprice404, line 18: interesting - so you're saying the timeout tests aren't relevant to the web sdk?  Is that because grpc-web doesn't support timeouts or have we just not wired that up to configuration there yet?

if the latter, this is probably okay for now but we will definitely want to create a ticket about following up on that, and it would be good to add a comment here explaining that most of the tests have moved to shared and explaining why these are still here.

bruuuuuuuce: idk if it supports timeouts or not, but I know we have not wired up the timeout configuration

bruuuuuuuce: I think it does, I think we just need to add some interceptors. But have not played around with it too much.

pgautier404: The latter-- the web SDK didn't seem to be wired up and I wasn't sure what the plan was there. Sounds like this is a loose end so I'll add a ticket and comments.

cprice404: :ty: yes I do think we will want to wire up timeouts before web sdk goes to 1.0.

cprice404 (approved):

PR #443, opened 2023-04-21
fix: pass 'public' flag to npm publish

cprice404:
At some point recently npm.js started requiring you to pass
the 'public' access flag to publish public packages, otherwise
the publish fails with an error about how you have to pay for
private packages.


bruuuuuuuce (approved):
:shipit:

PR #444, opened 2023-04-21
feat: publish the node.js client library from the new package structure

cprice404:
This commit updates the publish scripts to publish the node.js
SDK package in addition to the core package.  We add a script
to wait for the core package to be available on node.js, then
update the dependency in the nodejs SDK package to use the newly
released version of core, then publish the nodejs package.

### DIFF
--- a/scripts/publish-all-packages.sh
+++ b/scripts/publish-all-packages.sh
@@ -20,3 +20,5 @@ ROOT_DIR="$(dirname "$0")/.."
 echo "publishing all packages"
 
 ${ROOT_DIR}/scripts/publish-package.sh "core" "${VERSION}"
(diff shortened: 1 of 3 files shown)


bruuuuuuuce (approved):
:shipit:

cprice404:
had to rebase

bruuuuuuuce:
@cprice404 looks like tests are failing

cprice404:
> @cprice404 looks like tests are failing

there is some weird intermittent test failure happening where the web sdk List collection methods show up as "undefined".  then if you re-run the tests they pass.  We need to get to the bottom of it; my best guess for now is that there is some github npm cache that is occasionally linking us against older commits of the core lib or something.

We need to figure it out if it keeps happening but for now I think I want to just get the release process working again first.

bruuuuuuuce (approved):
:shipit:

pgautier404 (approved):

PR #445, opened 2023-04-24
chore: migrate common dict integration tests

bruuuuuuuce:


packages/client-sdk-web/src/internal/data-client.ts
```diff
@@ -1082,4 +1650,37 @@ export class DataClient<
...
+    if (elements instanceof Map) {
+      return [...elements.entries()].map(element =>
+        new _DictionaryFieldValuePairGrpc()
+          .setField(this.convert(element[0]))
+          .setValue(this.convert(element[1]))
+      );
+    } else {
+      return Object.entries(elements).map(element =>
+        new _DictionaryFieldValuePairGrpc()
+          .setField(element[0])
```
pgautier404, line 1668: I was doing some inflight troubleshooting here, just FYI. This should be `.setField(this.convert(element[0]))` instead. Looks like changing it back resolves a couple of the test failures as well.

PR #451, opened 2023-04-24
fix: change error type to concrete type rather than `object`

cprice404:
A user noticed that our `innerException` signatures were declaring
that they returned `object`, when in actuality they always return
an SdkError.  This commit fixes the signature so that end users
can do more useful things with the Errors.


malandis:
Along these lines, the `Error` responses do not derive from `Error`, which is unexpected.

cprice404:
> Along these lines, the `Error` responses do not derive from `Error`, which is unexpected.

that's fair but that's a bigger topic, would rather deal with that separately

pgautier404 (approved):

PR #455, opened 2023-04-25
chore: remove nodejs implementations of abstract client funcs

pgautier404:
This commit removes nodejs SDK cache client implementations of methods that have been moved to the AbstractCacheClient class we share with the web SDK.


cprice404 (approved):
my favorite kind of PR: all red! :)

PR #458, opened 2023-04-26
WIP: Common dict int tests3

cprice404:


packages/client-sdk-nodejs/test/integration/foo.test.ts
```diff
@@ -0,0 +1,36 @@
+import {CacheClient, CredentialProvider, Configurations} from '../../src';
```
cprice404, line 1: this file should be deleted.  I was using it to do some writes from the SDK and then test whether i could `get` the corresponding keys from the CLI.  This is how I confirmed that the collections were not being stored to the right keys.

packages/client-sdk-web/package.json
```diff
@@ -16,11 +16,13 @@
     "prebuild": "eslint . --ext .ts",
     "test": "jest",
     "unit-test": "jest unit --passWithNoTests",
-    "integration-test": "jest integration --passWithNoTests",
```
cprice404, line 19: i don't think we should need these `passWithNoTests` lines

packages/client-sdk-web/src/index.ts
```diff
@@ -59,11 +59,9 @@ import {
   CacheInfo,
   CollectionTtl,
   SortedSetOrder,
-  // Credentials/Auth
```
cprice404, line 62: these bit me again :(

packages/client-sdk-web/src/internal/data-client.ts
```diff
@@ -136,7 +136,7 @@ export class DataClient<
       return new CacheGet.Error(normalizeSdkError(err as Error));
     }
     this.logger.trace(`Issuing 'get' request; key: ${key.toString()}`);
-    const result = await this.sendGet(cacheName, this.convert(key));
+    const result = await this.sendGet(cacheName, this.convertToB64String(key));
```
cprice404, line 139: I renamed this function to make it more clear what it's doing

packages/client-sdk-web/src/internal/data-client.ts
```diff
@@ -1091,9 +1091,9 @@ export class DataClient<
     );
     const result = await this.sendDictionarySetField(
       cacheName,
-      dictionaryName,
-      this.convert(field),
-      this.convert(value),
+      this.convertToB64String(dictionaryName),
```
cprice404, line 1094: This is the crux of the issue.  You need to b64 the collection name or else it gets stored with a weird bytes key that is not accessible by other SDKs or the CLI.

You need to be careful to do this in *EVERY* collection API.

I think it's also crucial that we add a DELETE test for every collection type.  It won't 100% guarantee that we do the base64 correctly in all APIs but it dramatically increases the odds of us catching a bug.

packages/client-sdk-web/src/internal/data-client.ts
```diff
@@ -1331,12 +1331,13 @@ export class DataClient<
   private async sendDictionaryGetFields(
     cacheName: string,
     dictionaryName: string,
-    fields: string[]
+    fields: string[] | Uint8Array[]
   ): Promise<CacheDictionaryGetFields.Response> {
     const request = new _DictionaryGetRequest();
     request.setDictionaryName(dictionaryName);
-    request.setFieldsList(fields);
+    request.setFieldsList(this.convertArrayToB64Strings(fields));
```
cprice404, line 1338: here - we need to convert the fields to b64 for the Request, but when we are constructing the response, we need to retain the un-encoded versions of them.

packages/client-sdk-web/test/integration/foo.test.ts
```diff
@@ -0,0 +1,27 @@
+import {CacheClient, CredentialProvider} from '../../src';
```
cprice404, line 1: same as the one above for nodejs, you can delete this file

packages/client-sdk-web/test/integration/integration-setup.ts
```diff
@@ -24,12 +28,18 @@ export function SetupIntegrationTest(): {
     const momento = momentoClientForTesting();
     await deleteCacheIfExists(momento, cacheName);
     const createResponse = await momento.createCache(cacheName);
+
+    // console.log(
+    //   `\n\n\nCACHE CREATE RESPONSE: ${createResponse.toString()}\n\n\n`
+    // );
+    await delay(5_000);
```
cprice404, line 35: shouldn't need this

packages/common-integration-tests/src/common-int-test-utils.ts
```diff
@@ -70,3 +70,19 @@ export function ItBehavesLikeItValidatesCacheName(
     );
   });
 }
+
+const bytesEncoderForTests = new TextEncoder();
+
+/**
+ * TODO
+ * TODO
+ * TODO explain
```
cprice404, line 79: This was the other major fix in this PR.  Something about how TextEncoder is populating the Uint8Array is different between nodejs SDK and web SDK; I suspect that the Uint8Array class can be backed by either an actual array of ints or some more fancy type of Buffer object, and that if you try to compare two Uint8Arrays that are using a different backing implementation, jest isn't smart enough to figure it out and you get the weird "object vs array" diffs that you were seeing, even though the bytes are the same.

You should document that this function ensures we end up with a Uint8Array backed by a simple array of ints, and that that is better for testing since it ensures Jest will be comparing the things we want it to compare.

You should also probably sweep through the shared int tests and get rid of any references to TextEncoder in favor of this function.  (You may need to add a similar helper for `decode` if any of the tests are relying on that.)

PR #460, opened 2023-04-26
chore: migrate common dict int tests

pgautier404:
This commit migrates dictionary integration tests to the shared integration tests package for reuse between the nodejs and web SDKs, adding implementations for each operation to the web SDK data client. It also removes methods from the nodejs CacheClient that are now defined in the AbstractCacheClient, fixes encoding of collection names in list operations, and adds a test to verify deleting list collections from the cache.


scripts/build-and-test-package.sh
```diff
@@ -11,6 +11,5 @@ echo "building and testing package: ${1}"
 ${ROOT_DIR}/scripts/build-package.sh ${1}
 
 pushd ${ROOT_DIR}/packages/${1}
-    npm run unit-test
-    npm run integration-test
+    npx jest --maxWorkers 1
```
cprice404, line 14: This was for me for debugging, I don't think we actually want to commit it.

cprice404 (commented):
Just one minor nit

scripts/build-and-test-package.sh
pgautier404: Ok, I'll revert it. It did, however, fix a persistent local problem I've been having with tests failing with `Rate exceeded` errors.

cprice404: Yeah it is useful for development.  Feel free to add another bash script that uses that flag, or another command in the package.json scripts section.  But that file that you had changed gets used by CI.

cprice404 (commented):
I'm good with this now but I would like to hold off on merging it until Matt has resolved the issue he's having with trying to use the web sdk from npm.js, because there is a user on discord waiting for that.

packages/client-sdk-web/package.json
```diff
@@ -15,7 +15,7 @@
   "scripts": {
     "prebuild": "eslint . --ext .ts",
     "test": "jest",
-    "unit-test": "jest unit",
+    "unit-test": "jest unit --passWithNoTests",
```
cprice404, line 18: why is this necessary?  If there is a file that doesn't have a test in it can we just delete that file?

pgautier404: Hm, actually, I think I need to put it back because the `build-and-test-package` script is explicitly calling it. This allows that script to call `unit-test` and allows `unit-test` to no-op.

pgautier404: Added back in 63af729 to support `build-and-test-package` script.

cprice404: I don't like the passWithNoTests flag.  Do we not have any unit tests?  Could we just add a "hello world" one that asserts true == true?

I have seen issues in the past where someone misconfigured a test source dir and then a flag like this would cause the build to pass even if there were tons of failing unit tests because they didn't actually end up getting run.

cprice404: to be clear - i want to keep the unit tests directive but i want to get rid of that flag :)

cprice404 (commented):
ready to go except I'm gonna be annoying about that passWithNoTests flag :)

packages/client-sdk-web/package.json
pgautier404: Yeah, that's fair. I added a no-op unit test in b579c3d.

cprice404 (approved):
hooray

PR #461, opened 2023-04-26
fix: verify response types before casting them

bruuuuuuuce:
We have seen some transient build errors that look like

```
    TypeError: respFetch.valueListString is not a function

      109 |         );
      110 |         expect(
    > 111 |           (respFetch as CacheListFetch.Hit).valueListString()
          |                                             ^
      112 |         ).toIncludeAllMembers(values);
      113 |       });
      114 |     };

      at Object.<anonymous> (../common-integration-tests/src/list.ts:111:45)
```

This is happening because in our integration tests, we are casting the response type before asserting the type of the response. This pr fixes this by asserting the response is of the right type before casting and verifying the contents of the response

### DIFF
--- a/packages/common-integration-tests/src/list.ts
+++ b/packages/common-integration-tests/src/list.ts
@@ -107,6 +107,7 @@ export function runListTests(
           IntegrationTestCacheName,
           listName
         );
+        expect(respFetch).toBeInstanceOf(CacheListFetch.Hit);
         expect(
           (respFetch as CacheListFetch.Hit).valueListString()
         ).toIncludeAllMembers(values);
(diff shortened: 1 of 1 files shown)


pgautier404 (approved):
Thanks!

packages/common-integration-tests/src/list.ts
```diff
@@ -107,6 +107,7 @@ export function runListTests(
           IntegrationTestCacheName,
           listName
         );
+        expect(respFetch).toBeInstanceOf(CacheListFetch.Hit);
```
cprice404, line 110: this is a major improvement but does this `expect` API give us a way to define a message that will get logged if the expectation fails?  Would be really valuable to log `respFetch.toString()` here if it fails

bruuuuuuuce: the error message will look something like

```
expected: CacheListFetch.Hit
received: CacheListFetch.Miss
```

cprice404: I'll +1 this, it's a big improvement, but if these keep failing I'd like to look into adding `jest-expect-message`: https://stackoverflow.com/a/53765479

cprice404 (approved):

PR #466, opened 2023-04-27
chore: migrate common set integration tests

pgautier404:
This commit migrates set integration tests to the shared integration tests package for reuse between the nodejs and web SDKs, adding implementations for each operation to the web SDK data client. It also removes methods from the nodejs CacheClient that are now defined in the AbstractCacheClient and adds a test to verify deleting set collections from the cache.


cprice404 (approved):
awesome!  almost to the finish line for collections!

almost there

PR #468, opened 2023-04-27
fix: package-lock.json invalid when installing core dependency for publishing

cprice404:
Prior to this commit the package-lock.json file for the SDK projects
is hard-coded to the 0.0.1 version of core; we then modify the package.json
file to depend on the new version, but attempt an `npm ci` without
updating the package-lock.json first.  In this commit we tweak the
publish logic to do an `npm install` on the appropriate version of
core before we do the `npm ci`.  This should cause the package-lock.json
file to get updated correctly prior to publishing.


bruuuuuuuce (approved):
:shipit:

PR #472, opened 2023-04-27
feat: add refresh token support to nodejs sdk

ben-kugler:
- add refresh token and generate token api's
- made test ignore auth.test.ts as these test will require a session token, no easy way right now to programmatically get a session token atm

Test run fine locally, hitting auth issues here
```
 npm ERR! code E401
npm ERR! Unable to authenticate, your authentication token seems to be invalid.
npm ERR! To correct this please trying logging in again with:
npm ERR!     npm login
```
image


CONTRIBUTING.md
```diff
@@ -61,6 +61,14 @@ export TEST_AUTH_TOKEN=<YOUR_AUTH_TOKEN>
 npm run test
 ```
 
+### Run auth integration tests
+
+```
+export TEST_AUTH_TOKEN=<YOUR_AUTH_TOKEN>
+export TEST_SESSION_TOKEN=<YOUR_SESSION_TOKEN>
+npm run integration-test-auth
```
cprice404, line 69: @bruuuuuuuce @rishtigupta next week we should talk about our options for making it possible to run these in CI

packages/client-sdk-nodejs/src/auth-client-props.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export interface AuthClientProps {
+  /**
+   * Configuration settings for the auth client
+   */
+  configuration: Configuration;
+  /**
+   * controls how the client will get authentication information for connecting to the Momento service
+   */
+  credentialProvider: CredentialProvider;
```
cprice404, line 12: I'm 90% sure that we don't want this field in our final public API for this.  That credential provider is for providing API tokens, which AFAIK will not be used for many/most of the APIs that people consume the AuthClient for.

Once this PR is merged we will either need to be happy committing to what is in it as a public API for AuthClient, since it would be included in the next node.js release, or we will be blocked on doing any more node.js releases until we file follow-up PRs to solidify the API.  My pref would be to try to get it all sorted in this PR - apologies that this is landing on you Ben, we were planning to do this stuff but you started working on this before we had time.

cprice404: cc: @rlinehan

packages/client-sdk-nodejs/src/auth-client.ts
```diff
@@ -0,0 +1,57 @@
...
+   *
+   * @param {string} sessionToken - The session token to allow access for generation of api tokens.
+   * @param {string} validForSeconds - How long the token is valid for in seconds.
+   * @returns {Promise<GenerateApiToken.Response>} -
+   * {@link GenerateApiToken.Success} containing the api token, refresh token, origin an how long the token is valid for.
+   * {@link GenerateApiToken.Error} on failure.
+   */
+  public async generateApiToken(
+    sessionToken: string,
+    validForSeconds?: number
```
cprice404, line 34: this is another thing that I'm not sure I want in our public API, see: 

https://github.com/momentohq/client-sdk-nodejs/issues/394

packages/client-sdk-nodejs/src/auth-client-props.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+/**
+ * @deprecated use {AuthClientProps} instead
+ */
+export type SimpleAuthClientProps = AuthClientProps;
```
cprice404, line 18: do we need to add this deprecated thing? :)

packages/client-sdk-nodejs/src/index.ts
```diff
@@ -1,5 +1,6 @@
 import {CacheClient, SimpleCacheClient} from './cache-client';
 import {TopicClient} from './topic-client';
+// import {AuthClient} from './auth-client';
```
cprice404, line 3: errant comment?

packages/client-sdk-nodejs/src/internal/auth-client.ts
```diff
@@ -0,0 +1,136 @@
...
+  private static readonly REQUEST_TIMEOUT_MS: number = 60 * 1000;
+  private readonly logger: MomentoLogger;
+
+  /**
+   * @param {AuthClientProps} props
+   */
+  constructor(props: AuthClientProps) {
+    this.logger = props.configuration.getLoggerFactory().getLogger(this);
+    const headers = [
+      new Header('Authorization', props.credentialProvider.getAuthToken()),
```
cprice404, line 39: This shouldn't be necessary for the Auth APIs, right?

cprice404: cc: @rlinehan @bruuuuuuuce

packages/client-sdk-nodejs/test/integration/auth.test.ts
```diff
@@ -0,0 +1,105 @@
...
+  it('should succeed for generating an api token that never expires', async () => {
+    const neverExpiresResponse = await Momento.generateApiToken(
+      SessionToken,
+      undefined
+    );
+    expect(neverExpiresResponse).toBeInstanceOf(GenerateApiToken.Success);
+    const neverExpireResponseSuccess =
+      neverExpiresResponse as GenerateApiToken.Success;
+    expect(neverExpireResponseSuccess.is_success);
+    expect(neverExpireResponseSuccess.validUntil).toEqual(undefined);
```
cprice404, line 36: The structure of this `success` response is another thing we need to think about the API for.  I suspect that we will want to use the same `ExpiresAt` or `ExpiresIn` type of object as I linked to in one of my comments above, as opposed to having `undefined` imply "never expires".  We can discuss this if people have thoughts but we have to get it right before we ship this.

packages/client-sdk-nodejs/test/integration/auth.test.ts
```diff
@@ -0,0 +1,105 @@
+import {
+  GenerateApiToken,
+  MomentoErrorCode,
+  RefreshApiToken,
+} from '@gomomento/sdk-core';
+import {SetupAuthIntegrationTest} from './integration-setup';
+
+const {Momento, SessionToken} = SetupAuthIntegrationTest();
```
cprice404, line 8: i know you are just copying an existing pattern here but it's a weird pattern that I want to clean up.  Can you make these names start with a lower-case letter, and instead of `Momento` can you call this `authClient`?  I would also love to see type annotations on these two variables to make it more obvious what they are.

packages/client-sdk-nodejs/test/integration/auth.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    expect(refreshSuccessRst.is_success);
+    expect(refreshSuccessRst.validUntil).toEqual(refreshSuccessRst.validUntil);
+  });
+
+  it("should not succeed for refreshing an api token that's expired", async () => {
+    const generateResponse = await Momento.generateApiToken(SessionToken, 1);
+    const generateSuccessRst = generateResponse as GenerateApiToken.Success;
+
+    await new Promise(r => setTimeout(r, 1000));
```
cprice404, line 80: supernit but there should be a `delay` function somewhere in this repo that you can just `await` directly, and it makes it a bit more readable than the setTimeout stuff.

packages/client-sdk-nodejs/test/integration/integration-setup.ts
```diff
@@ -45,6 +46,10 @@ function momentoClientForTesting() {
   return new CacheClient(IntegrationTestCacheClientProps);
 }
 
+function momentoAuthClientForTesting() {
```
cprice404, line 49: can we annotate return type here please

packages/client-sdk-nodejs/test/integration/integration-setup.ts
```diff
@@ -74,6 +79,21 @@ export function SetupIntegrationTest(): {
   return {Momento: client, IntegrationTestCacheName: cacheName};
 }
 
+export function SetupAuthIntegrationTest(): {
+  Momento: AuthClient;
+  SessionToken: string;
```
cprice404, line 84: same comment about the names of these vars

packages/core/src/messages/responses/refresh-api-token.ts
```diff
@@ -0,0 +1,62 @@
...
+    );
+  }
+
+  public getRefreshToken() {
+    return this.refreshToken;
+  }
+
+  public getValidUntil() {
+    return this.validUntil;
+  }
```
cprice404, line 38: explicity return type annotations on all of these please

cprice404 (changes requested):

packages/client-sdk-nodejs/src/internal/auth-client.ts
rlinehan: yep, shouldn't be necessary

packages/client-sdk-nodejs/test/integration/auth.test.ts
ben-kugler: For sure!

packages/client-sdk-nodejs/src/internal/auth-client.ts
ben-kugler: No it shouldn't, just a copy-paste from another client - removing!

packages/client-sdk-nodejs/src/auth-client-props.ts
cprice404: this is probably the biggest UX/API issue we need to sort out.  We need the control plane endpoint for this client, but not all of the APIs in this client require an API token.  We don't currently have a mechanism for a user to get access to the control plane endpoint without parsing it from an API token.

From a practical perspective this client should only require the endpoint uri to its constructor, but from a UX perspective I am on the fence about how they would discover that and get it into their code.

ben-kugler: We have the same issue for momento cli atm, we plan to just hardcode endpoints for now w/ custom option. Seems like we don't have a real solution without a endpoint discovery system, which it seems we are going to need really soon here. 

For now, seems best that we supply a helper `(aws/gcp, region): endpoint` in node sdk and have the `AuthClient` only accept a url.

packages/client-sdk-nodejs/package-lock.json
```diff
@@ -12534,7 +12534,9 @@
       "link": true
     },
     "node_modules/@gomomento/generated-types": {
-      "version": "0.57.0",
+      "version": "0.59.0",
+      "resolved": "https://momento-prod-401011790710.d.codeartifact.us-west-2.amazonaws.com/npm/npm-upstream/@gomomento/generated-types/-/generated-types-0.59.0.tgz",
```
bruuuuuuuce, line 12538: this is why you are getting an npm error. These are dependencies installed via our internal/private npm registries. We should point all deps to the public npm registry. To fix this, you can remove this file, remove your `~/.npmrc` file, and `npm i` to reinstall all deps using the public registry

ben-kugler:
Newest build fails, after fighting the linter this error I'm not entirely sure how to correct as they appear to be coming from the generated node modules.

cprice404 (changes requested):

packages/core/src/utils/expiration.ts
```diff
@@ -0,0 +1,153 @@
...
+   * Whether or not the token expires or not.
+   * @returns {boolean}
+   */
+  public doesExpire(): boolean {
+    return this._doesExpire;
+  }
+}
+
+export class ExpiresIn extends Expiration {
+  private readonly _expiresAtEpoch: number;
```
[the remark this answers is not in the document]
ben-kugler: Yeah, just forgot to remove that, updated!

cprice404:
@ben-kugler if you update the generated-types dep to the one that just got released from Matt's client-protos PR, I think it will fix the build issue.

packages/core/test/unit/utils/expiration.test.ts
```diff
@@ -0,0 +1,71 @@
...
+      expect(expiresIn.seconds()).toBe(Infinity);
+    });
+
+    test('expiresIn.seconds is valid', () => {
+      const epoch = Math.floor(new Date().getTime() / 1000 + 10);
+      const expiresIn = ExpiresIn.seconds(10);
+
+      expect(expiresIn.doesExpire()).toBe(true);
+      expect(expiresIn.seconds()).toBe(10);
+      expect(expiresIn.epoch()).toBeWithin(epoch - 1, epoch + 2);
```
cprice404: this shouldn't compile with your latest changes should it?

packages/client-sdk-nodejs/src/internal/auth-client.ts
```diff
@@ -0,0 +1,150 @@
...
+import Never = grpcAuth._GenerateApiTokenRequest.Never;
+import Expires = grpcAuth._GenerateApiTokenRequest.Expires;
+import {
+  ExpiresIn,
+  ExpiresAt,
+  CredentialProvider,
+} from '@gomomento/sdk-core/dist/src';
+
+export interface AuthClientProps {
+  configuration: Configuration;
```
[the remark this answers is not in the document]
ben-kugler: It is being used in the construction of `IdleGrpcClientWrapper` in `generateApiToken`/`refreshApiToken` calls and then we also log a little but here. 

I'm also now seeing errors from the `client-sdk-web` due to changes in the shared code, so guess I'm making some updates there to match the `client-node-sdk` changes.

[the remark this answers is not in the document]
cprice404: ah, I see.  You don't need that IdleGrpcClientWrapper for now, you can just use the grpc.AuthClient object directly.  So let's get rid of that and then I think my recommendation for now would be to remove the log statements and remove this props object.  That gives us the most flexibility in the future without committing to a specific API for config/logging settings for now.  lmk if that causes any trouble

packages/client-sdk-nodejs/src/auth-client-props.ts
```diff
@@ -0,0 +1,10 @@
+import {Configuration} from './config/configuration';
+
+export interface AuthClientProps {
```
cprice404: is this still used?

packages/client-sdk-nodejs/src/auth-client.ts
```diff
@@ -0,0 +1,65 @@
...
+import {
+  GenerateApiToken,
+  ExpiresIn,
+  CredentialProvider,
+} from '@gomomento/sdk-core';
+
+export class AuthClient extends AbstractAuthClient implements IAuthClient {
+  private readonly logger: MomentoLogger;
+
+  constructor(props: AuthClientProps) {
```
cprice404: I was thinking we could ditch this here in the public API for now, not just the internal one, sorry if that wasn't clear.

cprice404 (commented):
i don't see anything else left other than the AuthClientProps thing, I think I will be +1 as soon as we take that out

cprice404 (approved):

PR #477, opened 2023-04-28
chore: migrate common sorted set integration tests

pgautier404:
This commit migrates sorted set integration tests to the shared integration tests package for reuse between the nodejs and web SDKs, adding implementations for each operation to the web SDK data client. It also removes methods from the nodejs CacheClient that are now defined in the AbstractCacheClient and adds a test to verify deleting sorted set collections from the cache.


packages/client-sdk-web/src/internal/data-client.ts
```diff
@@ -1782,6 +1807,767 @@ export class DataClient<
...
+        {
+          ...this.authHeaders,
+          ...metadata,
+        },
+        (err, resp) => {
+          if (resp?.getMissing()) {
+            resolve(new CacheSortedSetGetRank.Miss());
+          } else if (resp?.getElementRank()) {
+            const rank = resp?.getElementRank()?.getRank();
+            if (!rank) {
```
pgautier404, line 2364: This differs from the nodejs client, where the test is for rank being `undefined`. However, in this client, we're getting back a 0 from `resp?.getElementRank()?.getRank()` when the value isn't in the set. Not sure why `getMissing()` isn't making the miss explicit here.

cprice404: cc: @wandaitzuchen @bruuuuuuuce . 

spec is here: https://www.notion.so/momentohq/SDK-API-Methods-Spec-cf12e4db2535405ca2acc3608814b083?pvs=4#fdace86a5cb349df98b37e8f39d91eea

cprice404 (approved):

packages/client-sdk-web/src/internal/data-client.ts
wandaitzuchen: Similar many other collection read APIs, `missing` means *the collection* does not exist; `element_rank` is like `found`, which means collection exists. An `element_rank` includes a `result` field, which is a `ECacheResult` specifying whether *the value* is a Hit or Miss. The semantics of `SortedSetGetRank` is pretty much the same as `SortedSetGetScore`, which, when the set exists, returns a `found` that includes a list of elements, each element includes a `result` field. 

See proto def for [SortedSetGetRank](https://github.com/momentohq/client-protos/blob/89d96a2cb369c5690ee73fe3aa6cd6af7f5b9792/proto/cacheclient.proto#L853-L868) and [SortedSetGetRank](https://github.com/momentohq/client-protos/blob/89d96a2cb369c5690ee73fe3aa6cd6af7f5b9792/proto/cacheclient.proto#L788-L805)

PR #480, opened 2023-04-28
feat: add npm packed files for core and sdk-web

rishtigupta:
## PR:
- Add npm packed files for core and sdk-web (Temporarily till we get the sdk published to npm sorted out)


cprice404 (changes requested):
oh @rishtigupta no, I think you need to add these in the console repo.  Let me see if i can find the commit where matt removed them.

PR #482, opened 2023-05-01
chore: make tests verbose

pgautier404:
This commit fixes a number of TTL-related test failures by increasing the TTL value of the CollectionTTL object involved in the tests from 1 second to 3 seconds. Presumably the low previous TTL wasn't giving the test itself enough latency breathing room, causing requests to exceed the TTL and produce false test failures.


bruuuuuuuce (approved):
:shipit:

PR #485, opened 2023-05-01
chore: make GenerateApiTokenResponse members private

bruuuuuuuce:
users should only be able to access the members from the response class by using the public getters, rather than accessing the members themselves


rishtigupta (approved):
🚢

cprice404 (approved):

cprice404:
thank you for catching this

PR #487, opened 2023-05-02
chore: add expect messages

pgautier404:
This commit adds an `expectWithMessage()` method that extends the built-in Jest `expect()` functionality by allowing users to supply extra details in a message for display when the assertion fails. For example, we can call `.toString()` on responses to get extra details about them:

```
expectWithMessage(() => {
    expect(result).toBeInstanceOf(CacheSortedSetGetRank.Hit);
}, `expected HIT but got ${result.toString()}`);
```


cprice404 (approved):
thank you, this doesn't look like much fun but I think it will be really useful in the future

PR #488, opened 2023-05-03
chore: add better test coverage for sortedSetGetRank

pgautier404:
An issue with `grpc-js` is causing the `SortedSetGetRank` operation to return `undefined` as the rank value for the topmost (0-th ranked according to sort order) item in the set. All other rank values are fine, and the same behavior is not observed with `grpc-web` and the web sdk.

This commit introduces a test to demonstrate the bug as well as a temporary workaround that detects the combination of a `hit` cache result and an undefined rank value and changes the rank to zero. This will prevent the top-ranked item in a sorted set from being mishandled as a cache miss. This change is intended to be removed when a root cause and remediation are found.

### DIFF
--- a/packages/common-integration-tests/src/sorted-set.ts
+++ b/packages/common-integration-tests/src/sorted-set.ts
@@ -1212,6 +1212,17 @@ export function runSortedSetTests(
         }, `expected HIT but got ${result.toString()}`);
         hitResult = result as CacheSortedSetGetRank.Hit;
         expect(hitResult.rank()).toEqual(2);
+
+        result = await Momento.sortedSetGetRank(
+          IntegrationTestCacheName,
+          sortedSetName,
+          'foo'
+        );
+        expectWithMessage(() => {
+          expect(result).toBeInstanceOf(CacheSortedSetGetRank.Hit);
+        }, `expected HIT but got ${result.toString()}`);
+        hitResult = result as CacheSortedSetGetRank.Hit;
+        expect(hitResult.rank()).toEqual(0);
       });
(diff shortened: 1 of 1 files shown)


krispraws:
I _think_ upgrading `protoc-gen-ts` to [0.8.5](https://github.com/thesayyn/protoc-gen-ts/releases/tag/0.8.5) in [momentohq/client-protos](https://github.com/momentohq/client-protos/blob/89d96a2cb369c5690ee73fe3aa6cd6af7f5b9792/javascript/package.json#L19) will likely fix this since it changes the accessors in the generated ts files to return default values. But it is a breaking change and I don't have enough context to know if it will affect anything else in our SDK.
`protoc-gen-ts v 0.6.0` (current version) generates `_SortedSetGetRankResponse::_RankResponsePart::get_rank()` as
```
            get rank() {
                return pb_1.Message.getField(this, 2) as number;
            }
```
`protoc-gen-ts v 0.8.5` generates
```
            get rank() {
                return pb_1.Message.getFieldWithDefault(this, 2, 0) as number;
            }
```

cprice404:
holding off on this until #494 is merged; then I want to run the new test without the workaround, to see if the thing that @krispraws found will solve it.  If not, then I think this hack will be the right way to go.

cprice404:
I rebased this on to my other branch where I was updating the protogen, and changed the name accordingly.  When running the tests locally, the fix from protogen made the new test pass so that is now the only change in this PR.

krispraws (approved):
:shipit:

PR #489, opened 2023-05-03
fix: use `buffer` package

poppoerika:
A user reported an issue where our nodejs SDK was throwing `throw new errors_1.InvalidArgumentError('failed to parse token')` with Node version 14.15.0 and 15.0.0.
This was due to using `btoa` and `atob` to encode/decode to base64 and these are deprecated in the node environment.

Tested with:
node version:

- 14.15.0,
- 15.0.0,
- 16.17.0

platform:

- console (specifically generate token where we use encodeToBase64 )
- nodejs - ran example as well as integration test and successfully ran.


packages/core/src/internal/utils/string.ts
```diff
@@ -8,6 +8,12 @@ are a few ways we can handle this.
 2. pull these functions out into their respective sdks
```
poppoerika, line 8: I will remove this comment once this PR is good to merge.

packages/core/src/internal/utils/string.ts
```diff
@@ -8,6 +8,12 @@ are a few ways we can handle this.
...
 3. maybe some 3rd party library can help? */
```
poppoerika, line 9: I will remove this comment once this PR is good to merge.

packages/core/src/internal/utils/string.ts
```diff
@@ -8,6 +8,12 @@ are a few ways we can handle this.
...
+import {Buffer} from 'buffer/index';
 
-export const encodeToBase64 = (str: string) => btoa(str);
+export const decodeFromBase64 = (base64: string) => {
+  return Buffer.from(base64, 'base64').toString('utf8');
+};
+
+export const encodeToBase64 = (str: string) => {
+  return Buffer.from(str, 'utf-8').toString('base64');
+};
```
malandis, line 19: Is the buffer package only needed for browser compatibility?

malandis (commented):
Only question is regarding the motivation for the change. Node.js has a built-in `Buffer` type, so I am not clear if the change is to support browser only?

packages/core/src/internal/utils/string.ts
poppoerika: The buffer package allows:
- encoding/decoding to base64 in a browser
- encoding/decoding to base64 with Node version 14 and 15. I have tested 14.15.0 and 15.0.0 specifically since those are the versions the user reported the issue with.

poppoerika: We were using `atob` and `btoa` from the built-in Buffer class for Node.js but the user reported that it didn't work with Node v 14 and 15 which @eaddingtonwhite and I were able to reproduce on our laptop.
@cprice404 also tested on arm/x86 EC2 with those Node versions, and that didn't work either.

malandis (approved):
💯

PR #494, opened 2023-05-10
feat: update to more modern version of protogen and grpc-js

cprice404:
This commit updates us to the latest versions of the generated
client-protos, which includes a newer version of protogen and
grpc-js.  This was inspired by this comment related to a bug
that we recently discovered:

https://github.com/momentohq/client-sdk-nodejs/pull/488#issuecomment-1534010405

Haven't run the new test case yet to confirm whether this addresses
it, but upgrading to the latest should be for the best regardless.


bruuuuuuuce (approved):
:shipit:

packages/client-sdk-nodejs/src/internal/data-client.ts
```diff
@@ -2213,6 +2214,12 @@ export class DataClient {
     sortedSetName: Uint8Array,
     value: Uint8Array
   ): Promise<CacheSortedSetGetRank.Response> {
+    console.log(`
```
krispraws, line 2217: Nit: Do we want this log message in the public sdk?

krispraws (approved):
Nitpick comment on the log message.

packages/client-sdk-nodejs/src/internal/data-client.ts
```diff
@@ -2226,9 +2233,12 @@ export class DataClient {
...
           (err, resp) => {
-            if (resp?.missing) {
+            if (
+              resp?.missing ||
+              resp?.element_rank?.result === ECacheResult.Miss
+            ) {
               resolve(new CacheSortedSetGetRank.Miss());
-            } else if (resp?.element_rank) {
+            } else if (resp?.element_rank?.result === ECacheResult.Hit) {
               if (resp?.element_rank.rank === undefined) {
```
krispraws, line 2242: Late thought - `if (resp?.element_rank.rank === undefined)` can never be true with the new protogen ts files if it is a hit, do we still need this check?

packages/client-sdk-nodejs/src/internal/data-client.ts
```diff
@@ -2213,6 +2214,12 @@ export class DataClient {
     sortedSetName: Uint8Array,
     value: Uint8Array
   ): Promise<CacheSortedSetGetRank.Response> {
+    console.log(`
+
+    SEND SORTED SET GET RANK YO
+
+    `);
+
```
cprice404, line 2222:
```suggestion
```

krispraws (approved):
:shipit:

PR #496, opened 2023-05-11
chore: rename *ApiToken -> *AuthToken

ben-kugler:


cprice404 (changes requested):
this lgtm, thanks for doing it.  one minor nit - can you change the commit message and the PR description to use `feat:` instead of `chore:`?  I'd rather not have this change get released as a `.z` version bump.

cprice404 (approved):

eaddingtonwhite (approved):

ben-kugler:
@cprice404 https://github.com/momentohq/client-sdk-nodejs/blob/main/packages/core/src/messages/responses/refresh-api-token.ts#L27-L31 is this something we also want to update? Discussion only around the method calls, but wanted to double check on renaming the response object fields.

cprice404:
> @cprice404 https://github.com/momentohq/client-sdk-nodejs/blob/main/packages/core/src/messages/responses/refresh-api-token.ts#L27-L31 is this something we also want to update? Discussion only around the method calls, but wanted to double check on renaming the response object fields.

Yes please!  If `grep -ir apitoken` yields any results we probably want to change all of them.  Thanks for catching that!

cprice404 (commented):
+1 on changing that value accessor as ben suggested

ben-kugler:
only things I can't rename is the grpc calls.

image

cprice404:
yep, generated code can't be changed, that's fine since it's not public.

cprice404 (approved):

eaddingtonwhite (approved):

PR #498, opened 2023-05-11
feat: Add type api

rishtigupta:
## PR Description: 
Add type api

## Test
`npm run build-and-run-tests` in both nodejs and web sdk

## Issue
closes https://github.com/momentohq/momento-console/issues/206


packages/core/src/messages/responses/item-type-api.ts
```diff
@@ -0,0 +1,76 @@
...
+ *   // of the Error class; e.g. `response.errorCode()`.
+ * }
+ * ```
+ */
+export abstract class Response extends ResponseBase {}
+
+class _Hit extends Response {
+  private readonly keyType: string;
+
+  constructor(keyType: string) {
```
bruuuuuuuce, line 34: I'm just wondering, should this `keyType` be an enum? Probably not because it would require us to update this enum if we returned new data types? And I could see us forgetting to do that. But it would make working with this api in typescript easier

packages/core/package.json
```diff
@@ -25,6 +25,7 @@
   "author": "",
   "license": "Apache-2.0",
   "devDependencies": {
+    "@gomomento/generated-types": "^0.61.0",
```
bruuuuuuuce, line 28: we should not depend on anything grpc inside of this core package

packages/core/src/messages/responses/item-type-api.ts
```diff
@@ -0,0 +1,79 @@
+import {SdkError} from '../../errors/errors';
+import {
+  ResponseBase,
+  ResponseError,
+  ResponseHit,
+  ResponseMiss,
+} from './response-base';
+import {cache_client} from '@gomomento/generated-types/dist/cacheclient';
```
bruuuuuuuce, line 8: these responses should be independent of any grpc code. For example, since you are importing the nodejs types here, this will not run in the web browser. I think the constructor still should take in a string, not a grcp response type

bruuuuuuuce: for example here https://github.com/momentohq/client-sdk-nodejs/blob/76b4b307feb8d197f503a294ab851486d302e900/packages/core/src/messages/responses/grpc-response-types.ts#L9-L14, we have our own enums that are independent of the enums inside of the protos, and these are the enums that the constructors are using https://github.com/momentohq/client-sdk-nodejs/blob/76b4b307feb8d197f503a294ab851486d302e900/packages/core/src/messages/responses/grpc-response-types.ts#L29

bruuuuuuuce: vs using the enums defined in the proto. Allows us to be more flexible down the line

packages/common-integration-tests/package.json
```diff
@@ -47,6 +47,7 @@
   },
   "dependencies": {
     "@gomomento/sdk-core": "file:../core",
+    "@gomomento/generated-types": "0.61.0",
```
cprice404, line 50: I don't think we should need this dependency - this seems like it is indicating that we have leaked something from one of the gRPC APIs into the tests.

packages/common-integration-tests/src/item-type.ts
```diff
@@ -0,0 +1,187 @@
...
+      itemTypeResponse = await Momento.itemType(
+        IntegrationTestCacheName,
+        new TextEncoder().encode(cacheKey)
+      );
+      expectWithMessage(() => {
+        expect(itemTypeResponse).toBeInstanceOf(ItemType.Hit);
+      }, `expected HIT but got ${itemTypeResponse.toString()}`);
+      hitResult = itemTypeResponse as ItemType.Hit;
+      expect(hitResult.getItemType()).toEqual(
+        _ItemGetTypeResponse.ItemType.SCALAR
```
cprice404, line 49: yeah, here you have leaked the generated enum from one of the two protos into the tests, we don't want this.  we will need to define our own enum in core and provide a mapping from each of the two sdks.

packages/core/package.json
cprice404: 💯 to what matt said

packages/core/src/messages/responses/item-type-api.ts
cprice404: yes, and in this case we should just define a public, first-class enum for these as part of this Response class.

cprice404 (changes requested):

packages/client-sdk-nodejs/package.json
```diff
@@ -50,7 +50,7 @@
     "uuid": "8.3.2"
   },
   "dependencies": {
-    "@gomomento/generated-types": "0.61.0",
+    "@gomomento/generated-types": "^0.61.0",
```
bruuuuuuuce, line 53:
```suggestion
    "@gomomento/generated-types": "0.61.0",
```

bruuuuuuuce (approved):
:shipit: LGTM, it would also be good to get @cprice404 to take another pass before we merge

packages/common-integration-tests/src/item-type.ts
```diff
@@ -0,0 +1,165 @@
+import {v4} from 'uuid';
+import {
+  ValidateCacheProps,
+  ItBehavesLikeItValidatesCacheName,
+  expectWithMessage,
+} from './common-int-test-utils';
+import {ICacheClient} from '@gomomento/sdk-core/dist/src/internal/clients/cache';
+import {ItemType} from '@gomomento/sdk-core';
+import {_ItemType} from '@gomomento/sdk-core/dist/src/messages/responses/grpc-response-types';
```
cprice404, line 9: What is the difference between these two imports?  `ItemType` vs `_ItemType`?  Are these both things that an end user might need to reference in their code?

packages/core/src/internal/clients/cache/ICacheClient.ts
```diff
@@ -265,4 +266,8 @@ export interface ICacheClient extends IControlClient {
     sortedSetName: string,
     values: string[] | Uint8Array[]
   ): Promise<CacheSortedSetRemoveElement.Response>;
+  itemType(
```
cprice404, line 269: In the specs, this API is called `itemGetType`.  If we want to revisit that naming discussion with the folks who participated in the design, we can; otherwise we need to use that name.

If we change that name then we will also need to change the response name to `ItemGetType.Response`, which might help us disambiguate the two names that I mentioned in my previous comment.

packages/core/src/messages/responses/item-type-api.ts
```diff
@@ -0,0 +1,78 @@
...
+ *   // `ItemType.Error` in this block, so you will have access to the properties
+ *   // of the Error class; e.g. `response.errorCode()`.
+ * }
+ * ```
+ */
+
+export abstract class Response extends ResponseBase {}
+
+class _Hit extends Response {
+  private readonly keyType: _ItemType;
```
cprice404, line 34: This is a value that an end user is going to have access to.  Therefore the type is something that the user will see and/or need to have access to in their code.  Therefore, it shouldn't start with an `_` and it shouldn't be defined in that `grpc-response-types` file, which contains internal classes that shouldn't be visible to users.

I think I would probably just define the user-facing enum directly in this file, or I would look and see where things like `CollectionTtl` are defined and put it in some file like that.

packages/core/src/messages/responses/item-type-api.ts
```diff
@@ -0,0 +1,78 @@
...
+  constructor(keyType: _ItemType) {
+    super();
+    this.keyType = keyType;
+  }
+
+  /**
+   * Returns the type of key.
+   * @returns string
+   */
+  public getItemType(): _ItemType {
```
cprice404, line 45: Can you double check some functions like `ListLength` and see if we are using `get` prefix for the accessors?  I feel like we are usually just calling them `length()`, etc., in which case this would just be `itemType()` - it's important for us to try to keep these APIs consistent.

cprice404 (changes requested):

cprice404 (approved):
:shipit:

PR #501, opened 2023-05-12
feat: rename sessionToken param to token

cprice404:
In the not-too-distant future we will need to support generating
auth tokens using other auth tokens.  This commit renames the
arguments in the various function signatures to just `token` so
that it won't be misnamed when we do support that.


rishtigupta (approved):

ben-kugler (approved):

bruuuuuuuce (approved):
:shipit:

eaddingtonwhite (approved):

rlinehan (approved):

PR #515, opened 2023-05-17
feat: introduce API for scoping generated token permissions

cprice404:
This commit tweaks the API for AuthClient.generateAuthToken a bit further, adding a new 'scope' argument that can be used to control the permissions of the generated token.

For now there is exactly one pre-built scope that we provide, currently called AllDataReadWrite.  For console use cases, it is possible to manually build up a list of restrictions rather than using this pre-built scope, and thus create a SuperUser token. End users who are consuming the API will not have permissions to do this, though, thus we do not currently provide a pre-built scope object for that.

This commit attempts to shape the API but does not yet wire up the actual functionality of requesting the token with the restricted scope; we can go ahead and merge the API changes now but the backing implementation is blocked until some server-side changes land.


packages/client-sdk-nodejs/src/auth-client.ts
```diff
@@ -14,36 +9,4 @@ export class AuthClient extends AbstractAuthClient implements IAuthClient {
...
-   *
-   * @param {string} controlEndpoint - Endpoint for control plane.
-   * @param {string} token - The token to allow access for generation of api tokens.
-   * @param {string} expiresIn - How long the token is valid for in epoch timestamp.
-   * @returns {Promise<GenerateAuthToken.Response>} -
-   * {@link GenerateAuthToken.Success} containing the api token, refresh token, origin and epoch timestamp when token expires.
-   * If the token never expires, then no refresh token will be returned and expires at timestamp will be infinite.
-   * {@link GenerateAuthToken.Error} on failure.
-   */
-  public async generateAuthToken(
```
cprice404, line 29: i changed these to just properties instead of functions, based on some feedback from @allenheltondev in the past.

packages/client-sdk-nodejs/src/internal/internal-auth-client.ts
```diff
@@ -35,6 +36,7 @@ export class InternalAuthClient implements IAuthClient {
   }
 
   public async generateAuthToken(
+    scope: TokenScope,
```
cprice404, line 39: new argument required in this API

packages/common-integration-tests/src/auth-client.ts
```diff
@@ -4,8 +4,11 @@ import {
   MomentoErrorCode,
   RefreshAuthToken,
 } from '@gomomento/sdk-core';
-import {IAuthClient} from '@gomomento/sdk-core/dist/src/internal/clients/auth/IAuthClient';
+import {IAuthClient} from '@gomomento/sdk-core/dist/src/clients/IAuthClient';
 import {expectWithMessage} from './common-int-test-utils';
+import {Restrictions} from '@gomomento/sdk-core/dist/src/auth/tokens/token-scope';
+
+const SUPER_USER_RESTRICTIONS: Restrictions = {restrictions: []};
```
cprice404, line 11: Here's how I can express SuperUser restrictions for use cases like the console, without exposing directly to users.

packages/common-integration-tests/src/auth-client.ts
```diff
@@ -80,13 +101,16 @@ export function runAuthClientTests(
...
       expect(refreshSuccessRst.is_success);
 
-      expect(generateSuccessRst.expiresAt.epoch()).toEqual(
-        refreshSuccessRst.getExpiresAt().epoch()
-      );
+      const expiresAtDelta =
+        refreshSuccessRst.expiresAt.epoch() -
+        generateSuccessRst.expiresAt.epoch();
+
+      expect(expiresAtDelta).toBeGreaterThanOrEqual(delaySecondsBeforeRefresh);
```
cprice404, line 108: hrm @rlinehan I should probably also add a lessThanOrEqual assertion here

packages/core/src/auth/tokens/token-scope.ts
```diff
@@ -0,0 +1,35 @@
...
+  topicRole: TopicRole;
+}
+
+export type Restriction = CacheRestriction | TopicRestriction;
+
+export interface Restrictions {
+  restrictions: Array<Restriction>;
+}
+
+export class AllDataReadWrite implements Restrictions {
```
kvcache, line 25: is it better to `export const ALL_DATA_READ_WRITE: Restrictions = { restrictions: [ ... ] }`?

kvcache (commented):
seems like a fine UX to me

rlinehan (commented):
I think this looks good to me. I already expressed my naming concern about cache and topics read/write, and can't come up with anything better right now so 👍 . I like the name `AllDataReadWrite`, and I like the hiding of `SuperUser`.

packages/core/src/auth/tokens/token-scope.ts
rlinehan: > So e.g. I could create a class called `SingleCacheReadOnly` and then have it take a constructor arg `cacheName`.

Just want to also call out that we will probably want to support a case where we want something like `SingleCacheReadOnly` for cache foo but `SingleCacheReadWrite` for cache bar.

kvcache: seems to me like a builder that returns a `Restrictions` after all the optional friendly named configurables were provided would be the future I'd steer toward. Constants don't seem too terribly incongruous with builders in my experience, so that's where my head has been

packages/core/src/auth/tokens/token-scope.ts
```diff
@@ -0,0 +1,35 @@
+export enum CacheRole {
```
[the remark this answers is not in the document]
allenheltondev: Does `ReadWrite` exclude control plane operations?

packages/core/src/auth/tokens/token-scope.ts
```diff
@@ -0,0 +1,35 @@
...
+  ReadWrite = 'readwrite',
+}
+
+export interface TopicRestriction {
+  topicRole: TopicRole;
+}
+
+export type Restriction = CacheRestriction | TopicRestriction;
+
+export interface Restrictions {
```
allenheltondev, line 21: This feels backward to me, at least with the naming. I'd hope that we are explicitly allowing access, instead of explicitly denying access to actions. 

The way this reads is that I pass in the things that are not allowed. If I'm misreading, I think renaming to something like `Permissions` would help with intuitiveness.

cprice404: lol.  This same convo/debate was happening while they were ironing out what to name these fields on the server-side, and in this PR I just copied what they called them on the server side for now.  But we could change it back, either here or in both places.

One other thing to keep in mind Allen is that for the most part I wouldn't expect an end user to ever mess with these directly.  We would want to give them pre-built things like `AllDataReadWrite`, and composable things (in the future) like `SingleCacheReadOnly`, that they can chain together to build up a Scope w/o ever having to see the word Restriction.  Only advanced use cases should ever need to go that deep.

Also of note: the `SuperUser` role is achieved by providing an empty list of Restrictions.  So I do think that makes some sense and is why they chose the current terminology.  It would seem a little weird to create a SuperUser role by providing an empty list of permissions.  But this is another thing that we don't actually intend to surface to users because they won't have permission to create such a token for now anyway.

All that said, I have no major skin in the game in the naming choice between `Permission` and `Restriction`.

@kvcache @rlinehan do you all want to weigh in on this at all?

cprice404: Also @allenheltondev since you can't directly see it in the PR, this is what I think it will look like in your web app, to create your tokens for browsers:

```
authClient.generateAuthToken(new AllDataReadWrite(), ExpiresIn.hours(2));
```

packages/common-integration-tests/src/auth-client.ts
cprice404: did this in https://github.com/momentohq/client-sdk-nodejs/pull/515/commits/7240edfe018b4d380f56f1c8d91a2f15aead6d85

packages/core/src/auth/tokens/token-scope.ts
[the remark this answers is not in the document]
cprice404: correct, `AllDataReadWrite` can do any data plane operations but no control plane operations.  Open to better name ideas if you have them.

packages/core/src/auth/tokens/token-scope.ts
cprice404: renamed from Restrictions to Permissions in https://github.com/momentohq/client-sdk-nodejs/pull/515/commits/477a5d4f15b06648bd5b7d226e055de793c87364

cprice404:
@rlinehan @allenheltondev I think this is ready for review/merge now.  The new permissions stuff is wired up to the server, and I've added integration tests validating which types of tokens can be used to generate new tokens.

packages/common-integration-tests/src/auth-client.ts
```diff
@@ -106,6 +134,118 @@ export function runAuthClientTests(
...
+
+    it('should support generating an AllDataReadWrite token when authenticated via a session token', async () => {
+      const generateResponse = await sessionTokenAuthClient.generateAuthToken(
+        AllDataReadWrite,
+        ExpiresIn.seconds(1)
+      );
+      expect(generateResponse).toBeInstanceOf(GenerateAuthToken.Success);
+    });
+
+    it('should support generating a superuser token when authenticated via a v1 superuser token', async () => {
```
rlinehan, line 154: ohhhhh wait a second. I think... we don't want this. Drat.

I think a v1 token can only generate another v1 token with less permissions 😓 . need to fix this on the server side probably.

rlinehan (approved):
I think this is fine, just found a thing on the server side that needs to be changed (and then a test change that here will need to be changed). I don't think that blocks this from going in.

allenheltondev (approved):

cprice404:
> I think this is fine, just found a thing on the server side that needs to be changed (and then a test change that here will need to be changed). I don't think that blocks this from going in.

okeydoke, just lmk when you are ready for me to fix it

PR #516, opened 2023-05-17
chore: move ping functionality into cache client

pgautier404:
This commit removes the public ping client and moves it into the cache client.


packages/client-sdk-web/src/cache-client.ts
```diff
@@ -51,3 +54,20 @@ function createDataClient(props: CacheClientProps): IDataClient {
     defaultTtlSeconds: 60,
   });
 }
+
+function createPingClient(props: CacheClientProps): IPingClient {
+  // remove the 'cache.' portion of the data endpoint to get the ping URL
+  const cacheEndpoint = props.credentialProvider.getCacheEndpoint();
+  const pingUrl = cacheEndpoint.split('.').slice(1).join('.');
```
bruuuuuuuce, line 61: 🙈

bruuuuuuuce: maybe we can add a `getPingEndpoint()` function to the creds provider?

pgautier404: Considering that the [ping client constructor just reconstructs the cache endpoint](https://github.com/momentohq/client-sdk-nodejs/blob/d7b79455c4cd4844839309f878bea42abc9d6241/packages/client-sdk-web/src/internal/ping-client.ts#L34), I'd love to just pass the cache endpoint instead and just have the constructor add the `https://` and `443` bits. The fact that it wasn't already done this way made me think that might be problematic for some reason. If you don't have an objection to that, I'm gonna go for it.

cprice404: yes please

packages/client-sdk-web/src/cache-client.ts
```diff
@@ -19,7 +21,8 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
   constructor(props: CacheClientProps) {
     const controlClient: IControlClient = createControlClient(props);
     const dataClient: IDataClient = createDataClient(props);
-    super(controlClient, [dataClient]);
+    const pingClient: IPingClient = createPingClient(props);
```
cprice404, line 24: can we do this lazily?  i.e., the class can have a nullable local variable for it, and then you can create an accessor function for it that checks to see if it's undefined and instantiates the client once if it is undefined.  That way we aren't opening an extra socket connection for every instantiated CacheClient when 90% of the time people aren't going to use the ping API?

packages/core/src/internal/clients/cache/AbstractCacheClient.ts
```diff
@@ -65,26 +65,39 @@ import {
...
 import {IControlClient} from './IControlClient';
 import {IDataClient} from './IDataClient';
+import {IPingClient} from './IPingClient';
 
 export abstract class AbstractCacheClient implements ICacheClient {
   // making these protected until we fully abstract away the nodejs client
   protected readonly controlClient: IControlClient;
   protected readonly dataClients: IDataClient[];
+  // TODO: Make pingClient required if and when the nodejs side starts adding
+  //  one as well
```
cprice404, line 75: can you just go ahead and add it to the node.js sdk while you are in here?  but I do think you still want this variable to be `?` for the reasons I mentioned above.  And then this variable becomes private, and you provide a protected accessor function that lazily instantiates the client when it is called for the first time.

cprice404 (approved):
this lgtm.  I have a few requests about node.js SDK and instantiating this client lazily rather than eagerly, but this looks like it fixes the public API for the part that I was concerned about, so please feel free to merge as-is and file a follow-up ticket for the rest of the work that I described.  we can prioritize that work after all of the web SDK launch blockers are done.

packages/core/src/internal/clients/cache/AbstractCacheClient.ts
pgautier404: Nice, makes sense, will do.

PR #518, opened 2023-05-18
feat: Add web sdk configurations

nand4011:
Move transport-strategy and grpc-configuration out of core since it contains options that are only useful for the node sdk.

Create web versions of transport strategy and grpc configuration.

Create web version of configurations.

Plumb deadline millis into the web sdk.


packages/client-sdk-web/src/utils/web-client-utils.ts
```diff
@@ -7,6 +7,10 @@ export function convertToB64String(v: string | Uint8Array): string {
...
 }
 
-export function createMetadata(cacheName: string): {cache: string} {
-  return {cache: cacheName};
+export function createMetadata(
+  cacheName: string,
+  timeoutMillis: number
+): {cache: string; deadline: string} {
+  const deadline = Date.now() + timeoutMillis;
+  return {cache: cacheName, deadline: deadline.toString()};
```
nand4011, line 15: According to [https://github.com/grpc/grpc-web#setting-deadline](url) the correct way of setting the deadline for grpc web is adding it to the metadata. Someone more familiar with pubsub should make sure that makes sense for that client.

packages/client-sdk-web/src/cache-client.ts
```diff
@@ -1,70 +1,32 @@
...
-}
+import {CacheClientProps} from './cache-client-props';
 
 export class CacheClient extends AbstractCacheClient implements ICacheClient {
   constructor(props: CacheClientProps) {
     const controlClient: IControlClient = createControlClient(props);
     const dataClient: IDataClient = createDataClient(props);
-    const pingClient: IPingClient = createPingClient(props);
-    super(controlClient, [dataClient], pingClient);
+    super(controlClient, [dataClient]);
```
pgautier404, line 15: It looks like my [ping client work](https://github.com/momentohq/client-sdk-nodejs/pull/516) got removed. Can you please add it back in here and restore the `createPingClient` function below as well? They're both likely to get moved out of this file when we work on [lazy loading the ping client](https://github.com/momentohq/client-sdk-nodejs/issues/517), but in the meantime we need to get it added back.

nand4011: I must have overwritten it when I merged that in. I'll add it back.

packages/client-sdk-web/src/utils/web-client-utils.ts
cprice404: i think it's right

packages/client-sdk-web/src/config/configurations.ts
```diff
@@ -0,0 +1,152 @@
...
+const defaultLoggerFactory: MomentoLoggerFactory =
+  new DefaultMomentoLoggerFactory();
+
+/**
+ * Laptop config provides defaults suitable for a medium-to-high-latency dev environment.  Permissive timeouts, retries, and
+ * relaxed latency and throughput targets.
+ * @export
+ * @class Laptop
+ */
+export class Laptop extends CacheConfiguration {
```
cprice404, line 22: For web, the use cases are very different from the other SDKs.  It's probably okay to keep a Laptop config for developers to use when testing.  But we probably don't need an InRegion config, at least not yet.  What we should add instead is a Browser config.  and the timeouts for the browser config should be lenient, probably the same as Laptop.

cc: @eaddingtonwhite would like your opinion.

cprice404 (commented):
nice!  lgtm other than:

1. Remove InRegion configs and add a browser config
2. would be nice to have at least one integration test that asserts a timeout error?  I thought i saw one somewhere that was commented out but i might be making that up.

packages/client-sdk-web/src/config/configurations.ts
nand4011: Makes sense. Do we care about low latency or should we just have Laptop and Browser?

cprice404: let's just start with laptop and browser imo

pgautier404 (approved):
LGTM!

PR #522, opened 2023-05-19
chore: add int tests to verify permissions of AllDataReadWrite tokens

cprice404:
This commit adds a token-scope shared integration test, which
creates a new AllDataReadWrite token, and then creates a cache
client using that token.  Then we make some positive and negative
assertions to validate that the client has access to the operations
we expect, and gets rejected from other ones.


rlinehan (approved):

cprice404:
rebased

packages/common-integration-tests/src/token-scope.ts
```diff
@@ -0,0 +1,73 @@
...
+      allDataReadWriteClient = cacheClientFactory(allDataReadWriteToken);
+    });
+    it('cannot create a cache', async () => {
+      const createCacheResponse = await allDataReadWriteClient.createCache(
+        'FOOFOOFOO'
+      );
+      expect(createCacheResponse).toBeInstanceOf(CreateCache.Error);
+      const createCacheError = createCacheResponse as CreateCache.Error;
+      expect(createCacheError.errorCode()).toEqual(
+        MomentoErrorCode.AUTHENTICATION_ERROR
```
rlinehan, line 42: I just merged the change that changes the grpc status from `UNAUTHENTICATED` to `PERMISSION_DENIED`

cprice404: okay well we can fix that on a subsequent PR when we see it fail?

rlinehan: sure I just think it might fail as soon as you merge this in to main 😅

rlinehan (approved):

PR #525, opened 2023-05-22
fix: remove maxWorkers=1 flag for prod

cprice404:
I added this flag a while back because without it, too many concurrent
tests will run on developer laptops and you get throttling errors.
I had added maxWorkers=1 to the script to allow the tests to pass on
my laptop, but did not mean to commit that for prod.

### DIFF
--- a/scripts/build-and-test-package.sh
+++ b/scripts/build-and-test-package.sh
@@ -11,5 +11,5 @@ ${ROOT_DIR}/scripts/build-package.sh ${1}
 
 pushd ${ROOT_DIR}/packages/${1}
     npm run unit-test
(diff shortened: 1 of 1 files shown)


pgautier404 (approved):

nand4011 (approved):
I think we should consider having max workers be 1 by default and having the github run supply an argument for more parallelism. It was confusing that the tests failed locally before I learned about maxWorkers.

cprice404:
> I think we should consider having max workers be 1 by default and having the github run supply an argument for more parallelism. It was confusing that the tests failed locally before I learned about maxWorkers.

i'd be fine with a PR that makes that happen.

PR #528, opened 2023-05-22
chore: Differentiate web from nodejs agent

nand4011:
Change web sdk agent to 'js-web'.

Switch from an interceptor to metadata for supplying the agent string
since interceptors don't work for the majority of clients.

Make a client metadata provider that provides the agent and
authorization headers.

Combine the deadline and cache name metadata into one call.


cprice404:
Did it look like everything was coming through as expected in lightstep?

nand4011:
Now that I've switched away from interceptors to metadata I can see the results in lightstep.

cprice404 (approved):
nice cleanup, thanks.

nand4011:
I'm still digging.

PR #530, opened 2023-05-22
feat: Add missing imports to client-sdk-web

rishtigupta:
## PR Description:
Add missing imports to client-sdk-web
1. ExpiresIn
2. ExpiresAt
3. AllDataReadWrite
4. TokenScope
5. InternalSuperUserPermissions

Done so that they can be directly accessed from @gomomento/client-sdk-web rather than from @gomomento/sdk-core in the console repo.


packages/client-sdk-web/src/index.ts
```diff
@@ -107,6 +113,9 @@ export {
   CredentialProvider,
   StringMomentoTokenProvider,
   EnvMomentoTokenProvider,
+  AllDataReadWrite,
+  InternalSuperUserPermissions,
```
cprice404, line 117: This one is private, we don't want to export it from the main index.  You could put another file at `src/internal/utils/auth.ts` and export it from there, and then when you import it from the console you will need to import it from `dist/src/internal/utils/auth`.  But we don't want to advertise this one in the main index.

packages/client-sdk-web/src/utils/auth.ts
```diff
@@ -0,0 +1,4 @@
+import {InternalSuperUserPermissions} from '@gomomento/sdk-core/dist/src/internal/utils/auth';
+import {AllDataReadWrite} from '@gomomento/sdk-core';
```
cprice404, line 2: lol I'm sorry... AllDataReadWrite is indeed public.  It's just the `Internal` one that is not.  but otherwise this looks good

cprice404 (approved):
:ty:

PR #537, opened 2023-05-24
fix: get rid of vendored printf dependency for logging in web sdk

cprice404:
Prior to this commit, we defined the Default logger in the core package and shared it between the two SDKs.  However, the default logger we were using in node.js depends on a vendored printf library that uses node's `Stream` type, which is not available in browsers. This commit moves the vendored printf code and the default logger implementation out of the core package and into the nodejs package, and introduces a distinct default logger for the web sdk.

For now the logger in the web SDK just doesn't handle the printf formatting; we can fix that in a future release.


bruuuuuuuce (approved):
:shipit:

PR #542, opened 2023-05-24
chore: add API examples for auth APIs, to be used in docs

cprice404:
This commit updates the docs examples file to include examples for the two new auth APIs.  These will be injected into the docs website in various places.

### DIFF
--- a/examples/nodejs/doc-examples-js-apis.ts
+++ b/examples/nodejs/doc-examples-js-apis.ts
@@ -19,7 +19,22 @@ import {
   CacheSet,
   CacheGet,
   CacheDelete,
+  AuthClient,
+  AllDataReadWrite,
+  ExpiresIn,
+  GenerateAuthToken,
+  RefreshAuthToken,
 } from '@gomomento/sdk';
+
+function example_API_InstantiateCacheClient() {
+  new CacheClient({
+    configuration: Configurations.Laptop.v1(),
+    credentialProvider: CredentialProvider.fromEnvironmentVariable({
+      environmentVariableName: 'MOMENTO_AUTH_TOKEN',
+    }),
+    defaultTtlSeconds: 60,
+  });
+}
 
 async function example_API_CreateCache(cacheClient: CacheClient) {
   const result = await cacheClient.createCache('test-cache');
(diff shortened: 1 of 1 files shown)


bruuuuuuuce (approved):
:shipit:

examples/nodejs/doc-examples-js-apis.ts
```diff
@@ -103,7 +118,43 @@ async function example_API_Delete(cacheClient: CacheClient) {
...
+    }),
+  });
+}
+
+async function example_API_GenerateAuthToken(authClient: AuthClient) {
+  const generateTokenResponse = await authClient.generateAuthToken(AllDataReadWrite, ExpiresIn.minutes(30));
+  if (generateTokenResponse instanceof GenerateAuthToken.Success) {
+    console.log('Auth token generated!');
+    console.log(`Auth token starts with: ${generateTokenResponse.authToken.substring(0, 10)}`);
+    console.log(`Refresh token starts with: ${generateTokenResponse.refreshToken.substring(0, 10)}`);
```
rlinehan, line 134: Initially I was confused as to why we were just logging what the token starts with, but I'm guessing it's because we don't want to print the whole thing for security? I guess I'm slightly worried that people are going to thing the `substring(0,10)` has some kind of meaning?

examples/nodejs/doc-examples-js-apis.ts
```diff
@@ -103,7 +118,43 @@ async function example_API_Delete(cacheClient: CacheClient) {
...
+    credentialProvider: CredentialProvider.fromEnvironmentVariable({
+      environmentVariableName: 'MOMENTO_AUTH_TOKEN',
+    }),
+  });
+}
+
+async function example_API_GenerateAuthToken(authClient: AuthClient) {
+  const generateTokenResponse = await authClient.generateAuthToken(AllDataReadWrite, ExpiresIn.minutes(30));
+  if (generateTokenResponse instanceof GenerateAuthToken.Success) {
+    console.log('Auth token generated!');
```
rlinehan, line 132: I wonder if it would be worthwhile to make it more obvious that this is an `AllDataReadWrite` token? Maybe like `AllDataReadWrite auth token generated` or `Auth token generated with AllDataReadWrite scope`?

rlinehan (approved):
A couple comments for how we could maybe improve clarity, but feel free to ignore if you don't agree. Looks good.

examples/nodejs/doc-examples-js-apis.ts
cprice404: https://github.com/momentohq/client-sdk-javascript/pull/542/commits/dc0615785b66c8d18d2e4d3e217d8c42efb71353

examples/nodejs/doc-examples-js-apis.ts
cprice404: https://github.com/momentohq/client-sdk-javascript/pull/542/commits/dc0615785b66c8d18d2e4d3e217d8c42efb71353

rlinehan (approved):

ben-kugler (approved):

PR #552, opened 2023-05-26
chore: Adding config for Inregion.Defualt.latest

NoSQLKnowHow:


examples/nodejs/doc-examples-js-apis.ts
```diff
@@ -50,6 +50,10 @@ function example_API_ConfigurationInRegionDefault() {
   Configurations.InRegion.Default.v1();
 }
 
+function example_API_ConfigurationInRegionDefaultLatest() {
+  Configurations.InRegion.Default.latest();
```
cprice404, line 54: these warnings above are indicating that this function is never called.  at the bottom of this file you'll see a `main` function that calls all of the other functions.  this is how we can make sure the code runs properly when our github CI jobs are executed.

cprice404 (approved):
:ty:

PR #557, opened 2023-05-30
feat: Add observability example

nand4011:
Add an example that uses OpenTelemetry to automatically capture grpc traces and exports them to Zipkin, as well as sending request count metrics to Prometheus via a middleware.


examples/nodejs/README.md
```diff
@@ -79,6 +79,30 @@ MOMENTO_AUTH_TOKEN=<YOUR AUTH TOKEN> npm run topic-subscribe my-cache dogs
...
+
+To run the observability example:
+
+```bash
+# Run the docker compose file to start Prometheus and Zipkin
+docker-compose up -d
+```
+```bash
+# Run the observability example
+MOMENTO_AUTH_TOKEN=<YOUR AUTH TOKEN> npm run observability
```
cprice404, line 97: we should add a few notes here about where to point your browser in order to check out the examples.

examples/nodejs/observability.ts
```diff
@@ -0,0 +1,55 @@
+import setupObservability from './utils/instrumentation';
+setupObservability('observability-example');
```
cprice404, line 2: I might add a comment explaining what this line is doing behind the scenes.

Also in a subsequent PR I think we might want to consider splitting these out to two separate examples since the instrumentation logic is so different. When they are combined I wonder if users won't understand which parts are providing metrics vs. traces.

examples/nodejs/README.md
cprice404: maybe even some screenshots?

nand4011: Done.

cprice404 (approved):

cprice404 (approved):

PR #570, opened 2023-06-02
chore: Adding mongodb-examples and README files

NoSQLKnowHow:
Adding MongoDB examples and some readme files.


examples/nodejs/mongodb-examples/simple-read-aside/jest.config.ts
```diff
@@ -0,0 +1,14 @@
+import type {Config} from 'jest';
```
cprice404, line 1: There are some files that I would expect to see in this directory that aren't here in this PR.  Namely, at least:

- `package.json`
- `package-lock.json`
- `tsconfig.json`

Also, not necessarily required, but we generally want to have some `.eslint*` and `.prettier` files in here to enforce consistent code style.

Do you have a mechanism for allowing users to run without the files in my list above, though?

examples/nodejs/mongodb-examples/simple-read-aside/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "dependencies": {
```
cprice404, line 2: i would expect to see a `scripts` section in here.

it might not be absolutely necessary but it usually helps developers figure out what options they have for how to build/consume the project.

Can you amend your README to include the npm/npx commands you are expecting users to run?  then i can pull this down and try it out and let you know whether i think the `scripts` section would be valuable.

PR #572, opened 2023-06-09
feat: create new TopicConfiguration class for configuring TopicClient

cprice404:
Prior to this commit, the TopicClient was using a CacheClient
Configuration object for configuration. There are many settings
on the CacheClient configuration that aren't relevant for Topics,
and soon the reverse will likely be true as well. This commit changes
the TopicClient constructor to take a new TopicConfiguration object
which will give us the ability to define the configuration API for the
two clients independently going forward.

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/pubsub-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/pubsub-client.ts
@@ -3,7 +3,6 @@ import grpcPubsub = pubsub.cache_client.pubsub;
 // older versions of node don't have the global util variables https://github.com/nodejs/node/issues/20365
 import {Header, HeaderInterceptorProvider} from './grpc/headers-interceptor';
 import {ClientTimeoutInterceptor} from './grpc/client-timeout-interceptor';
-import {createRetryInterceptorIfEnabled} from './grpc/retry-interceptor';
 import {cacheServiceErrorMapper} from '../errors/cache-service-error-mapper';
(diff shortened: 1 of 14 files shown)


malandis (approved):

packages/client-sdk-web/src/internal/pubsub-client.ts
```diff
@@ -1,9 +1,7 @@
 import * as pubsub from '@gomomento/generated-types-webtext/dist/CachepubsubServiceClientPb';
```
eaddingtonwhite, line 1: not related to this pr but we should probably rename this client purge pubsub and just call this topic.

packages/client-sdk-nodejs/src/internal/pubsub-client.ts
```diff
@@ -3,7 +3,6 @@ import grpcPubsub = pubsub.cache_client.pubsub;
 // older versions of node don't have the global util variables https://github.com/nodejs/node/issues/20365
```
eaddingtonwhite, line 3: same thought here s/pubsub/topic/ can be another pr.

packages/client-sdk-web/src/internal/pubsub-client.ts
cprice404: it's internal, so not public API, but sure we could do that

eaddingtonwhite (approved):

PR #573, opened 2023-06-09
feat: export new TopicConfiguration namespaces from index

cprice404:
Prior to this commit we weren't exporting the topic config namespaces
from the index, so they weren't easy to use from outside of the SDK.


malandis (approved):
🍰

PR #574, opened 2023-06-09
chore: update examples to 1.25.1

cprice404:
This includes updating the TopicClient examples to use the new
TopicConfiguration API.


eaddingtonwhite (approved):

examples/nodejs/cache/doc-example-files/doc-examples-js-apis.ts
```diff
@@ -772,7 +773,7 @@ function example_API_InstantiateAuthClient() {
 
 function example_API_InstantiateTopicClient() {
   new TopicClient({
-    configuration: Configurations.Laptop.v1(),
+    configuration: TopicConfigurations.Default.latest(),
```
malandis, line 776: Do we want the examples to demonstrate `v1` vs `latest` so the end user biases toward an unchanging config?

malandis (commented):
Looks fine just one question avout `v1` vs `latest`

examples/nodejs/cache/doc-example-files/doc-examples-js-apis.ts
cprice404: i have not introduced a 'v1' for this yet because I don't really know what things we will care about for this in the future. the important thing is just to get this change out there so that users aren't trying to still use the CacheClient configuration.

PR #587, opened 2023-06-15
chore: Allow Web SDK integration tests to be run against localhost

cprice404:
This commit makes some refactors to the CredentialProvider, as well
as a few minor test changes, to allow the integration tests to be
run against localhost. Changes include:

* Keep track of whether the endpoints for CredentialProvider were
  overridden at construction time so that we can branch accordingly.
* Don't add `web` prefix if the endpoints were explicitly overridden.
* Add support for a `MOMENTO_SDK_TESTS_USE_LOCALHOST` in the integration
  tests; if provided it will override the endpoints to localhost, and
  cause most control plane operations to be skipped.

### DIFF
--- a/packages/client-sdk-web/src/utils/web-client-utils.ts
+++ b/packages/client-sdk-web/src/utils/web-client-utils.ts
@@ -23,7 +23,7 @@ export function getWebControlEndpoint(
   if (credentialProvider.isControlEndpointOverridden()) {
     return withProtocolPrefix(credentialProvider.getControlEndpoint());
   }
-  return withProtocolPrefix(`${credentialProvider.getControlEndpoint()}`);
+  return withProtocolPrefix(`web.${credentialProvider.getControlEndpoint()}`);
 }
 
 export function getWebCacheEndpoint(
@@ -32,7 +32,7 @@ export function getWebCacheEndpoint(
   if (credentialProvider.isCacheEndpointOverridden()) {
     return withProtocolPrefix(credentialProvider.getCacheEndpoint());
   }
-  return withProtocolPrefix(`${credentialProvider.getCacheEndpoint()}`);
+  return withProtocolPrefix(`web.${credentialProvider.getCacheEndpoint()}`);
 }
 
 function withProtocolPrefix(endpoint: string): string {
--- a/packages/client-sdk-web/src/utils/web-client-utils.ts
+++ b/packages/client-sdk-web/src/utils/web-client-utils.ts
@@ -20,11 +20,24 @@ export function createCallMetadata(
 export function getWebControlEndpoint(
   credentialProvider: CredentialProvider
 ): string {
-  return `web.${credentialProvider.getControlEndpoint()}`;
+  if (credentialProvider.isControlEndpointOverridden()) {
+    return withProtocolPrefix(credentialProvider.getControlEndpoint());
+  }
+  return withProtocolPrefix(`${credentialProvider.getControlEndpoint()}`);
 }
 
 export function getWebCacheEndpoint(
   credentialProvider: CredentialProvider
 ): string {
-  return `web.${credentialProvider.getCacheEndpoint()}`;
+  if (credentialProvider.isCacheEndpointOverridden()) {
+    return withProtocolPrefix(credentialProvider.getCacheEndpoint());
+  }
+  return withProtocolPrefix(`${credentialProvider.getCacheEndpoint()}`);
+}
+
+function withProtocolPrefix(endpoint: string): string {
+  if (endpoint.match(/^http(?:s)?:\/\//)) {
+    return endpoint;
+  }
+  return `https://${endpoint}`;
 }
--- a/packages/client-sdk-web/test/integration/integration-setup.ts
+++ b/packages/client-sdk-web/test/integration/integration-setup.ts
@@ -1,6 +1,8 @@
 import {
   deleteCacheIfExists,
   testCacheName,
+  isLocalhostDevelopmentMode,
+  createCacheIfNotExists,
 } from '@gomomento/common-integration-tests';
 import {
   CreateCache,
@@ -18,13 +20,28 @@ import {
 import {ITopicClient} from '@gomomento/sdk-core/dist/src/clients/ITopicClient';
 import {ICacheClient} from '@gomomento/sdk-core/dist/src/clients/ICacheClient';
 
-const credsProvider = CredentialProvider.fromEnvironmentVariable({
-  environmentVariableName: 'TEST_AUTH_TOKEN',
-});
+let _credsProvider: CredentialProvider | undefined = undefined;
+
+function credsProvider(): CredentialProvider {
+  if (_credsProvider === undefined) {
+    if (isLocalhostDevelopmentMode()) {
+      _credsProvider = CredentialProvider.fromEnvironmentVariable({
+        environmentVariableName: 'TEST_AUTH_TOKEN',
+        controlEndpoint: 'https://no-controlplane-requests-allowed:9001',
+        cacheEndpoint: 'https://localhost:9001',
+      });
+    } else {
+      _credsProvider = CredentialProvider.fromEnvironmentVariable({
+        environmentVariableName: 'TEST_AUTH_TOKEN',
+      });
+    }
+  }
+  return _credsProvider;
+}
 
 export const IntegrationTestCacheClientProps: CacheClientProps = {
   configuration: Configurations.Laptop.latest(),
-  credentialProvider: credsProvider,
+  credentialProvider: credsProvider(),
   defaultTtlSeconds: 1111,
 };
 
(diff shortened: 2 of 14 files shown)


packages/client-sdk-web/src/utils/web-client-utils.ts
```diff
@@ -20,11 +20,24 @@ export function createCallMetadata(
...
 }
 
 export function getWebCacheEndpoint(
   credentialProvider: CredentialProvider
 ): string {
-  return `web.${credentialProvider.getCacheEndpoint()}`;
+  if (credentialProvider.isCacheEndpointOverridden()) {
+    return withProtocolPrefix(credentialProvider.getCacheEndpoint());
+  }
+  return withProtocolPrefix(`${credentialProvider.getCacheEndpoint()}`);
```
virratanasangpunth, line 35: Not sure if I see the differences between line 33 and line 35?

packages/client-sdk-web/src/utils/web-client-utils.ts
```diff
@@ -20,11 +20,24 @@ export function createCallMetadata(
...
+}
+
+function withProtocolPrefix(endpoint: string): string {
+  if (endpoint.match(/^http(?:s)?:\/\//)) {
```
malandis, line 39: Should we make this a case-insensitive regex match?

edit: I see where you inline the protocol later

packages/client-sdk-web/test/integration/integration-setup.ts
```diff
@@ -12,13 +14,28 @@ import {CacheClient, TopicClient, Configurations, AuthClient} from '../../src';
...
-});
+let _credsProvider: CredentialProvider | undefined = undefined;
+
+function credsProvider(): CredentialProvider {
+  if (_credsProvider === undefined) {
+    if (isLocalhostDevelopmentMode()) {
+      _credsProvider = CredentialProvider.fromEnvironmentVariable({
+        environmentVariableName: 'TEST_AUTH_TOKEN',
+        controlEndpoint: 'https://no-controlplane-requests-allowed:9001',
+        cacheEndpoint: 'https://localhost:9001',
```
virratanasangpunth, line 25: 1. I think this should be `http` and not `https`? I think you had to do some work around with tls termination otherwise?
2. port should be `9002` (for http) or `9112` (for https).
3. I got stuck big time with `localhost`. I think it has to be `0.0.0.0`.

scripts/make-localhost-token-from-dev-token.sh
```diff
@@ -0,0 +1,14 @@
...
+
+DEV_TOKEN=$1
+if [ "$DEV_TOKEN" == "" ]
+then
+  echo "Missing required argument: dev token"
+  echo "Usage: $0 <dev token>"
+  exit 1
+fi
+
+echo "{\"endpoint\":\"localhost\",\"api_key\":\"${DEV_TOKEN}\"}" |base64
```
virratanasangpunth, line 14: I don't think we need this script anymore?

virratanasangpunth (commented):
LGTM. A few comments.

packages/client-sdk-web/test/integration/integration-setup.ts
```diff
@@ -12,13 +14,28 @@ import {CacheClient, TopicClient, Configurations, AuthClient} from '../../src';
...
-  environmentVariableName: 'TEST_AUTH_TOKEN',
-});
+let _credsProvider: CredentialProvider | undefined = undefined;
+
+function credsProvider(): CredentialProvider {
+  if (_credsProvider === undefined) {
+    if (isLocalhostDevelopmentMode()) {
+      _credsProvider = CredentialProvider.fromEnvironmentVariable({
+        environmentVariableName: 'TEST_AUTH_TOKEN',
+        controlEndpoint: 'https://no-controlplane-requests-allowed:9001',
```
malandis, line 24: So [we're allowed to have one request](https://giphy.com/embed/l2Jea35pzsnoJUBDG) 😆

malandis (approved):

packages/common-integration-tests/src/common-int-test-utils.ts
```diff
@@ -10,15 +11,24 @@ import {
...
 import {v4} from 'uuid';
 
+export function isLocalhostDevelopmentMode(): boolean {
+  const useLocalhost = process.env.MOMENTO_SDK_TESTS_USE_LOCALHOST;
+  return useLocalhost !== undefined;
+}
 export function testCacheName(): string {
-  const name = process.env.TEST_CACHE_NAME || 'js-integration-test-default';
-  return name + v4();
+  return process.env.TEST_CACHE_NAME || `js-integration-test-default-${v4()}`;
```
malandis, line 19: This change is responsible for the integration test failure

scripts/make-localhost-token-from-dev-token.sh
cprice404: we definitely need it, the raw token you get from the server's console output is not in a format that the SDK knows what to do with. it expects it to be wrapped in base64 JSON. That's what the console does and that's what users supply to the SDK.

packages/client-sdk-web/src/utils/web-client-utils.ts
cprice404: good catch, i borked something here, will fix

packages/client-sdk-web/test/integration/integration-setup.ts
cprice404: 1. http did not work at all for me, and kenny said that it can't/won't. https works but requires some TLS setup that I documented in notion. if you are able to get it working with plaintext http that would simplify the instructions so i'd love to go that route, but i couldn't.
2. data plane port is 9001 on my box, which is pretty much a fresh clone of the repo. not sure how yours is 9002, but if it is, maybe we need to make this configurable. kenny told me 9112 is an admin port that is not relevant or useful for this task.
3. not sure how that could have happened. do you have anything weird in /etc/hosts? localhost works fine for me, and historically putting 0.0.0.0 in on the client side of stuff doesn't work for most tools in my experience :) but if 0.0.0.0 ends up working for people and localhost doesn't, we can change this

packages/common-integration-tests/src/common-int-test-utils.ts
cprice404: thx

cprice404:
@wandaitzuchen @virratanasangpunth - for tomorrow - CI is red on this due to the same set of test failures (against alpha). I can either override and merge or y'all can just check out this branch until we get the problem sorted out.

Just FYI that this will be blocking all PRs against this repo until tests can pass against alpha again though.

packages/client-sdk-web/test/integration/integration-setup.ts
```diff
@@ -43,10 +60,7 @@ export function SetupIntegrationTest(): {
     // Use a fresh client to avoid test interference with setup.
     const momento = momentoClientForTesting();
     await deleteCacheIfExists(momento, cacheName);
-    const createResponse = await momento.createCache(cacheName);
-    if (createResponse instanceof CreateCache.Error) {
-      throw createResponse.innerException();
-    }
+    await createCacheIfNotExists(momento, cacheName);
```
wandaitzuchen, line 63: Same on line 69-72.

packages/client-sdk-web/test/integration/integration-setup.ts
virratanasangpunth: 1. http okay i thought i had it worked in the past. nbd.
2. lol yes you're right it should be 9001, i had an override in my branch.
3. Not sure. If this works on your machine then let's keep it at localhost then.

scripts/make-localhost-token-from-dev-token.sh
virratanasangpunth: I have this code in my side branch and it seems to work just fine?
```
const credsProvider = new StringMomentoTokenProvider({
  authToken:
    'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzUxMiJ9.eyJzdWIiOiJsYXB0b3BAbW9tZW50b2hxLmNvbSIsImVtYWlsIjpudWxsLCJpc3MiOm51bGwsInZlciI6bnVsbCwiY3AiOiJzbyBpdCBsb29rcyBsaWtlIGEgbGVnYWN5IHRva2VuIiwiYXVkIjpudWxsfQ.gIMozwpfsh8JpCvLQwgfYhv2Qs7k8VnZrsKL233Ojs8sStfiCU71zxR290n2G9SuIF50tfLc0MR9e03HGCfM_A',
  cacheEndpoint: '0.0.0.0:9002',
  controlEndpoint: '0.0.0.0:9002',
 });
 ```

cprice404: okay, yeah, that works as a matter of happenstance since you override the endpoints, but it is definitely not a token format that the SDKs are intended to support.

malandis (approved):

PR #589, opened 2023-06-15
chore: update CONTRIBUTING to reflect new directory structure

cprice404:


CONTRIBUTING.md
```diff
@@ -10,44 +10,37 @@ Happy coding :dancer:
...
 
 ```
-npm run lint
+./scripts/build-all-packages.sh
 ```
 
-<br/>
-
-## Tests :zap:
+From that point you can change directories into any of the `package` subdirs to work on an individual package. Take a look at the `scripts` section of the `package.json` in each package directory to see what build commands are available. They all support `npm run build` to compile the code. Most also have `npm run unit-test` and `npm run integration-test` to run the unit and integration tests, respectively.
```
malandis, line 37: When running through these commands, I first hit `RESOURCE_EXHAUSTED` on the integration tests. I lowered the concurrency as explained below but it was still a surprise while following the steps here (and I already knew about this). Perhaps we should adjust the script command?

After lowering the concurrency, I the following test failed:
image

malandis (commented):
Thanks for the detailed instructions. I hit a snag when following along

malandis (approved):

PR #590, opened 2023-06-16
chore: migrate web sdk examples to use JSDom

cprice404:
Prior to this commit, we had been having trouble with some scenarios
when trying to use the Web SDK in example node.js programs. This is
because there are a few things that the grpc-web library depends on
that are generally only available in browsers, not in node.js.

We had added a dep on the `xhr` library to give us an implementation
of XmlHttpRequest, which allowed most of the basic examples to work.
However, some cases were still not working, including topic subscriptions
(which uses grpc streaming rather than only unary requests).

After digging into this I observed that the jest setup for the web SDK
tests was using a library called JSDom to set up a browser-compatible
environment, and the topic subscriptions were working fine there.

In this commit I have added an `initJSDom` function to a file in the
utils dir, which allows us to do a similar thing to what the jest
tests were doing to set up a JSDom environment to run the code in.
After this change, and removing the xhr dependency, I've verified
that the topic subscriptions now work, as well as all of the other
examples.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -114,6 +114,42 @@ jobs:
             npm ci
             npm run build
           popd
+
+      - name: Send CI failure mail
+        if: ${{ steps.validation.outcome == 'failure' }}
+        uses: momentohq/standards-and-practices/github-actions/error-email-action@gh-actions-v2
+        with:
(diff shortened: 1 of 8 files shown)


pgautier404 (dismissed):
This is great, thanks! So much cleaner.

rishtigupta (approved):

PR #598, opened 2023-06-24
feat: Add fine grained cache and topic level permissions to tokens

krispraws:
### Summary
- This PR adds the ability to scope tokens to certain topics or caches per the scenarios in this [doc](https://www.notion.so/momentohq/Authorization-Scenarios-f1cea5c1caee4523b3ebe85687c65a55).
- See related PR for proto changes: https://github.com/momentohq/client-protos/pull/179
- **There are breaking changes in `packages/core/src/auth/tokens/token-scope.ts`**
    - `Permission`, `CachePermission` and `TopicPermission` have been changed from `class` to `interface`
    - `None` has been removed from the `CacheRole` and `TopicRole` enums
    - `TopicRole.ReadWrite` has been renamed to `TopicRole.PublishSubscribe`
- New example for scoped tokens has been added in `examples/nodejs/access-control/access-control.ts`
However it has not been added to dependabot and workflows/build.yml because the CI jobs for testing examples do not build the sdk so we get errors like below. The new example will be added to workflows after the SDK is released.
```
/Users/runner/work/client-sdk-javascript/client-sdk-javascript/examples/nodejs/access-control/access-control.ts
Error:    18:8    error    Unable to resolve path to module '@gomomento/sdk'                import/no-unresolved
Error:    18:8    error    "@gomomento/sdk" is not found                                    node/no-missing-import
```

### Testing
- Added unit tests and integration tests


malandis:
Looks good. Will approve once build failures are fixed.

packages/core/src/auth/tokens/token-scope.ts
```diff
@@ -1,26 +1,66 @@
 export enum CacheRole {
   None = 'none',
   ReadWrite = 'readwrite',
+  ReadOnly = 'readonly',
 }
 
+export class Any {}
```
krispraws, line 7: Specifically looking for approvals on names and shapes in this file today.

malandis: My apologies. I thought this was the client-protos PR!

packages/client-sdk-nodejs/src/internal/internal-auth-client.ts
```diff
@@ -169,39 +184,152 @@ function tokenPermissionToGrpcPermission(
...
+
+  const cachePermissionMap: ResourceToPermissionMap = {};
+
+  function groupByCache(permission: CachePermission) {
+    const key = JSON.stringify(permission.cache);
+    const entry = cachePermissionMap[key];
+    if (entry instanceof CachePermission) {
+      // Cannot add multiple permissions for the same resource.
+      throw new InvalidArgumentError(
+        `Cache ${key} cannot have multiple permissions [${permission.cacheRole}, ${entry.cacheRole}]`
```
cprice404, line 302: do we want to do this validation in the client? i almost think this would be better to just do on the server so that the errors are consistent across the different SDKs and we don't have to repeat the code so much?  just a thought.

packages/core/src/auth/tokens/token-scope.ts
```diff
@@ -1,26 +1,66 @@
 export enum CacheRole {
   None = 'none',
   ReadWrite = 'readwrite',
+  ReadOnly = 'readonly',
 }
 
+export class All {}
```
cprice404, line 7: I don't feel strongly about this but if you liked `Any` better, you could call this `AnyCache`.

packages/core/src/auth/tokens/token-scope.ts
```diff
@@ -1,26 +1,66 @@
...
+export class CacheName {
+  name: string;
+  constructor(name: string) {
+    this.name = name;
+  }
+}
+export type CacheResource = All | CacheName;
```
cprice404, line 14: 👍 to how you are doing the typing here where you are unioning the things together.

As for the name "Resource" here, I don't hate it, might have a slight pref for something like "Selector" or "Descriptor", but I don't love those either, and would be fine being outvoted

packages/core/src/auth/tokens/token-scope.ts
```diff
@@ -1,26 +1,66 @@
...
+
 export class CachePermission {
   cacheRole: CacheRole;
+  cache: CacheResource;
 
-  constructor(cacheRole: CacheRole) {
+  constructor(cacheRole: CacheRole, cache?: CacheResource) {
     this.cacheRole = cacheRole;
+    if (cache) {
```
cprice404, line 22: nit, I think you could remove the `?` from the argument type and just put a default value directly in the signature?

packages/core/src/auth/tokens/token-scope.ts
```diff
@@ -1,26 +1,66 @@
...
 export class TopicPermission {
   topicRole: TopicRole;
-
-  constructor(topicRole: TopicRole) {
+  cache: CacheResource;
+  topic: TopicResource;
+  constructor(
+    topicRole: TopicRole,
+    cache?: CacheResource,
+    topic?: TopicResource
```
cprice404, line 51: when you start having multiple optional positional args in JS it is common to create an interface to represent them. e.g.:

```
export interface TopicPermissionOpts {
   cache?: CacheResource,
   topic?: TopicResource
}
```

Then users can call it like this:

```
new TopicPermission(role, {topic: ...});
```

Otherwise it will feel weird that you are allowing `cache` to be optional but basically forcing them to pass `undefined` explicitly if they don't want to set it.

packages/core/src/auth/tokens/token-scope.ts
cprice404: if you end up making an Options interface for the Topic one as per my comment below, you will want to do it here too for consistency. In which case you won't be able to set a default value in the function signature and you can ignore my comment about that above :)

cprice404 (commented):
handful of small things, food for thought, definitely open to discussion

packages/client-sdk-nodejs/src/internal/internal-auth-client.ts
krispraws: I was going to add the validation in both client and server, but I can remove it from the client if we have been following that pattern for other apis.

packages/core/src/auth/tokens/token-scope.ts
krispraws: Allen pointed out Any has special meaning in JS. I am fine with All - this way I don't have to create two messages `AnyCache` and `AnyTopic`

packages/core/src/auth/tokens/token-scope.ts
krispraws: I renamed Resource to Selector in the client-protos and am working on doing that here as well!

packages/core/src/auth/tokens/token-scope.ts
krispraws: Is one preferred over the other w.r.t. default values vs interface? I don't have strong opinions on Javascript so I can follow whatever the standards are.

cprice404: i would go with options interface, it is definitely the most common pattern when you have multiple optional arguments like you do in the topic one.

packages/core/src/auth/tokens/token-scope.ts
cprice404: that's fine with me but there will probably be some language where `All` is a keyword :P

we can just give it a different name in that SDK though if/when he hit that :)

packages/client-sdk-nodejs/src/internal/internal-auth-client.ts
cprice404: i wouldn't say it's a pattern, more of a case-by-case thing. but the more complex the validation is, the more concerned i am about the implementation drifting in one SDK vs the next vs the server. So I honestly might err towards just server-side on this one. but it's not a blocker for me

krispraws: Makes sense. Validation of this complexity is better off in the server.

cprice404:
about to review this again but just curious what happens if this gets merged before the server-side changes?

packages/client-sdk-nodejs/test/unit/integration/auth-client.test.ts
```diff
@@ -1,5 +1,15 @@
 import {InternalSuperUserPermissions} from '@gomomento/sdk-core/dist/src/internal/utils';
-import {AllDataReadWrite} from '@gomomento/sdk-core';
```
cprice404, line 2: I apologize for this comment, I realize it's not something that was introduced in this PR and that I must have just missed in an earlier PR, my bad.

Two observations about this file:

1. The path says `unit/integration`. it should be one or the other. It looks like this is just a unit test? if so can we move it up a directory?
2. It's pretty unusual for us to be writing tests that directly reference the generated GRPC classes. Is there a way that we could get the same test coverage that are after here using only our public API classes? Such a test would be far more valuable in my eyes.

I won't block merging the PR on those points but I'd really like to address #2 in a fast follow if possible, if we don't do it in this PR. I don't want people to start copy/pasting the pattern of writing tests that reference the generated classes unless it's truly necessary :)

But please tell me if I'm wrong and these tests have high value and cannot be written using only public API classes.

packages/core/src/auth/tokens/token-scope.ts
```diff
@@ -1,26 +1,68 @@
...
+}
+
+export class All {}
+export class CacheName {
+  name: string;
+  constructor(name: string) {
+    this.name = name;
+  }
+}
+export type CacheSelector = All | CacheName;
```
cprice404, line 14: what would you think about supporting `| string` here and then having the code implicitly convert that to a `CacheName`?

packages/core/src/auth/tokens/token-scope.ts
```diff
@@ -1,26 +1,68 @@
...
+  ReadOnly = 'readonly',
+}
+
+export class TopicName {
+  name: string;
+  constructor(name: string) {
+    this.name = name;
+  }
+}
+export type TopicSelector = All | TopicName;
```
cprice404, line 45: same comment here about `| string`

packages/core/src/auth/tokens/token-scope.ts
```diff
@@ -35,8 +77,11 @@ export class Permissions {
 }
 
 export const AllDataReadWrite: Permissions = new Permissions([
-  new CachePermission(CacheRole.ReadWrite),
-  new TopicPermission(TopicRole.ReadWrite),
+  new CachePermission(CacheRole.ReadWrite, {cache: new All()}),
```
cprice404, line 80: not that I mind this being explicit but technically you shouldn't need to do this, right? Isn't it the default?

packages/core/src/auth/tokens/token-scope.ts
```diff
@@ -35,8 +77,11 @@ export class Permissions {
...
+  new TopicPermission(TopicRole.ReadWrite, {
+    cache: new All(),
+    topic: new All(),
+  }),
```
cprice404, line 84: can we add some test coverage of this stuff in the common-integration-tests code? Besides the value of the test coverage it would be helpful in terms of seeing some examples of what user code that consumes this will look like.

krispraws: Absolutely - I am working on those, and will add them to this PR as another commit

packages/core/src/auth/tokens/token-scope.ts
krispraws: yes, it is the default so not needed technically.

packages/client-sdk-nodejs/test/unit/integration/auth-client.test.ts
krispraws: 1. I agree - it confused me too when I saw the tests. I will move it up.
2. These are essentially unit tests for testing the conversion from SDK types to the client proto types so they have to use the generated GRPC classes. The integration tests will be the high value tests since those will implicitly test the conversions as well. If there was an api that would describe an existing token, I could have written a test that created a token and described it back and verified the permissions were the same.

krispraws:
> about to review this again but just curious what happens if this gets merged before the server-side changes?



> about to review this again but just curious what happens if this gets merged before the server-side changes?

Things would break if anyone tried to generate a AllDataReadWrite token because it will now try to create the token with the new permission fields. But I don't plan to merge this before server side changes are deployed.

packages/core/src/auth/tokens/token-scope.ts
krispraws: I can do that. Curious why that is better? Is it so that you can do something like CacheSelector('foo')?

cprice404: I'm kind of just guessing because I don't think we have a place in the PR where you show examples of what user code will look like, but I think it will allow something like:

`new CachePermission(ReadOnly, 'my-awesome-cache')` instead of `new CachePermission(ReadOnly, new CacheName('my-awesome-cache')`.

and actually now that I think about it I think maybe CacheName should be an interface instead of a class. Maybe we can zoom about this for a second tomorrow morning?

packages/core/src/auth/tokens/token-scope.ts
cprice404: excellent, thanks

cprice404:
> > about to review this again but just curious what happens if this gets merged before the server-side changes?
> 
> > about to review this again but just curious what happens if this gets merged before the server-side changes?
> 
> Things would break if anyone tried to generate a AllDataReadWrite token because it will now try to create the token with the new permission fields. But I don't plan to merge this before server side changes are deployed.

ok cool, in that case probably safest to convert this to a "draft" PR until then.

packages/client-sdk-nodejs/test/unit/integration/auth-client.test.ts
cprice404: okay i will defer to your judgment on #2.

packages/core/src/auth/tokens/token-scope.ts
krispraws: Sure, will ping you tomorrow morning

krispraws: Here are a couple of integration tests that show user code: https://github.com/momentohq/client-sdk-javascript/pull/598/files#diff-ae2e3efaec41bcc9a06f77ab73967c608830d8c711127a11f30253ca462a3fe3
I was waiting to test these with the server before I write more.

cprice404:
@krispraws based on our call the other day and the thing you showed me where you were able to pass in an object literal to the function, I think you and I are on the same page, but please feel free to re-request review whenever you are ready for another pass.

packages/client-sdk-nodejs/test/unit/integration/auth-client.test.ts
krispraws: Moved the unit test out of the `unit/integration` folder.

packages/core/src/auth/tokens/token-scope.ts
krispraws: Removed the explicit arguments for All and also the redundant unit test I had added, since the unit test for `AllDataReadWrite` will also test we are using the right defaults.

packages/core/src/auth/tokens/token-scope.ts
krispraws: Integration tests added.

krispraws:
> > > about to review this again but just curious what happens if this gets merged before the server-side changes?
> > 
> > 
> > > about to review this again but just curious what happens if this gets merged before the server-side changes?
> > 
> > 
> > Things would break if anyone tried to generate a AllDataReadWrite token because it will now try to create the token with the new permission fields. But I don't plan to merge this before server side changes are deployed.
> 
> ok cool, in that case probably safest to convert this to a "draft" PR until then.

I was actually wrong about things breaking. Control plane would just have ignored the new fields in the proto, so nothing would break.

packages/core/src/auth/tokens/token-scope.ts
krispraws: Changed CacheName and TopicName to interface instead of class. Also added string as a member of the CacheSelector and TopicSelector unions.

packages/core/src/auth/tokens/token-scope.ts
krispraws: done

packages/common-integration-tests/src/auth-client.ts
```diff
@@ -336,6 +351,207 @@ export function runAuthClientTests(
...
+          'it is a nice day today!'
+        )
+      ).toBeInstanceOf(TopicPublish.Success);
+      expect(
+        await sessionTokenTopicClient.publish(
+          FGA_CACHE_1,
+          FGA_TOPIC_2,
+          'This weird trick will make you a better developer...'
+        )
+      ).toBeInstanceOf(TopicPublish.Success);
```
cprice404, line 401: idk if these 2 publish calls really do anything for the test?

packages/common-integration-tests/src/auth-client.ts
```diff
@@ -336,6 +351,207 @@ export function runAuthClientTests(
...
+        trivialHandlers
+      );
+      expect(subResp).toBeInstanceOf(TopicSubscribe.Error);
+      expect(subResp).toHaveProperty(
+        'errorCode',
+        MomentoErrorCode.PERMISSION_ERROR
+      );
+    });
+
+    it('can read/write cache foo and all topics in cache bar', async () => {
```
cprice404, line 471: nit: reading these tests I'm having a little trouble mapping "foo" and "bar" to FGA_CACHE_* without scrolling back and forth

packages/common-integration-tests/src/auth-client.ts
```diff
@@ -336,6 +351,207 @@ export function runAuthClientTests(
...
+
+      // Read/Write on cache `foo` is allowed
+      const setResp = await cacheClient.set(FGA_CACHE_1, 'ned', 'flanders');
+      expect(setResp).toBeInstanceOf(CacheSet.Success);
+
+      const hitResp = await cacheClient.get(FGA_CACHE_1, 'ned');
+      expect(hitResp).toBeInstanceOf(CacheGet.Hit);
+      expect(hitResp).toHaveProperty('valueString', 'flanders');
+
+      // Read/Write on cache `bar` is not allowed
```
cprice404, line 495: same nit re: foo/bar in these comment strings

packages/common-integration-tests/src/auth-client.ts
```diff
@@ -336,6 +351,207 @@ export function runAuthClientTests(
...
+
+    afterAll(async () => {
+      expect(
+        await sessionTokenCacheClient.deleteCache(FGA_CACHE_1)
+      ).toBeInstanceOf(DeleteCache.Success);
+      expect(
+        await sessionTokenCacheClient.deleteCache(FGA_CACHE_2)
+      ).toBeInstanceOf(DeleteCache.Success);
+    });
+  });
```
cprice404, line 554: would love to see one more test that did ReadOnly on topics

packages/core/src/auth/tokens/token-scope.ts
```diff
@@ -5,13 +5,13 @@ export enum CacheRole {
 }
 
 export class All {}
```
cprice404, line 7: now that I see that the class definition for `All` is empty, I think you can just do `export const` instead of `export class`, and then you won't need the `new All` stuff. You will have to choose a type for it that can be disambiguated from `String` though. could be an enum or an integer. or you can still use a class and just make the class private, and have `All` be a constant instance of the class.

cprice404 (commented):
looking really good! can't wait to play with it.

Just a few nits/suggestions/questions

packages/common-integration-tests/src/auth-client.ts
krispraws: They are just creating existing topics on a cache. One of the tests tries to publish to both existing and new topics and ensures both fail when you don't have topic permissions.

cprice404: topics aren't a resource that actually exists or has state on the server though, so I don't think these lines are needed.

packages/client-sdk-nodejs/test/unit/auth-client.test.ts
```diff
@@ -154,84 +152,30 @@ describe('internal auth client', () => {
       grpcPermissions.explicit = explicitPermissions;
       const cacheAndTopicPermissions: Permissions = new Permissions([
         new CachePermission(CacheRole.ReadOnly, {cache: new All()}),
-        new CachePermission(CacheRole.ReadWrite, {cache: new CacheName('foo')}),
+        new CachePermission(CacheRole.ReadWrite, {
+          cache: {name: 'foo'},
+        }),
         new TopicPermission(TopicRole.ReadOnly, {
           cache: new All(),
```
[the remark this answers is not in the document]
krispraws: Done in https://github.com/momentohq/client-sdk-javascript/pull/598/commits/f2aad35500cbec9f6c42ca858e70abafffacbf58

packages/common-integration-tests/src/auth-client.ts
krispraws: Removed in https://github.com/momentohq/client-sdk-javascript/pull/598/commits/f2aad35500cbec9f6c42ca858e70abafffacbf58

packages/common-integration-tests/src/auth-client.ts
krispraws: Fixed in https://github.com/momentohq/client-sdk-javascript/pull/598/commits/f2aad35500cbec9f6c42ca858e70abafffacbf58

packages/common-integration-tests/src/auth-client.ts
krispraws: Added in https://github.com/momentohq/client-sdk-javascript/pull/598/commits/f2aad35500cbec9f6c42ca858e70abafffacbf58

packages/core/src/auth/tokens/token-scope.ts
krispraws: Done in https://github.com/momentohq/client-sdk-javascript/pull/598/commits/f2aad35500cbec9f6c42ca858e70abafffacbf58

cprice404 (dismissed):
I'm good with this PR at this point; my only note:

I don't want to do a release of the JS SDK until we have spent some time bug-bashing this with an eye on the API, just in case we come up with any other usability changes.

and it would be great if we can add an example in the examples dir, and possibly add web sdk support, prior to release.

we don't have any urgent things that I know of that need to get out in the JS SDK so I think it's okay to merge and block release for a few days or whatever, but we just need to make sure everyone's aware of that plan once we merge this.

cprice404 (approved):
same caveat I mentioned in my previous approval about not wanting to do a release of the SDK with this just yet, but I'm fine with merging

krispraws:
> I'm good with this PR at this point; my only note:
> 
> I don't want to do a release of the JS SDK until we have spent some time bug-bashing this with an eye on the API, just in case we come up with any other usability changes.
> 
> and it would be great if we can add an example in the examples dir, and possibly add web sdk support, prior to release.
> 
> we don't have any urgent things that I know of that need to get out in the JS SDK so I think it's okay to merge and block release for a few days or whatever, but we just need to make sure everyone's aware of that plan once we merge this.

Added an example in https://github.com/momentohq/client-sdk-javascript/pull/598/commits/8480434030e88fee8d8fb67964104614b471131d 
It is very basic and works on my machine.
Re: web-sdk support, the work is likely minimal but I would need to scope it. @danielamiao - do we want to add web sdk support right now?

danielamiao:
@krispraws @cprice404 let's not make web sdk support a release blocker, but a fast follow post release.

examples/nodejs/access-control/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "eslint": "8.19.0",
+    "eslint-config-prettier": "8.5.0",
+    "eslint-plugin-import": "2.26.0",
+    "eslint-plugin-node": "11.1.0",
+    "eslint-plugin-prettier": "4.2.1",
+    "prettier": "2.7.1",
+    "typescript": "4.4.3"
+  },
+  "dependencies": {
+    "@gomomento/sdk": "file:../../../packages/client-sdk-nodejs"
```
cprice404, line 28: this is the right thing to do for this PR but we should come back and change it to a release version as soon as we have one.

We should also add this new example dir to CI, in the .github actions for PRs and for dependabot.

Those do not need to block this PR but we should do them when we can.

cprice404 (dismissed):

examples/nodejs/access-control/package.json
krispraws: Do we have a runbook for everything that needs to be done for releasing new SDKs? It would be really useful.

cprice404: this isn't really about new SDKs or releases, it's about new examples. we should add something in CONTRIBUTING.md.

cprice404: https://github.com/momentohq/client-sdk-javascript/pull/620

examples/nodejs/access-control/README.md
```diff
@@ -0,0 +1,40 @@
...
+# Node.js Client SDK
+
+_Read this in other languages_: Coming Soon
+
+<br>
+
+## Example Requirements
+
+- Node version 14 or higher is required
+- A Momento Auth Token is required, you can generate one using the [Momento CLI](https://github.com/momentohq/momento-cli)
```
poppoerika, line 18: I thought we want to encourage users to use our console to generate a token?

krispraws: Good catch, I copied this from the examples/nodejs/cache/README.md

krispraws:
> > I'm good with this PR at this point; my only note:
> > I don't want to do a release of the JS SDK until we have spent some time bug-bashing this with an eye on the API, just in case we come up with any other usability changes.
> > and it would be great if we can add an example in the examples dir, and possibly add web sdk support, prior to release.
> > we don't have any urgent things that I know of that need to get out in the JS SDK so I think it's okay to merge and block release for a few days or whatever, but we just need to make sure everyone's aware of that plan once we merge this.
> 
> Added an example in [8480434](https://github.com/momentohq/client-sdk-javascript/commit/8480434030e88fee8d8fb67964104614b471131d) It is very basic and works on my machine. Re: web-sdk support, the work is likely minimal but I would need to scope it. @danielamiao - do we want to add web sdk support right now?

web-sdk changes are now part of this PR - thanks to @rishtigupta !

examples/nodejs/access-control/README.md
NoSQLKnowHow, line 18:
```suggestion
- A Momento auth token is required, you can generate one using the [Momento CLI](https://github.com/momentohq/momento-cli)
```

No upper case for "auth token" as it is not a product name.

packages/core/src/auth/tokens/token-scope.ts
```diff
@@ -23,20 +23,25 @@ export interface CachePermissionOptions {
...
+//     this.cacheRole = cacheRole;
+//     this.cache = options?.cache ?? AllCaches;
+//   }
+// }
+
 export enum TopicRole {
   None = 'none',
-  ReadWrite = 'readwrite',
-  ReadOnly = 'readonly',
+  PublishSubscribe = 'readwrite',
```
krispraws, line 43: Do we want to change the enum values as well for consistency?

examples/nodejs/access-control/README.md
krispraws: fixed in https://github.com/momentohq/client-sdk-javascript/pull/598/commits/d583b64ee0db14557dc380df1c86785c67dbd326

krispraws: fixed in https://github.com/momentohq/client-sdk-javascript/pull/598/commits/d583b64ee0db14557dc380df1c86785c67dbd326

packages/core/src/auth/tokens/token-scope.ts
malandis: +1 from me --a developer may print a enum value be surprised by the inconsistency.

malandis (commented):
Looks good! Just one suggestion and I agree we should change the enum values to be consistent.

packages/client-sdk-nodejs/test/unit/cache-client.test.ts
```diff
@@ -7,13 +7,9 @@ import {
...
 import {SimpleCacheClientProps} from '../../src/cache-client-props';
-
-// This auth token is syntactically correct but not actually valid; it won't work with the real Momento Servers.
-// Used only for unit testing the constructors etc.
-const fakeAuthTokenForTesting =
-  'eyJhbGciOiJIUzUxMiJ9.eyJzdWIiOiJzcXVpcnJlbCIsImNwIjoiY29udHJvbCBwbGFuZSBlbmRwb2ludCIsImMiOiJkYXRhIHBsYW5lIGVuZHBvaW50In0.zsTsEXFawetTCZI';
 const credentialProvider = new StringMomentoTokenProvider({
-  authToken: fakeAuthTokenForTesting,
+  authToken:
+    'eyJhbGciOiJIUzUxMiJ9.eyJzdWIiOiJzcXVpcnJlbCIsImNwIjoiY29udHJvbCBwbGFuZSBlbmRwb2ludCIsImMiOiJkYXRhIHBsYW5lIGVuZHBvaW50In0.zsTsEXFawetTCZI',
```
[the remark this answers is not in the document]
krispraws: This change snuck in because of a merge from main followed by revert. I have merged the changes from main again.
Fixed in https://github.com/momentohq/client-sdk-javascript/pull/598/commits/bd1add565f4a5c62a79606fdc11edc43904bb02e

packages/core/src/auth/tokens/token-scope.ts
krispraws: Fixed in https://github.com/momentohq/client-sdk-javascript/pull/598/commits/649148592f9a49d01715fa7258ff2bee2c8e741d

malandis (approved):
🎉

rishtigupta (approved):

malandis (dismissed):

packages/client-sdk-web/src/index.ts
```diff
@@ -95,6 +95,12 @@ import {
   TokenScope,
   ExpiresIn,
   ExpiresAt,
+  CacheName,
```
bruuuuuuuce, line 98: do we also need to export `TokenScopes` here too?

bruuuuuuuce (commented):
Looks great! One comment about exports for the web client, but otherwise LGTM

packages/client-sdk-web/src/index.ts
krispraws: Good catch! fixed in https://github.com/momentohq/client-sdk-javascript/pull/598/commits/37fef37d3c8a3a47ee92595a9d3070de4bbf8d1c

bruuuuuuuce (approved):
:shipit:

PR #601, opened 2023-06-27
fix: allow for grpc web pubsub client to reconnect streams on network…

bruuuuuuuce:
… errors

Inside of a browser, we were seeing that our topic client was not attempting to resubscribe during/after a brief network outage. This pr addresses this by making sure that we are checking for the correct grpc error message from grpc web, and makes sure that we correctly set the `firstMessage` option to be true _only_ when the user is subscribing to a new stream, and not on reconnection attempts

### DIFF
--- a/packages/client-sdk-web/src/internal/pubsub-client.ts
+++ b/packages/client-sdk-web/src/internal/pubsub-client.ts
@@ -6,15 +6,15 @@ import {
   TopicItem,
   UnknownError,
 } from '@gomomento/sdk-core';
-import {Request, UnaryResponse, RpcError, StatusCode} from 'grpc-web';
+import {Request, RpcError, StatusCode, UnaryResponse} from 'grpc-web';
 import {TopicClientProps} from '../topic-client-props';
 import {truncateString} from '@gomomento/sdk-core/dist/src/internal/utils';
 import {TopicPublish, TopicSubscribe} from '../index';
 import {cacheServiceErrorMapper} from '../errors/cache-service-error-mapper';
 import {
   AbstractPubsubClient,
-  SendSubscribeOptions,
   PrepareSubscribeCallbackOptions,
+  SendSubscribeOptions,
 } from '@gomomento/sdk-core/dist/src/internal/clients/pubsub/AbstractPubsubClient';
 import {
   convertToB64String,
@@ -36,6 +36,8 @@ export class PubsubClient<
   protected readonly logger: MomentoLogger;
   private readonly clientMetadataProvider: ClientMetadataProvider;
 
+  private static readonly RST_STREAM_NO_ERROR_MESSAGE =
+    'Received RST_STREAM with code 0';
   private static readonly BROWSER_DISCONNECT =
     'Http response at 400 or 500 level, http status code: 0';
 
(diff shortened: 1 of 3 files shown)


packages/client-sdk-web/src/internal/pubsub-client.ts
```diff
@@ -36,8 +36,8 @@ export class PubsubClient<
   protected readonly logger: MomentoLogger;
   private readonly clientMetadataProvider: ClientMetadataProvider;
 
-  private static readonly RST_STREAM_NO_ERROR_MESSAGE =
-    'Received RST_STREAM with code 0';
```
malandis, line 40: Does the deleted message not happen in the web client?

bruuuuuuuce: hmm good point, I didn't see this in my testing, but you are right I think this could still happen if there is a deployment going on. I will update it to retry on both of these errors

packages/client-sdk-web/src/internal/pubsub-client.ts
```diff
@@ -144,7 +146,11 @@ export class PubsubClient<
       const prepareCallbackOptions: PrepareSubscribeCallbackOptions = {
         ...options,
         restartedDueToError: false,
-        firstMessage: true,
+        // We call subscribe send on retryable errors. All retry-able errors happen
+        // after the first message. Here we can check if there has been a message that
+        // has already been sent from this stream by checking to see if there is a resumable
+        // sequence number. If there is one, we know that this isn't the first message.
+        firstMessage: !options.subscriptionState.resumeAtTopicSequenceNumber,
```
cprice404, line 153: i thought you decided that this was a red herring? I don't understand how this comes into play based on my understanding of what the bug was.

bruuuuuuuce: this is required because our error callback keys off of this `firstMessage` boolean. Normally if the first message is an error we want to end the stream. But if the stream was previously connected, and subsequently severed, we want to make attempts to reconnect the stream. If the first message is always set to true, and we keep getting errors from `subscribeSend` trying to reconnect, the sdk will terminate the connection and stop trying to reconnect. An alternative approach would be to check for `isRstStream` errors first before checking for first message

https://github.com/momentohq/client-sdk-javascript/blob/b42f50310415c8817d04ebc40045efd35b528f6a/packages/core/src/internal/clients/pubsub/AbstractPubsubClient.ts#L171-L207

malandis: Thanks for the explanation. Better to make the `firstMessage` semantics explicit like you did. The only issue I can think of is we are now coupling the first message detection to the message sequence number. As long as we don't change the latter, we will be fine.

Can you submit an issue for us to come back to this to explicitly model "first message in the user-facing subscription" vs first message in the stream?

malandis (approved):

packages/client-sdk-web/src/internal/pubsub-client.ts
cprice404: if we did what Matt said above and "check for isRstStream" first, then we wouldn't need two state variables for "firstMessage" like you are hinting at, right Michael?

feels like a code smell here that we are maybe overloading the semantics of `firstMessage` to mean two different things, and if we can avoid that by simply re-ordering the `if` blocks, that sounds better to me?

malandis: That works too.

cprice404: @bruuuuuuuce - I know that @danielamiao wants to get this merged and released so that we have it in our pocket in case the customer needs it. So if it would take too much time for you to test the alternate fix of re-ordering the `if` blocks, feel free to merge this and do a release and then just fast-follow tomorrow morning with the update?

bruuuuuuuce: I'm going to release it as is and fast follow tomorrow with the swapping the if statements fix, my main concern being that changing the ordering of those will affect both the nodejs sdk and the web sdk. So will just require extra testing

Issue #113, opened 2022-11-28
client times out intermittently, possibly related to storing in a lambda global variable


cprice404:
We have had a user report that when using our node.js client in a lambda environment (specifically, via [SST](https://sst.dev/)), they are observing intermittent timeout errors, even when they have configured the timeout to be an excessively large value (e.g. 50 seconds).  

Need to try to create a reproducer environment for this and fix it.

cprice404:
One theory that we have at the moment is that this could be related to the client being stored as a global variable in the lambda code.  In such a scenario it is possible that after a certain period of inactivity the grpc channel might be closed even though the client object is still retained in the global variable.  Then when we try to use the client to make a request, we time out due to the closed channel.  This would be a good scenario to try to set up a reproducer for.

If it does reproduce, we need to try to figure out why (a) the closed connection is not being detected, and (b) if there is some keepalive setting that we should be using to prevent this from happening.

cprice404:
I've been able to create a reproducer for this.

Branch can be found here: https://github.com/cprice404/client-sdk-javascript/tree/sst-sandbox

Most relevant bit of the code is here: https://github.com/cprice404/client-sdk-javascript/blob/sst-sandbox/examples/sst/my-sst-app/services/functions/lambda.ts#L27

That branch vendors our SDK code into the lambda subdir because that was the easiest way to allow us to test changes to the SDK.

Through trial and error I've been able to determine that there is a 15-minute timeout somewhere.  If you run the sst app locally and invoke the lambda a few times, it will succeed.  If you then allow it to sit idle for 15 minutes, and then invoke it again, you'll get the timeout error.

This is also reproducible in the AWS lambda environment.  It's a little harder to do there because you have to ensure that the lambda container survives in an idle state for 15-minutes, but I can usually reproduce it if I set up provisioned capacity for the lambda.  I have a log message that shows whether the global variable is being re-initialized or not, so it's easy to see if you are hitting the same container from a previous run.  In all cases where I've let it sit idle for 15 minutes *and* observed the absence of that log message, it times out.

I've googled a bit and it seems like the solution is going to be to reduce the client keepalive setting to < 15 minutes, but before I do that I would like to see if we can enable some more verbose logging - especially for the gRPC layer.  It would be nice to see if we can determine whether the client or the server is the one severing this connection.  If it's the server, it would be nice to know if we have any metrics that would provide visibility for this.

cprice404:
[...]

PR #605, opened 2023-06-29
feat: remove idle grpc wrapper client

rishtigupta:
## PR Description:
Remove idle grpc wrapper client

## Testing
Integration tests : `npm run build-and-run-tests`

## Issue 
closes https://github.com/momentohq/dev-eco-issue-tracker/issues/380

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/pubsub-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/pubsub-client.ts
@@ -7,8 +7,6 @@ import {cacheServiceErrorMapper} from '../errors/cache-service-error-mapper';
 import {ChannelCredentials, Interceptor, ServiceError} from '@grpc/grpc-js';
 import {Status} from '@grpc/grpc-js/build/src/constants';
 import {version} from '../../package.json';
-import {IdleGrpcClientWrapper} from './grpc/idle-grpc-client-wrapper';
-import {GrpcClientWrapper} from './grpc/grpc-client-wrapper';
 import {TopicClientProps} from '../topic-client-props';
 import {middlewaresInterceptor} from './grpc/middlewares-interceptor';
 import {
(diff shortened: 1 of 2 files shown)


packages/client-sdk-nodejs/src/internal/pubsub-client.ts
```diff
@@ -154,7 +145,10 @@ export class PubsubClient extends AbstractPubsubClient {
         options.subscriptionState.resumeAtTopicSequenceNumber,
     });
 
-    const call = this.clientWrapper.getClient().Subscribe(request, {
+    // const call = this.clientWrapper.getClient().Subscribe(request, {
+    //   interceptors: this.streamingInterceptors,
+    // });
```
cprice404, line 150: mean to leave this in?

rishtigupta: nope, will get rid of it

malandis:
OOTL on the idle client wrapper -- what was the purpose of the idle wrapper and why are we removing it?

cprice404:
> OOTL on the idle client wrapper -- what was the purpose of the idle wrapper and why are we removing it?

I went over this in sprint planning the other day, but to recap:

The idle wrapper was introduced to resolve issues using the CacheClient in lambdas; see: https://github.com/momentohq/client-sdk-javascript/issues/113#issuecomment-1334212083

It doesn't really make a lot of sense for us to have it in the TopicClient, because TopicClient subscriptions won't be a common use case for lambda. I think the only reason it exists on the TopicClient is because we copypasta'd it from the CacheClient when we were initially creating the TopicClient.

In real-world TopicClient use cases, having the idle wrapper in place means that if someone is participating in a chat and there are no messages for 4 minutes, we forcefully close and reconnect the stream for no good reason. Folks who were testing in browsers noticed this behavior and inquired about it, and after discussion we decided it didn't make sense for the TopicClient.

cprice404 (approved):

PR #617, opened 2023-07-06
fix: remove TEST_CACHE_NAME from release job

cprice404:
The tests don't need a hard-coded cache name and they don't work
well with one anymore; this commit removes the TEST_CACHE_NAME
env var from the release job.


bruuuuuuuce (approved):
:shipit:

PR #619, opened 2023-07-06
feat: add support for scopped access token in web-sdk

rishtigupta:
## PR Description:
Add support for scopped access token in web-sdk


packages/client-sdk-web/test/unit/internal/auth-client.test.ts
```diff
@@ -44,5 +63,148 @@ describe('internal auth client', () => {
...
+          topic: 'cat',
+        }),
+      ]);
+
+      console.log(grpcPermissions);
+
+      expect(permissionsFromScope(cacheAndTopicPermissions)).toEqual(
+        grpcPermissions
+      );
+    });
```
cprice404, line 208: @krispraws can you weigh in on these tests please?

packages/client-sdk-web/src/index.ts
```diff
@@ -95,6 +95,16 @@ import {
   TokenScope,
   ExpiresIn,
   ExpiresAt,
+  CachePermissionOptions,
+  TopicPermissionOptions,
+  CacheName,
+  isCacheName,
+  TopicName,
+  isTopicName,
```
cprice404, line 103: I would guess that we don't need these `isCacheName` and `isTopicName` to be considered part of the public API, so I wouldn't export them from this index. wdyt @krispraws ?

cprice404 (approved):
this lgtm assuming tests pass and pending @krispraws 's feedback

packages/client-sdk-web/src/index.ts
krispraws: Agree. I exported them by mistake in the nodejs index. I will fix that.

packages/client-sdk-web/src/internal/auth-client.ts
```diff
@@ -164,37 +176,111 @@ function tokenPermissionToGrpcPermission(
...
+
+  if (permission.cache === AllCaches) {
+    grpcPermission.setAllCaches(
+      new _GenerateApiTokenRequest.PermissionsType.All()
+    );
+  } else if (typeof permission.cache === 'string') {
+    cacheSelector.setCacheName(permission.cache);
+    grpcPermission.setCacheSelector(cacheSelector);
+  } else if (isCacheName(permission.cache)) {
+    cacheSelector.setCacheName(permission.cache.name);
```
krispraws, line 209: The `Fine grained authorization scenarios › can read/write cache FGA_CACHE_1 and read/write all topics in cache FGA_CACHE_2` auth integ test is failing because you are missing a `grpcPermission.setCacheSelector(cacheSelector);` here.

krispraws: Creating such a token should not have been allowed. Fixed that in https://github.com/momentohq/control-plane-service/pull/335

packages/client-sdk-web/test/unit/internal/auth-client.test.ts
krispraws: I didn't do a line by line comparison but these seem to be the same as the test from the nodejs pr. Looks fine to me.

packages/client-sdk-web/src/internal/auth-client.ts
rishtigupta: Thanks, I'll make correction in this PR too

krispraws (approved):

PR #622, opened 2023-07-07
feat: add nextjs chat demo

bruuuuuuuce:


rishtigupta:
Can we change the name of the app from "Create Next App" to "momento-chat" or something similar:
Screenshot 2023-07-10 at 1 57 57 PM

examples/web/nextjs-chat/src/app/api/momento/caches/route.ts
```diff
@@ -0,0 +1,25 @@
+import {
```
rishtigupta, line 1: This might be a stupid doubt, but why is this file named "route" ?

bruuuuuuuce: this is just a nextjs thing. Its a very opinionated way of writing fullstack code, and to add an http endpoint you need to add a route file. The path to the route file is the endpoint

https://nextjs.org/docs/app/building-your-application/routing/router-handlers

Its a bit goofy getting used to it tbh

rishtigupta (approved):

rishtigupta:
Super cool 🚢

Issue #581, opened 2023-06-13
add support for SortedSetLength and SortedSetLengthByScore


cprice404:

cprice404:
Matt is a good reference for general questions on the JS repo and SDKs.
Pratik is a good reference for these two APIs, since he just added them to golang.

PR #625, opened 2023-07-11
feat: implementation for sorted set length and sorted set length by score

anitarua:
Addresses #581 by exposing the SortedSetLength and SortedSetLengthByScore endpoints


packages/client-sdk-nodejs/src/internal/data-client.ts
```diff
@@ -2563,6 +2565,151 @@ export class DataClient implements IDataClient {
...
+          }
+        }
+      );
+    });
+  }
+
+  public async sortedSetLengthByScore(
+    cacheName: string,
+    sortedSetName: string,
+    minScore?: number,
```
bruuuuuuuce, line 2631: Nice! This is much better than with the optional booleans

packages/core/src/internal/clients/cache/AbstractCacheClient.ts
```diff
@@ -1148,6 +1150,51 @@ export abstract class AbstractCacheClient implements ICacheClient {
...
+  public async sortedSetLength(
+    cacheName: string,
+    sortedSetName: string
+  ): Promise<CacheSortedSetLength.Response> {
+    const client = this.getNextDataClient();
+    return await client.sortedSetLength(cacheName, sortedSetName);
+  }
+
+  /**
+   * Fetch length (number of items) of sorted set by score
```
bruuuuuuuce, line 1171:
```suggestion
   * Fetch length (number of items) of sorted set within the provided score range
```

packages/core/src/internal/clients/cache/AbstractCacheClient.ts
```diff
@@ -1148,6 +1150,51 @@ export abstract class AbstractCacheClient implements ICacheClient {
     );
   }
 
+  /**
+   * Fetch length (number of items) of sorted set
+   * @param {string} cacheName - The cache containing the sorted set.
+   * @param {string} sortedSetName - The sorted set to remove from.
```
bruuuuuuuce, line 1156:
```suggestion
   * @param {string} sortedSetName - The sorted set name.
```

packages/core/src/internal/clients/cache/AbstractCacheClient.ts
```diff
@@ -1148,6 +1150,51 @@ export abstract class AbstractCacheClient implements ICacheClient {
...
+    sortedSetName: string
+  ): Promise<CacheSortedSetLength.Response> {
+    const client = this.getNextDataClient();
+    return await client.sortedSetLength(cacheName, sortedSetName);
+  }
+
+  /**
+   * Fetch length (number of items) of sorted set by score
+   * @param {string} cacheName - The cache containing the sorted set.
+   * @param {string} sortedSetName - The sorted set to remove from.
```
bruuuuuuuce, line 1173:
```suggestion
   * @param {string} sortedSetName - The sorted set name.
```

packages/core/src/internal/clients/cache/AbstractCacheClient.ts
```diff
@@ -1148,6 +1150,51 @@ export abstract class AbstractCacheClient implements ICacheClient {
...
+   * @param {number} minScore - The lower bound on the score range to search in.
+   * @param {boolean} minScoreInclusive - Whether the lower bound is inclusive or exclusive.
```
bruuuuuuuce, line 1175:
```suggestion
```

packages/core/src/internal/clients/cache/AbstractCacheClient.ts
```diff
@@ -1148,6 +1150,51 @@ export abstract class AbstractCacheClient implements ICacheClient {
...
+   * @param {number} maxScore - The upper bound on the score range to search in.
+   * @param {boolean} maxScoreInclusive - Whether the upper bound is inclusive or exclusive.
```
bruuuuuuuce, line 1177:
```suggestion
```

packages/core/src/internal/clients/cache/IDataClient.ts
```diff
@@ -239,6 +241,16 @@ export interface IDataClient {
     sortedSetName: string,
     values: string[] | Uint8Array[]
   ): Promise<CacheSortedSetRemoveElement.Response>;
+  sortedSetLength(
+    cacheName: string,
+    sortedSetName: string
+  ): Promise<CacheSortedSetLength.Response>;
+  sortedSetLengthByScore(
```
bruuuuuuuce, line 248: This is more of a comment on our protos than anything in this pr, but it seems like these could have easily been a single method with optional params? Cause if you call `sortedSetLengthByScore` and don't pass in a min or max value, its the same as just called `sortedSetLength`

bruuuuuuuce: With our `listFetch` api, we pass in optional indexes, similar to our `sortedSetLengthByScore` here.

```
  listFetch(
    cacheName: string,
    listName: string,
    startIndex?: number,
    endIndex?: number
  ): Promise<CacheListFetch.Response>;
```

bruuuuuuuce (commented):
Looking good! Left a few comments related to naming and jsdocs

packages/core/src/internal/clients/cache/IDataClient.ts
anitarua: That's a good point. Should I collapse it into just one new endpoint `sortedSetLength` with the optional start and end score parameters?

anitarua: Well, I guess for consistency with the [Go SDK](https://github.com/momentohq/client-sdk-go/pull/339), it should probably remain as two endpoints.

bruuuuuuuce: ah yeah if its already in the go sdk then yea probably good to keep it consistent

bruuuuuuuce: cc @malandis @pratik151192

packages/core/src/internal/clients/cache/AbstractCacheClient.ts
```diff
@@ -1148,6 +1150,49 @@ export abstract class AbstractCacheClient implements ICacheClient {
...
+   * @returns {Promise<CacheSortedSetLengthByScore.Response>}
+   * {@link CacheSortedSetLengthByScore.Hit} containing the length if the sorted set exists.
+   * {@link CacheSortedSetLengthByScore.Miss} if the sorted set does not exist.
+   * {@link CacheSortedSetLengthByScore.Error} on failure.
+   */
+  public async sortedSetLengthByScore(
+    cacheName: string,
+    sortedSetName: string,
+    minScore?: number,
+    maxScore?: number
```
malandis, line 1185: For these arguments, I recommend using a single optional object. See for example `listFetch` (or the `sortedSetFetch` methods). In JavaScript there is no way for a caller to omit the `minScore` and pass the `maxScore`; they would have to pass `null` for `minScore` explicitly.

To get around this, the standard is to pass an object, which can take optional properties. The caller can then just omit the optional properties.

malandis (commented):
Looks good. Let's align on the public API by using an object for the optional range parameters.

packages/core/src/internal/clients/cache/IDataClient.ts
pratik151192: That's true @bruuuuuuuce; but we designed the ```fetch``` equivalent similarly as well. So it was too late to revert to clubbing under the same API.

packages/client-sdk-nodejs/src/internal/data-client.ts
```diff
@@ -2563,6 +2565,155 @@ export class DataClient implements IDataClient {
...
+      request.inclusive_min = scoreRange.minScore;
+    } else {
+      request.unbounded_min = new grpcCache._Unbounded();
+    }
+
+    if (scoreRange?.maxScore) {
+      request.inclusive_max = scoreRange.maxScore;
+    } else {
+      request.unbounded_max = new grpcCache._Unbounded();
+    }
```
malandis, line 2686: Subtle bug here: in JavaScript, `undefined` and `0` both evaluate to `false`. Thus with the code as-is, if a developer passes `minScore` (`maxScore`) as `0`, this block will then set the request `min` (`max`) to `Unbounded`, which we don't want.

Instead test explicitly for `undefined`:

```suggestion
    if (scoreRange?.minScore === undefined) {
      request.unbounded_min = new grpcCache._Unbounded();
    } else {
      request.inclusive_min = scoreRange.minScore;
    }

    if (scoreRange?.maxScore === undefined) {
      request.unbounded_max = new grpcCache._Unbounded();
    } else {
      request.inclusive_max = scoreRange.maxScore;
    }
```

You can see examples of this elsewhere in this file :)

packages/client-sdk-web/src/internal/data-client.ts
```diff
@@ -2566,6 +2570,153 @@ export class DataClient<
...
+      request.setInclusiveMin(scoreRange.minScore);
+    } else {
+      request.setUnboundedMin(new _Unbounded());
+    }
+
+    if (scoreRange?.maxScore) {
+      request.setInclusiveMax(scoreRange.maxScore);
+    } else {
+      request.setUnboundedMax(new _Unbounded());
+    }
```
malandis, line 2689: Same issue here regarding `undefined` vs `0`. Let's add some tests in the integration tests for these cases as well.

malandis (commented):
Almost there!

malandis (approved):
🎉

PR #642, opened 2023-07-13
feat: add working cloudflare example with momento http

bruuuuuuuce:


examples/cloudflare-workers/README.md
```diff
@@ -0,0 +1,28 @@
...
+npm install
+```
+
+2. Create a Momento cache inside of the [console](https://console.gomomento.com)
+3. Create a [token for your cache](https://console.gomomento.com/tokens)
+4. Copy the `Auth Token`, `Rest Endpoint`, and cache name to the `wranger.toml` file
+5. Start the development server
+
+```bash
+npm run start
```
pratik151192, line 23: Getting a 401 on opening the page; DM'd you about it but I just used my own token that I use through the CLI and got the endpoint by decoding the token

examples/cloudflare-workers/src/worker.ts
```diff
@@ -0,0 +1,86 @@
...
+	async set(cacheName: string, key: string, value: string, ttl_milliseconds: number = 3000) {
+		const resp = await fetch(`${this.baseurl}/${cacheName}?key=${key}&auth_token=${this.apiToken}&&ttl_milliseconds=${ttl_milliseconds}`, {
+			method: 'PUT',
+			body: value
+		});
+
+		if (resp.status < 300) {
+			console.log(`successfully set ${key} into cache`);
+		} else {
+			throw new Error(`failed to set item into cache message: ${resp.statusText} status: ${resp.status}`);
```
pratik151192, line 39: nit: maybe put the cache name as well in this message

examples/cloudflare-workers/src/worker.ts
```diff
@@ -0,0 +1,86 @@
...
+			throw new Error(`failed to delete ${key} from cache. Message: ${resp.statusText}; Status: ${resp.status}`);
+		}
+
+		return resp;
+	}
+}
+
+export interface Env {
+	MOMENTO_AUTH_TOKEN: string;
+	MOMENTO_REST_ENDPOINT: string;
```
pratik151192, line 61: If we can infer the endpoint through the token, can we just infer it programmatically in this example? Like I am not sure how beneficial is to copy from the console if they're creating applications where they hard code it anyway. Console is helpful for ```curl``` and other-use cases though

examples/cloudflare-workers/src/worker.ts
```diff
@@ -0,0 +1,86 @@
...
+class MomentoFetcher {
+	private readonly apiToken: string;
+	private readonly baseurl: string;
+	constructor(token: string, endpoint: string) {
+		this.apiToken = token;
+		this.baseurl = `https://api.cache.${endpoint}/cache`;
+	}
+
+	async get(cacheName: string, key: string) {
+		const resp = await fetch(`${this.baseurl}/${cacheName}?key=${key}&auth_token=${this.apiToken}`);
```
bruuuuuuuce, line 20:
```suggestion
		const resp = await fetch(`${this.baseurl}/${cacheName}?key=${key}&token=${this.apiToken}`);
```

examples/cloudflare-workers/src/worker.ts
```diff
@@ -0,0 +1,86 @@
...
+			console.log(`successfully set ${key} into cache`);
+		} else {
+			throw new Error(`failed to set item into cache message: ${resp.statusText} status: ${resp.status}`);
+		}
+
+		return;
+	}
+
+	async delete(cacheName: string, key: string) {
+		const resp = await fetch(`${this.baseurl}/${cacheName}?key=${key}&auth_token=${this.apiToken}`, {
```
bruuuuuuuce, line 46:
```suggestion
		const resp = await fetch(`${this.baseurl}/${cacheName}?key=${key}&token=${this.apiToken}`, {
```

examples/cloudflare-workers/src/worker.ts
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * Learn more at https://developers.cloudflare.com/workers/
+ */
+
+class MomentoFetcher {
+	private readonly apiToken: string;
+	private readonly baseurl: string;
+	constructor(token: string, endpoint: string) {
+		this.apiToken = token;
+		this.baseurl = `https://api.cache.${endpoint}/cache`;
```
danielamiao, line 16: Chatted with Matt on Slack but basically I think we should assume the user gets the REST endpoint from the docs or console, and they are copying and pasting it here. So instead of prepending `api.cache.` to the parameter passed in, we should just use the parameter as is

examples/cloudflare-workers/src/worker.ts
bruuuuuuuce: will change this in a future pr

pratik151192 (approved):

PR #643, opened 2023-07-13
chore: adds keyv and apollo examples for docs

eaddingtonwhite:


examples/nodejs/cache/.eslintrc.json
```diff
@@ -36,6 +36,10 @@
     "prefer-arrow-callback": "error",
     "no-trailing-spaces": "error",
     "quotes": ["warn", "single", {"avoidEscape": true}],
+    "@typescript-eslint/no-unsafe-call": 0,
+    "@typescript-eslint/no-unsafe-assignment": 0,
+    "@typescript-eslint/no-unsafe-member-access": 0,
+    "node/no-extraneous-import": 0,
```
cprice404, line 42: do we want to disable these?

eaddingtonwhite, line 42: needed these to get build passing w/ these librarys. Wondering if should break these into separate project

.github/workflows/build.yml
```diff
@@ -81,7 +81,7 @@ jobs:
   test-examples:
     strategy:
       matrix:
-        node: [14, 16, 18]
+        node: [16, 18]
```
cprice404, line 84: Are we not planning to support 14 anymore? I know it's EOL and I'm open to the convo, just don't want it to be an arbitrary decision.

examples/nodejs/cache/package.json
```diff
@@ -27,9 +27,14 @@
...
+    "typescript": "^5.0.4"
   },
   "dependencies": {
-    "@gomomento/sdk": "1.27.2"
+    "@apollo/server-plugin-response-cache": "^4.1.3",
+    "@apollo/utils.keyvadapter": "^3.0.0",
+    "@gomomento-poc/node-keyv-adaptor": "^0.3.0",
+    "keyv": "^4.5.2",
+    "@apollo/server": "^4.7.5",
+    "@gomomento/sdk": "1.28.0"
```
cprice404, line 38: thanks for adding the examples! but the reason we broke the examples directory up into subdirs is so that the package.json file for the basic cache examples won't have a bunch of dependencies that most users don't need. Can we create a new subdir with a new package.json that adds all these deps?

.github/workflows/build.yml
eaddingtonwhite: @cprice404 left comment here 
https://github.com/momentohq/client-sdk-javascript/pull/643/files#r1265604485

Dont have super strong opinions on this. I think would be nice to not support 14 since its EOL but could also break this out to a `3rdParty-examples` project where we run on smaller subsets of node for examples. Would need to update docs as well then to pull dynamically from there also.

examples/nodejs/cache/.eslintrc.json
eaddingtonwhite: @cprice404 sorry I left comments after tagging u. left comment here
https://github.com/momentohq/client-sdk-javascript/pull/643/files#r1265604775

.github/workflows/build.yml
```diff
@@ -81,7 +81,7 @@ jobs:
   test-examples:
     strategy:
       matrix:
-        node: [14, 16, 18]
```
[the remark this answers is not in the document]
cprice404: i'm totally fine with having two sections for examples in CI, one that runs some subdirs with 14 and one that runs without 14. would like to keep the basic cache examples in the 14 section as long as there are no major issues for building them for 14, or until we talk about officially discontinuing support for 14.

totally fine with the examples that you are adding here going into the non-14 section in CI though.

I think I already made a non-14 section in CI for web SDK. we could just rename it.

cprice404:
@eaddingtonwhite do you have a plan for this?

Issue #614, opened 2023-07-05
Add support for updateTtl/increaseTtl/decreaseTtl


cprice404:
Matt is a good resource if you have JS questions.
One of these is implemented in the .NET SDK and that is the one that I'd use as a reference.
The Go SDK has all 3 of them if you need another reference.

Michael is the best person to review the final PR.

PR #644, opened 2023-07-14
feat: update/increase/decrease ttl

anitarua:
Addresses #614 by adding the update/increase/decrease TTL APIs to the JS SDK


malandis (commented):
Looks good!

One detail about increase and decrease TTL -- they are conditional on the new TTL being more (less) than the existing TTLs. Can you document this behavior in the public API and also add integration tests?

rishtigupta (dismissed):

malandis (approved):
🎉

Issue #612, opened 2023-07-03
Add example using AWS secrets manager


pratik151192:
More details here https://github.com/momentohq/public-dev-docs/pull/335

cprice404:
Matt and Pratik can provide more context on this.

PR #648, opened 2023-07-17
chore: add aws secrets example

anitarua:
Addresses #612, creates the example in the JS SDK to pull into the [AWS Secrets doc page](https://docs.momentohq.com/develop/integrations/aws-secrets-manager#example-code-for-aws-secrets-manager).


examples/nodejs/cache/package.json
```diff
@@ -30,6 +31,7 @@
     "typescript": "4.4.3"
   },
   "dependencies": {
+    "@aws-sdk/client-secrets-manager": "^3.370.0",
```
cprice404, line 34: Overall this PR looks great! One small detail - this "cache" example directory is supposed to be only the most basic things that people would need to get started with momento. If we add a secrets manager dep here, and people just copy this package.json file, then they might be taking on that extra secrets manager dependency unnecessarily and it can make their build a lot bigger.

So for this directory we only want to show examples that don't have any third-party dependencies.

I think we should have an `aws` examples directory that we put this into, which may contain lots of other aws examples in the future.

So you'll basically need to create a new directory and copy in the boilerplate files (package.json, tsconfig.json, .eslint*, .prettier*, etc.) and then move your script over there.

cprice404 (commented):
looking good!

examples/nodejs/cache/doc-example-files/doc-examples-js-aws-secrets.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  /* Try-catch block that gets the Momento_Auth_Token stored in AWS Secrets Manager.
+  The secret was stored as a plaintext format in Secrets Manager to avoid parsing JSON.
+  You don't have to store the Momento auth token in something like AWS Secrets Manager,
+  but it is best practice. You could pass the Momento auth token in from an environment variable.
+  */
+  try {
+    const client = new SecretsManagerClient({region: regionName});
+    const response = await client.send(
+      new GetSecretValueCommand({
```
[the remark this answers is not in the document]
anitarua: Yup! I can run this locally after copy-pasting my AWS tokens into the terminal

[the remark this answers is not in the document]
cprice404: nice! we should get you set up with an aws config file and our dev tool soon so you don't have to use those env vars, but that is great for now!

examples/nodejs/aws/.gitignore
```diff
@@ -0,0 +1 @@
+advanced-middlewares-example-metrics.csv
```
cprice404, line 1: this is harmless but you probably don't need it :)

examples/nodejs/aws/README.md
```diff
@@ -0,0 +1,37 @@
...
+[![project status](https://momentohq.github.io/standards-and-practices/badges/project-status-official.svg)](https://github.com/momentohq/standards-and-practices/blob/main/docs/momento-on-github.md)
+[![project stability](https://momentohq.github.io/standards-and-practices/badges/project-stability-stable.svg)](https://github.com/momentohq/standards-and-practices/blob/main/docs/momento-on-github.md)
+
+# Node.js Client SDK
+
+
+## Example Requirements
+
+- Node version 14 or higher is required
+- A Momento Auth Token is required, you can generate one using the [Momento CLI](https://github.com/momentohq/momento-cli)
```
cprice404, line 15: This is totally not your fault but this string is out of date. We don't support generating tokens from the CLI anymore; you can see the new verbiage in the generated READMEs at the root of each SDK but it basically looks like this now:

```
To get started with Momento you will need a Momento Auth Token. You can get one from the [Momento Console](https://console.gomomento.com).
```

If you happen to notice the old text hanging around in other READMEs, bonus points for updating them :)

cprice404 (commented):
looks good! one last thing we need to do is wire up the CI and dependabot workflows to cover this new example dir.

That will happen in these two spots:

https://github.com/momentohq/client-sdk-javascript/blob/main/.github/workflows/build.yml#L97-L115

and here:

https://github.com/momentohq/client-sdk-javascript/blob/main/.github/dependabot.yml#L29-L34

examples/nodejs/topics/README.md
```diff
@@ -15,7 +15,7 @@ _Read this in other languages_: [日本語](README.ja.md)
 ## Example Requirements
 
 - Node version 14 or higher is required
-- A Momento Auth Token is required, you can generate one using the [Momento CLI](https://github.com/momentohq/momento-cli)
+- To get started with Momento you will need a Momento Auth Token. You can get one from the [Momento Console](https://console.gomomento.com).
```
cprice404, line 18: Thanks for fixing these!

cprice404 (approved):

PR #661, opened 2023-07-19
fix: lazily instantiate session token cred provider during tests

cprice404:
Session tokens are complicated to acquire, and the tests that require
them are not relevant for most dev work on this package. Prior to this
commit we were eagerly constructing CredentialProviders that used
session tokens on every run of the integration test suite, regardless
of whether we would actually be executing the related tests or not.

This commit changes the instantiation of the session token CredProvider
to be lazy, so that it is not necessary to provide the session token
env var on test runs that do not exercise it.


anitarua (approved):
lgtm!

bruuuuuuuce (approved):
:shipit:

PR #663, opened 2023-07-21
feat: allow configurable token vending machine (permission scope and expiry)

anitarua:
Addresses the first part of [#385](https://github.com/momentohq/dev-eco-issue-tracker/issues/385) by creating a new config file where users can specify token permissions and expiry before deploying the token vending machine. Instructions are in the README and examples are in the `config.ts` file.

The TVM CDK stack can be successfully deployed and a locally running vite-chat-app will get a new token from the API gateway endpoint. However, the app was expecting a string rather than a JSON object (which is what the TVM was providing) so I made a change there to accept JSON instead. The app was also using the wrong environment variable name for the cache name so I fixed that as well. I also had to include the CORS header in the Lambda response so that the app would connect to it.

### DIFF
--- a/examples/nodejs/token-vending-machine/README.md
+++ b/examples/nodejs/token-vending-machine/README.md
@@ -23,11 +23,18 @@ The primary use for the Token Vending Machine is to provide temporary, restricte
 
 ## Configuring the Token Vending Machine App
 
-Before you deploy the token vending machine, you will need to configure the scope of permissions and the expiry date for the tokens that it will vend.
+Before you deploy the token vending machine, you will need to configure the scope of permissions and the expiry duration for the tokens that it will vend. For example, you can restrict the permissions for these browser tokens so that they have read-only access or read-write access, and you can also restrict them to specific caches or topics.
 
 These two required variables live in the [lambda/config.ts](./lambda/config.ts) file.
 
 ## Deploying the Token Vending Machine App
+
+First make sure to start Docker and install the dependencies in the `lambda` directory, which is where the AWS Lambda code lives.
+
+```bash
+cd lambda
+npm install
+```
 
 The source code for the CDK application lives in the `infrastructure` directory.
 To build and deploy it you will first need to install the dependencies:
--- a/examples/nodejs/token-vending-machine/README.md
+++ b/examples/nodejs/token-vending-machine/README.md
@@ -21,6 +21,12 @@ This repo contains an example Token Vending Machine application, built using AWS
 
 The primary use for the Token Vending Machine is to provide temporary, restricted scope Momento Auth tokens. These tokens can be used by browsers that are running apps written against the [Momento Web SDK](https://github.com/momentohq/client-sdk-javascript/tree/main/packages/client-sdk-web). For example, you can create a browser-based chat application that allows pub/sub communication between your users via [Momento Topics](https://docs.momentohq.com/introduction/momento-topics); each browser will need a Momento auth token in order to communicate with the Momento Topics server, and the Token Vending Machine can provide those tokens.
 
+## Configuring the Token Vending Machine App
+
+Before you deploy the token vending machine, you will need to configure the scope of permissions and the expiry date for the tokens that it will vend.
+
+These two required variables live in the [lambda/config.ts](./lambda/config.ts) file.
+
 ## Deploying the Token Vending Machine App
 
 The source code for the CDK application lives in the `infrastructure` directory.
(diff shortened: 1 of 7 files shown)


examples/web/vite-chat-app/src/App.tsx
```diff
@@ -98,7 +98,7 @@ export default function Home() {
   return (
     <ChatRoom
       topicName={topic}
-      cacheName={import.meta.env.MOMENTO_CACHE_NAME}
+      cacheName={import.meta.env.VITE_MOMENTO_CACHE_NAME}
```
bruuuuuuuce, line 101: thanks for fixing this!

bruuuuuuuce (approved):
:shipit:

anitarua:
Since the configuration file is just two variables that have to be set, I could implement the same approach with the NextJS app and bundle it into this PR as well if that makes sense to do.

examples/nodejs/token-vending-machine/README.md
```diff
@@ -21,6 +21,12 @@ This repo contains an example Token Vending Machine application, built using AWS
 
 The primary use for the Token Vending Machine is to provide temporary, restricted scope Momento Auth tokens. These tokens can be used by browsers that are running apps written against the [Momento Web SDK](https://github.com/momentohq/client-sdk-javascript/tree/main/packages/client-sdk-web). For example, you can create a browser-based chat application that allows pub/sub communication between your users via [Momento Topics](https://docs.momentohq.com/introduction/momento-topics); each browser will need a Momento auth token in order to communicate with the Momento Topics server, and the Token Vending Machine can provide those tokens.
 
+## Configuring the Token Vending Machine App
+
+Before you deploy the token vending machine, you will need to configure the scope of permissions and the expiry date for the tokens that it will vend.
```
cprice404, line 26: Can we add one more sentence here to give it a bit more color?

something like "For example, you can restrict the permissions for these browser tokens so that they have read-only access or read-write access, and you can also restrict them to specific caches or topics."

cprice404, line 26:
```suggestion
Before you deploy the token vending machine, you will need to configure the scope of permissions and the expiry duration for the tokens that it will vend.
```

examples/nodejs/token-vending-machine/lambda/config.ts
```diff
@@ -0,0 +1,44 @@
...
+ *        },
+ *        {
+ *          role: TopicRole.PublishSubscribe | TopicRole.SubscribeOnly, 
+ *          cache: AllCaches | "your-cache-name",
+ *          topic: AllTopics | "your-topic-name"
+ *        }
+ *      ]
+ *    };
+ * 
+ * More information here: https://github.com/momentohq/client-sdk-javascript/blob/62a34741059cb08c8dfff4e7011be29facab8d80/packages/core/src/auth/tokens/token-scope.ts
```
cprice404, line 24: i love that you used the permalink for the gh link instead of HEAD :)

but I think for these readme's let's try to steer people toward the dev docs when we can.  so for this one we can use:

https://docs.momentohq.com/develop/sdks/web#credentials-for-browsers

examples/nodejs/token-vending-machine/lambda/config.ts
```diff
@@ -0,0 +1,44 @@
...
+      role: TopicRole.PublishSubscribe, 
+      cache: "default-cache",
+      topic: AllTopics
+    }
+]};
+
+/**
+ * Set the TTL for your tokens in terms of seconds, minutes, hours,  
+ * days, or using epoch format. You may also set tokens to never expire.
+ * More information here: https://github.com/momentohq/client-sdk-javascript/blob/62a34741059cb08c8dfff4e7011be29facab8d80/packages/core/src/utils/expiration.ts#L17
```
cprice404, line 42: for this one, maybe:

https://docs.momentohq.com/develop/api-reference/auth-tokens#generateauthtoken-api ?

or feel free to poke around the dev docs site and see if you find something better. or add something better to the dev docs if you feel like there's nothing good enough there yet :) but i think i'd be good with the link above.

examples/nodejs/token-vending-machine/lambda/handler.ts
```diff
@@ -15,6 +16,10 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
     const vendedAuthToken = await vendAuthToken(vendorAuthTokenSecretName);
     return {
       statusCode: 200,
+      headers: {
+        "Content-Type": "application/json",
+        "Access-Control-Allow-Origin": "*"
+      },
```
cprice404, line 22: This was a great find/fix, thanks.

cprice404 (commented):
awesome job. I just had a couple of nits about the readme text but otherwise 🚢 

if I'm not around when you update the PR, feel free to merge it once @bruuuuuuuce approves it again.

bruuuuuuuce (approved):
:shipit:

PR #666, opened 2023-07-21
feat: allow configurable tokens in nextjs app example

anitarua:
Addresses the first part of [#385](https://github.com/momentohq/dev-eco-issue-tracker/issues/385) but this time for the NextJS example app. Since we're really just setting two variables, I figured it would be overkill to try to extract the same code into a separate SDK function or library, but open to suggestions!

Similar to #663, I created a new config file where users can specify token permissions and expiry before deploying the token vending machine. Instructions are in the README and examples are in the `src/app/api/momento/token/config.ts` file. (I am also open to moving the config file out of the token directory. My first thought was that keeping them together would make more sense but if it seems too difficult for people to find, definitely open to moving it.)


examples/web/nextjs-chat/src/app/api/momento/token/config.ts
```diff
@@ -0,0 +1,52 @@
...
+ *        }
+ *      ]
+ *    };
+ * 
+ * More information here: https://docs.momentohq.com/develop/api-reference/auth-tokens#tokenscope-objects
+ */
+export const tokenPermissions: TokenScope = {
+  permissions: [
+    {
+      role: CacheRole.ReadWrite,
```
bruuuuuuuce, line 37: This would actually be cool to be part of the api, maybe query params or something. The user could select a cache and a topic, and then we can make a request to get a scoped token for that specific cache and topic

anitarua: Would that be similar to this? 
https://github.com/momentohq/client-sdk-javascript/blob/2f6541f169644b4025f262b8a489fb71762ca015/packages/core/src/auth/tokens/token-scopes.ts#L4

Would there need to be a way to stack multiple permissions though?

bruuuuuuuce: Well I was thinking for the demo would be cool for it to just scope it down as low as possible. But we could also just accept a post body of all the permissions that the user wants the token to have

cprice404: I think the point of Anita's config file is to show app owners/maintainers how to restrict the scope (at app deploy time) so that the browser can't get tokens that have permissions to unexpected/unwanted caches. If we make it super dynamic then it's not really that different from just giving out an AllDataReadWrite token.

We can discuss this more for follow-up PRs, if there is something dynamic that is cool and worth demo'ing I'm open to talking about it! But for now this is more about showing how to make it secure.

bruuuuuuuce: I think the issue that I see here is that now the demo only works if a user has a cause `default-cache` already created

examples/web/nextjs-chat/src/app/api/momento/token/config.ts
```diff
@@ -0,0 +1,46 @@
...
+ *          role: TopicRole.PublishSubscribe | TopicRole.SubscribeOnly, 
+ *          cache: AllCaches | "your-cache-name",
+ *          topic: AllTopics | "your-topic-name"
+ *        }
+ *      ]
+ *    };
+ * 
+ * More information here: https://docs.momentohq.com/develop/api-reference/auth-tokens#tokenscope-objects
+ */
+export const tokenPermissions: TokenScope = TokenScopes.topicPublishSubscribe("default-cache", AllTopics);
```
cprice404, line 39: Something that I missed earlier - Matt's app allows users to dynamically select the cache from a list of existing caches. Therefore if we default to a permission that only allows a specific cache, the app will break if the user chooses a different cache from the dialog.

So I think I would change this to default to AllCaches, but possibly include your current line, commented out, with a sentence that says "to restrict browsers to only accessing a specific cache, uncomment this line" or something like that.

cprice404 (commented):
looks good! just one minor suggestion about the default state of the config file.

examples/web/nextjs-chat/src/app/api/momento/token/config.ts
cprice404: yeah I agree with that Matt, see my other comment and see if that addresses your concern

cprice404: tbh i wonder if we should just take the cache chooser out of the app. doesn't have to be in this PR but just thinking.

examples/web/nextjs-chat/src/app/api/momento/token/config.ts
anitarua: Sounds good! I changed the example in the large comment block to show it can be `AllCaches` or a specific cache name

examples/web/nextjs-chat/src/app/api/momento/token/config.ts
```diff
@@ -17,7 +17,7 @@ import {
  * 
  * TokenScopes provides several functions that will return the permissions you
  * request for a given cache and topic name.
- *    export const tokenPermissions: TokenScope = TokenScopes.topicPublishSubscribe("default-cache", AllTopics);
+ *    export const tokenPermissions: TokenScope = TokenScopes.topicPublishSubscribe(AllCaches | "your-cache-name", AllTopics);
```
cprice404, line 20: in case folks are tempted to copy/paste this, I wouldn't use the `|` here because it will do something surprising :)
I would just have two copies of this line, one with `AllCaches` and one with the string version.

cprice404 (commented):
one tiny nit in the doc comment and then we should be good to go!

examples/web/nextjs-chat/src/app/api/momento/token/config.ts
```diff
@@ -17,7 +17,10 @@ import {
  * 
  * TokenScopes provides several functions that will return the permissions you
  * request for a given cache and topic name.
- *    export const tokenPermissions: TokenScope = TokenScopes.topicPublishSubscribe(AllCaches | "your-cache-name", AllTopics);
+ *    export const tokenPermissions: TokenScope = TokenScopes.topicPublishSubscribe("your-cache-name", AllTopics);
+ * 
+ * You can also set it to subscribe to all caches if you prefer:
```
cprice404, line 22:
```suggestion
 * You can also set it to allow subscriptions to topics in all caches if you prefer:
```

cprice404 (approved):

PR #675, opened 2023-07-24
feat: add logging implementation of experimental metrics middleware

cprice404:
This commit refactors the "Experimental Metrics CSV" middleware to
be an abstract class that is destination-agnostic. This will allow
us to provide implementations of it that write to other targets
besides CSV. Also included in this PR is an implementation that
writes the metrics to the logger, which is a slightly more scalable
destination for most of our users to consider enabling.

### DIFF
--- a/packages/client-sdk-nodejs/src/config/middleware/experimental-metrics-logging-middleware.ts
+++ b/packages/client-sdk-nodejs/src/config/middleware/experimental-metrics-logging-middleware.ts
@@ -1,0 +1,47 @@
+import {MomentoLogger, MomentoLoggerFactory} from '../../';
+import {
+  ExperimentalMetricsMiddleware,
+  ExperimentalMetricsMiddlewareRequestHandler,
+  ExperimentalRequestMetrics,
+} from './impl/experimental-metrics-middleware';
+
+class ExperimentalMetricsLoggingMiddlewareRequestHandler extends ExperimentalMetricsMiddlewareRequestHandler {
+  constructor(parent: ExperimentalMetricsMiddleware, logger: MomentoLogger) {
+    super(parent, logger);
+  }
+
+  emitMetrics(metrics: ExperimentalRequestMetrics): Promise<void> {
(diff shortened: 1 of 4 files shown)


packages/client-sdk-nodejs/src/config/middleware/experimental-metrics-logging-middleware.ts
```diff
@@ -0,0 +1,47 @@
...
+ * The metrics format is currently considered experimental; in a future release,
+ * once the format is considered stable, this class will be renamed to remove
+ * the Experimental prefix.
+ *
+ * WARNING: enabling this middleware may have minor performance implications,
+ * so enable with caution.
+ *
+ * WARNING: depending on your request volume, the CSV file size may grow quickly;
+ * neither sampling nor file compression / rotation are included at this time
+ * (though they may be added in the future).
```
pratik151192, line 33: This makes me think that very few people will be comfortable enbaling this in their production workflows

pratik151192 (dismissed):

packages/client-sdk-nodejs/src/config/middleware/experimental-metrics-logging-middleware.ts
cprice404: these comments are wrong, this implementation doesn't actually write to a CSV file. will fix.

cprice404: added a commit to fix this comment in https://github.com/momentohq/client-sdk-javascript/commit/53e734bc7334d4275dee7010c4a05afe747774d2 .

nand4011 (approved):
Looks good. My only comment is that if feels strange that the abstract class is in the impl directory instead of the concrete versions.

pratik151192 (approved):

cprice404:
> Looks good. My only comment is that if feels strange that the abstract class is in the impl directory instead of the concrete versions.

fair, maybe i'll try to think of a better name for that dir in a subsequent commit

PR #677, opened 2023-07-26
feat: authentication options for the token vending machine

anitarua:
Addresses the tasks in [#385](https://github.com/momentohq/dev-eco-issue-tracker/issues/385) regarding making auth configurable for the token vending machine and adding examples. There is a new config variable that is used in the CDK appliance to determine whether to attach one of the authorizers to the GET endpoint.

The Lambda Authorizer example simply uses a hardcoded username/password check to determine whether to allow access to the TVM endpoint. The endpoint expects a couple of headers (`username` and `password`) as part of the GET request. I have verified this works with the Vite app with minimal changes (passing the headers and configuring API Gateway to expect those headers).

[edited to reflect new CDK stack] The Cognito example creates a user pool, user pool client, and a demo user. The flow to be able to get Momento tokens from the TVM now looks like: sign in to Cognito using username, password, and client ID to be authenticated; pass the returned ID token under the `Authorization` header in the GET request to the TVM endpoint. I haven't yet verified that this works with the Vite app as it seems it will take more setup than the other one, but including the user and client creation in the CDK has eliminated a few steps there.

Finally, the remaining changes involve updates to the docs and `package.json` files.


examples/nodejs/token-vending-machine/lambda/config.ts
```diff
@@ -41,4 +45,16 @@ export const tokenPermissions: TokenScope = {
  * days, or using epoch format. You may also set tokens to never expire.
  * More information here: https://docs.momentohq.com/develop/api-reference/auth-tokens#generateauthtoken-api
  */
-export const tokenExpiresIn: ExpiresIn = ExpiresIn.hours(1);
\ No newline at end of file
+export const tokenExpiresIn: ExpiresIn = ExpiresIn.hours(1);
+
+/**
```
cprice404, line 50: let's create an enum called AuthenticationMethod just above your comments here, and provide the three legal values. That way it's not possible for a user to configure it to an invalid value.

examples/nodejs/token-vending-machine/README.md
```diff
@@ -23,9 +23,11 @@ The primary use for the Token Vending Machine is to provide temporary, restricte
...
+Before you deploy the Token Vending Machine, you will need to configure the scope of permissions and the expiry duration for the tokens that it will vend. For example, you can restrict the permissions for these browser tokens so that they have read-only access or read-write access, and you can also restrict them to specific caches or topics.
 
-These two required variables live in the [lambda/config.ts](./lambda/config.ts) file.
+You will also need to specify the authentication method for the Token Vending Machine's API Gateway Endpoint. You may choose to leave it open, allowing the API Gateway URL to be publicly accessible, or you may choose to use Lambda Authorizer or Amazon Cognito as an authenticator. Basic examples for each of these methods is provided in the [Lambda Authorizer handler](./lambda/authorizer/authorizer.ts) and [CDK definition](./infrastructure/lib/token-vending-machine-stack.ts) files.
```
cprice404, line 28: nice

examples/nodejs/token-vending-machine/infrastructure/lib/token-vending-machine-stack.ts
```diff
@@ -46,6 +48,48 @@ export class TokenVendingMachineStack extends cdk.Stack {
       requestTemplates: {'application/json': '{ "statusCode": "200" }'},
     });
 
-    api.root.addMethod('GET', tvmIntegration);
+    if (config.authenticationMethod === "lambda-authorizer") {
```
cprice404, line 51: stylistic nit - when branching on an enum like this, i usually find a `switch` statement to be more readable. but it's subjective :)

in either case, whether you stick with an `if` or move to a `switch`, you should have a final `else` or `default` branch that throws an error if you get an unexpected value. That way if we add a new option to the enum in the future and forget to update the code here, we'll at least find out about it at runtime.

examples/nodejs/token-vending-machine/infrastructure/lib/token-vending-machine-stack.ts
```diff
@@ -46,6 +48,48 @@ export class TokenVendingMachineStack extends cdk.Stack {
...
+      const userPool = new cognito.UserPool(this, 'MomentoTokenVendingMachineUserPool', {
+        userPoolName: 'MomentoTokenVendingMachineUserPool',
+        signInAliases: {
+          username: true
+        },
+        passwordPolicy: {
+          minLength: 8,
+          requireSymbols: true,
+        }
+      });
```
cprice404, line 83: just out of curiosity, does CDK have a resource type for users in the pool here? i.e. since you are defining the pool in CDK, can we also add members to it in CDK?

examples/nodejs/token-vending-machine/lambda/authorizer/authorizer.ts
```diff
@@ -0,0 +1,35 @@
+import { APIGatewayAuthorizerResult, APIGatewayRequestAuthorizerEvent } from 'aws-lambda';
+
+// Lambda function for API Gateway Lambda Authorizer
+// Input format: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-lambda-authorizer-input.html 
+// Output format: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-lambda-authorizer-output.html 
+export const handler = async (event: APIGatewayRequestAuthorizerEvent): Promise<APIGatewayAuthorizerResult> => {
```
cprice404, line 6: in Lambda code it's a good idea to add a few `console.log` statements to help the user debug if something goes wrong. I would probably add a log statement indicating whether the headers were found or not, one that indicates "Allow", and one that gives a hint as to why we were denied if we are denied.

you can see these log outputs in the AWS lambda console or in the AWS CloudWatch console after your lambda executes.

examples/nodejs/token-vending-machine/lambda/authorizer/authorizer.ts
```diff
@@ -0,0 +1,35 @@
...
+    principalId: principalId,
+    policyDocument: {
+      Version: '2012-10-17',
+      Statement: [{
+        Action: 'execute-api:Invoke',
+        Effect: effect,
+        Resource: resource
+      }]
+    }
+  };
```
cprice404, line 33: nicely done, this stuff can be tricky.

examples/nodejs/token-vending-machine/lambda/authorizer/package.json
```diff
@@ -0,0 +1,33 @@
+{
```
cprice404, line 1: npm sometimes does very weird/unexpected things when you have a package.json file in a subdirectory of another dir that also has a package.json.

Roughly: https://flaviocopes.com/npm-install-packages-parent-folder/

So I would recommend creating a new subdir `lambda/token-vending-machine` for the previous lambda code, and then you can have that as a sibling dir for your new `authorizer` dir.

examples/nodejs/token-vending-machine/lambda/config.ts
```diff
@@ -1,11 +1,15 @@
...
 
 /**
  * Set the scope of permissions for your tokens. 
  * 
  * AllDataReadWrite provides read and write permissions to all of your caches:
  *    export const tokenPermissions: TokenScope =  AllDataReadWrite;
  * 
+ * TokenScopes provides several functions that will return the permissions you
+ * request for a given cache and topic name.
+ *    export const tokenPermissions: TokenScope = TokenScopes.topicPublishSubscribe("default-cache", AllTopics);
```
cprice404, line 11: nice addition to the docs, thank you

examples/nodejs/token-vending-machine/lambda/package.json
```diff
@@ -1,16 +1,11 @@
...
+  "description": "Momento token vending machine AWS Lambda function",
   "main": "index.ts",
   "scripts": {
     "prebuild": "eslint . --ext .ts",
     "build": "tsc",
-    "basic": "tsc && node dist/basic.js",
-    "advanced": "tsc && node dist/advanced.js",
-    "dictionary": "tsc && node dist/dictionary.js",
-    "doc-examples": "tsc && node dist/doc-example-files/doc-examples-js-apis.js",
-    "validate-examples": "tsc && node dist/readme.js && node dist/basic.js && node dist/advanced.js && node dist/dictionary.js && node dist/doc-example-files/doc-examples-js-apis.js && node dist/doc-example-files/cheat-sheet-main.js",
```
cprice404, line 13: whoops, thank you for catching these

cprice404 (commented):
nice job! handful of minor comments

cprice404:
> I have verified this works with the Vite app with minimal changes (passing the headers and configuring API Gateway to expect those headers).

have you thought about how we can add some configurability to the vite app to support this as well? would love to do that in a follow-up PR.

Same thing with Cognito but I think we can go ahead and get this PR merged and then update the Vite app to use the lambda authorizer first, and then circle back for cognito.

examples/nodejs/token-vending-machine/infrastructure/lib/token-vending-machine-stack.ts
anitarua: It might, I can take a deeper look into that today

anitarua:
> have you thought about how we can add some configurability to the vite app to support this as well? would love to do that in a follow-up PR.
> 
> Same thing with Cognito but I think we can go ahead and get this PR merged and then update the Vite app to use the lambda authorizer first, and then circle back for cognito.

Yeah, I think there could probably be another few environment variables that the Vite app can use to determine which authentication route to take and what username/password/clientId values to use to connect to the TVM. I can certainly mess around with it more to see if there's a way to easily incorporate configurability for both lambda authorizer and cognito.

cprice404:
> > have you thought about how we can add some configurability to the vite app to support this as well? would love to do that in a follow-up PR.
> > Same thing with Cognito but I think we can go ahead and get this PR merged and then update the Vite app to use the lambda authorizer first, and then circle back for cognito.
> 
> Yeah, I think there could probably be another few environment variables that the Vite app can use to determine which authentication route to take and what username/password/clientId values to use to connect to the TVM. I can certainly mess around with it more to see if there's a way to easily incorporate configurability for both lambda authorizer and cognito.

let's start with the lambda authorizer one and then we can handle cognito in another PR if we decide to do it.

examples/nodejs/token-vending-machine/infrastructure/lib/token-vending-machine-stack.ts
```diff
@@ -48,48 +49,145 @@ export class TokenVendingMachineStack extends cdk.Stack {
...
+  // reference: https://github.com/awesome-cdk/cdk-userpool-user/blob/master/lib/UserPoolUser.ts 
+  private createCognitoUser(
+    scope: Construct,
+    userPool: cognito.IUserPool, 
+    username: string, 
+    password: string, 
+    group?: string
+  ) {
+    // Basically use the AWS SDK to fill in this gap in the CDK for creating a user with a password
+    const newUser = new AwsCustomResource(scope, 'AwsCustomResource-CreateCognitoUser', {
```
cprice404, line 148: wow, you made a custom resource! those have a pretty steep learning curve, nice job :)

cprice404 (approved):
awesome! 🚢

PR #681, opened 2023-07-27
feat: vite chat app with authenticated token vending machine

anitarua:
Follow up to #677 to show how the Vite app can be configured to use the token vending machine with any of the auth options (no auth, lambda authorizer, or cognito user pool).


cprice404 (approved):
this looks fantastic, thank you!!

PR #683, opened 2023-07-28
fix: update dependencies to resolve dependabot alert

pgautier404:
This commit updates the version of the `@grpc/grpc-js` dependency such that its transitive dependency with `protobufjs` is bumped to the latest: 7.2.4. Prior versions had a vulnerability which is addressed in this latest version.

### DIFF
--- a/examples/nodejs/observability/package.json
+++ b/examples/nodejs/observability/package.json
@@ -30,7 +30,7 @@
   },
   "dependencies": {
     "@gomomento/sdk": "1.28.0",
-    "@grpc/grpc-js": "1.8.20",
+    "@grpc/grpc-js": "1.8.17",
     "@opentelemetry/api": "1.4.1",
     "@opentelemetry/exporter-prometheus": "0.39.1",
     "@opentelemetry/exporter-zipkin": "1.13.0",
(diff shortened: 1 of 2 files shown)


examples/nodejs/observability/package.json
```diff
@@ -29,8 +29,8 @@
     "typescript": "4.4.3"
   },
   "dependencies": {
-    "@gomomento/sdk": "1.26.2",
-    "@grpc/grpc-js": "1.8.14",
+    "@gomomento/sdk": "1.28.0",
+    "@grpc/grpc-js": "1.8.17",
```
cprice404, line 33: this looks fine but i'm wondering if for the grpc-js one, if we used one of those fuzzy match things like `^1.whatever`, if that would make it less likely to hit this in the future. I genuinely don't know; probably not, if the version gets locked in package-lock.json anyway.

not a blocker either way, just food for thought.

cprice404 (approved):

examples/nodejs/observability/package.json
pgautier404: Yeah, since we're shipping `package-lock` I also don't think that would help here. If this becomes a recurrent issue, we should think about maybe relaxing some of these selectors and removing the lockfiles for the examples.

PR #685, opened 2023-08-01
feat: add auth config to nextjs chat app example

anitarua:
Addresses [#385](https://github.com/momentohq/dev-eco-issue-tracker/issues/385) by providing a basic username/password auth option to the Next.js chat app example. Also removed the cache selector and updated the README. 

Currently, the user would have to specify the authentication method in both config.ts and .env.local because I couldn't find a way to get the client-side page.tsx to access the variables in config.ts (the compiler complained it couldn't find the `fs` module). Open to suggestions for making it a better user experience there, though.

## Screenshots

Currently, the flow for the nextjs app with the next-auth credentials provider requires the user to sign in with credentials matching what's found in the environment variables before allowing them to input a chat room name and username. Next-auth provides the login screen based on what providers are configured. 

Screenshot 2023-08-02 at 11 05 15 AM

Screenshot 2023-08-02 at 11 05 48 AM

Screenshot 2023-08-02 at 11 06 01 AM


examples/web/nextjs-chat/src/app/api/momento/token/route.ts
```diff
@@ -29,3 +36,28 @@ export async function GET(_request: Request) {
...
+
+async function fetchTokenWithAuthCredentials() {
+  const envUsername = String(process.env.MOMENTO_AUTH_USERNAME);
+  const envPassword = String(process.env.MOMENTO_AUTH_PASSWORD);
+
+  if (!envUsername || !envPassword) {
+    throw new Error("Username and/or password not detected in environment variables");
+  }
+
+  if (envUsername !== "momento" || envPassword !== "\$erverless") {
```
bruuuuuuuce, line 55: This is still essentially a public endpoint right? If a user deploys the next application with the correct username and password as part of the env vars, anyone can hit that endpoint and get a token? I think we would need the frontend to pass the username and password to the backend as part of the body/query params. Then inside this function here can verify that the username/password passed is correct

anitarua: Oh yeah, okay I'll implement that

bruuuuuuuce (dismissed):
:shipit:

anitarua:
Going to address Pete's comments before merging:

- would it be possible to remove the "Please sign in" screen and button and just send users directly to the login form if the auth method is credentials? 
- i also think for users who are coming to this app without instructions, the fact that you have to enter a chat room name is a little confusing. can we add just a little help text on that screen, something like "Please enter the name of the chat room you'd like to join. If it doesn't exist, it will be created." 
- it would also be cool if clicking the Leave button also cleared the user's credentials so they got bounced back to the login screen instead of the lobby, but i realize with multiple auth options that one might be trickier. if it's easy i think it'd be nice but if it isn't i really don't care that much :slightly_smiling_face:

anitarua:
Hey @pgautier404 I got the second and third items, but when I tried to set the login page as the first screen, it would only ever keep going back to the login screen. I added more explanations to show why each screen is present, though:

Screenshot 2023-08-02 at 3 17 20 PM

Screenshot 2023-08-02 at 3 17 38 PM

Screenshot 2023-08-02 at 3 17 52 PM

pgautier404 (approved):
Thank you!

bruuuuuuuce (approved):
:shipit:

PR #686, opened 2023-08-02
feat: introduce `CacheClient.cache` function to curry cache name across API calls

cprice404:
This commit introduces a new function, `CacheClient.cache`. If you call
it, you will get back an instance of a new type, `MomentoCache`. This
object has all of the same APIs that are available on the `CacheClient`,
but it curries the `cacheName` so that you no longer have to specify that
on every API call.


cprice404:
Going to try to add another commit to this to do the reflection-based unit test validation that the cache interface is in sync with the CacheClient, but this is ready for review.

anitarua (dismissed):
lgtm!

pgautier404 (dismissed):

allenheltondev (approved):

packages/core/test/unit/clients/imomentocache-interface.test.ts
```diff
@@ -0,0 +1,34 @@
+import {Project} from 'ts-morph';
+
+describe('IMomentoCache', () => {
+  it('has all of the data plane methods that exist on the ICacheClient interface', () => {
```
pratik151192, line 4: Nice, thank you!

cprice404: this test actually failed when I rebased because Rishti had added `dictionaryLength`. So it was a good call.

packages/core/src/clients/ICacheClient.ts
```diff
@@ -79,6 +80,8 @@ export type SortedSetIncrementOptions = CollectionCallOptions;
 export type SortedSetLengthByScoreOptions = SortedSetLengthByScoreCallOptions;
 
 export interface ICacheClient extends IControlClient, IPingClient {
+  cache(cacheName: string): IMomentoCache;
```
pratik151192, line 83: This was quite clever. At first I thought why we need to plumb this in `ICacheClient`. With this it's a single new line of code change to get our flavor; without this, we'd have needed changes in constructor initialization as well (Just adding my thought process here; no comment as such)

pratik151192 (approved):

PR #687, opened 2023-08-04
feat: determine vite app vended token scopes based on cognito user group"

anitarua:
Addresses the last comments in [#385](https://github.com/momentohq/dev-eco-issue-tracker/issues/385) for extending the cognito example to show two different levels of auth (e.g. subscribe-only vs. publish-and-subscribe).

This PR creates two Cognito users in the TVM CDK, creates two user pool groups, and adds one user to each group. The Vite app accepts two sets of username-password credentials in the environment variables and provides a dropdown menu for people to select which Cognito user they want to enter the chat app with. 

Also added to the Vite app README and added more comments to the chat app pages to make it consistent with the Next.js app.


examples/web/vite-chat-app/README.md
```diff
@@ -62,13 +62,17 @@ The username and password you provide here should match what was hardcoded into
 If you deployed your Token Vending Machine with the Amazon Cognito ("cognito") authentication method, you'll need to include these variables in your `.env.development` file:
```
cprice404, line 62: Can you add a few more words here that say something like:

"To demonstrate the Token Vending Machine's ability to vend different tokens to different types of users, the example Cognito authentication method provides two roles: a ReadOnly role that is only able to subscribe to the chat topic, and a ReadWrite role that can both publish and subscribe."

feel free to modify that as necessary to make it fit into your text above!

cprice404 (commented):
looks good! just a few small nits.

cprice404 (approved):
perfect

PR #688, opened 2023-08-05
feat: update grpc-js for keepalives and grpc-web for wasm compat

cprice404:
This commit updates to the latest version of our client-protos,
which includes updating to the latest grpc-js. This will pick up
the following changes from upstream:

* Implement channel idle timeout and the channel option grpc.client_idle_timeout_ms (#2471)
* Implement gRFC A62: pick_first: sticky TRANSIENT_FAILURE and address order randomization (#2511)
* Fix premature leaving of context due to improper Http2ServerCallStream handling (#2501 contributed by @CedricKassen)
* Add channel option grpc-node.tls_enable_trace to enable Node TLS tracing (#2507)
* Cancel deadline timer on server when call is cancelled (#2508)

* Added grpc.experimental.createResolver

* Fix propagation of UNIMPLEMENTED error messages (#2528)
* Fix a crash when the channel option grpc.keepalive_permit_without_calls is set (#2519)
* Update keepalive behavior to more correctly handle short calls and long periods of inactivity (#2513)
* Fix reporting of call stacks in unary request errors (#2503)
* Fix reporting of proxy info in channelz socket responses (#2503)

We also pick up the latest build of our grpc-web support, which uses
a more strict configuration for protoc generation. This should eliminate
some `eval` calls that were happening in google-protobuf, which should
in turn allow the web SDK to be used in a broader array of WASM environments,
such as CloudFlare.

### DIFF
--- a/packages/client-sdk-nodejs/package-lock.json
+++ b/packages/client-sdk-nodejs/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "@gomomento/sdk",
   "version": "0.0.1",
-  "lockfileVersion": 3,
+  "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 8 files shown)


krispraws (approved):
Changes look good. Can be merged after the Node18 tests pass. The failed auth test looked like a transient error related to timing. Thanks for getting this out!

PR #691, opened 2023-08-07
chore: move cloudflare workers http example to http directory

pratik151192:
No functional changes. Just moved all the cloudflare worker examples under ```http``` directory to have an additional ```web directory``` for web-sdk examples. The examples posted in this blog https://www.gomomento.com/blog/momento-cache-is-now-accessible-at-the-edge point to the examples/cloudflare-workers directory so that wouldn't be affected and customers can choose which sub-directory to navigate to.

I couldn't find a reference from public-dev-docs but please point me to one if it needs an update.


cprice404 (approved):
This is the right thing to do but we need a README in the parent dir now. Fine to do that in a separate PR.

PR #692, opened 2023-08-08
feat: add examples for web sdk and cloudflare workers

pratik151192:
- chore: move cloudflare workers http example to http directory
- fix: append http directory to readme commands
- feat: add examples for web sdk and cloudflare workers


examples/cloudflare-workers/web/README.md
```diff
@@ -0,0 +1,53 @@
+# Using Momento inside a Cloudflare Worker
```
cprice404, line 1: nit - I'd like for our users to be able to make a reasonable guess as to what the difference between the two examples is when they navigate to the cloudflare-directory. Maybe we should call the directories `http-api` and `web-sdk`.

examples/cloudflare-workers/web/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+This example uses
```
cprice404, line 3: We are going to want to update the title header on this page to clarify that this is different from the HTTP API example and probably link back to that one. We should also add some text here explaining what the difference is and what the tradeoffs are.

cprice404 (approved):
Code looks great. I left a few comments about directory structure and beefing up the README. In addition to those we also need to:

* add a README in the parent dir
* Update dev docs to reflect that we now have two examples

i'm okay with all of these things happening in follow-up PRs but we need to make sure we hit them before we close out the ticket.

PR #708, opened 2023-08-09
feat: add a configuration for lambda

nand4011:
Make the number of data clients a cache client creates configurable. 6 is likely too many for lambdas as the extra latency for the first 6 calls isn't worth it in a short-lived environment.

Create a new configuration, InRegion.Lambda, for use inside lambdas. It only creates 1 data client instead of 6.

### DIFF
--- a/packages/client-sdk-nodejs/src/config/configurations.ts
+++ b/packages/client-sdk-nodejs/src/config/configurations.ts
@@ -59,6 +59,7 @@ export class Laptop extends CacheConfiguration {
     const grpcConfig: GrpcConfiguration = new StaticGrpcConfiguration({
       deadlineMillis: deadlineMillis,
       maxSessionMemoryMb: defaultMaxSessionMemoryMb,
+      numClients: 6,
     });
     const transportStrategy: TransportStrategy = new StaticTransportStrategy({
       grpcConfiguration: grpcConfig,
@@ -99,6 +100,7 @@ class InRegionDefault extends CacheConfiguration {
     const grpcConfig: GrpcConfiguration = new StaticGrpcConfiguration({
       deadlineMillis: deadlineMillis,
       maxSessionMemoryMb: defaultMaxSessionMemoryMb,
+      numClients: 6,
     });
     const transportStrategy: TransportStrategy = new StaticTransportStrategy({
       grpcConfiguration: grpcConfig,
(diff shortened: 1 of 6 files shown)


packages/client-sdk-nodejs/src/config/configurations.ts
```diff
@@ -154,6 +157,47 @@ class InRegionLowLatency extends CacheConfiguration {
   }
 }
 
+class InRegionLambda extends CacheConfiguration {
```
cprice404, line 160: I don't think we need to call this `InRegion` (even though technically it almost certainly always will be).

Here's the prior art we have on naming choices for this:

https://github.com/momentohq/client-sdk-dotnet/blob/0022707888f8e288b352c90c07221d462131fa12/src/Momento.Sdk/Config/Configurations.cs#L172

packages/client-sdk-nodejs/src/config/configurations.ts
```diff
@@ -154,6 +157,47 @@ class InRegionLowLatency extends CacheConfiguration {
...
+  /**
+   * Provides v1 recommended configuration for an in-region lambda environment.  This configuration is guaranteed not
+   * to change in future releases of the Momento node.js SDK.
+   * @param {MomentoLoggerFactory} [loggerFactory=defaultLoggerFactory]
+   * @returns {CacheConfiguration}
+   */
+  static v1(
+    loggerFactory: MomentoLoggerFactory = defaultLoggerFactory
+  ): CacheConfiguration {
+    const deadlineMillis = 1100;
```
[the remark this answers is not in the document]
cprice404: definitely not LowLatency. If we were going to change it from `InRegion.Default` I think it would be to increase it, not decrease it, but my gut is to go with the same value we use for `InRegion.Default`.

@eaddingtonwhite lmk if you have any super strong opinions here.

packages/client-sdk-nodejs/src/config/transport/grpc-configuration.ts
```diff
@@ -9,6 +9,12 @@ export interface GrpcConfigurationProps {
    * more than this amount will return a ResourceExhausted error.
    */
   maxSessionMemoryMb: number;
+
+  /**
+   * The number of internal clients a cache client will create to communicate with Momento. More of them allows
+   * more concurrent requests, at the cost of more open connections and the latency of setting up each client.
+   */
+  numClients: number;
```
cprice404, line 17: This is technically a breaking change to public API. If we wanted to make this backward-compatible I think we would want to make it an optional field with a default value of 6.

I very much doubt that anyone is constructing any of these from scratch today so I think it's a safe change and probably worth it to keep things simple - but we really need to all start being very diligent about assessing changes like this for backward compat, and if you file a PR that is technically breaking backward compat, you need to clearly call it out in the PR and explain why you think it's the right thing to do so that others can weigh in.

packages/client-sdk-nodejs/test/unit/config/transport/transport-strategy.test.ts
```diff
@@ -39,9 +41,11 @@ describe('StaticGrpcConfiguration', () => {
 describe('StaticTransportStrategy', () => {
   const testDeadlineMillis = 90210;
   const testMaxSessionMemoryMb = 90211;
+  const testNumClients = 6;
```
cprice404, line 44: let's use a very unique value here so that we can be certain the test is not accidentally passing due to the test value matching some default value.

cprice404 (commented):
minor nits but looking very good, thanks for knocking this out!

packages/client-sdk-nodejs/src/config/transport/grpc-configuration.ts
nand4011: Good point. I wasn't thinking about people constructing the configuration from scratch.

packages/client-sdk-nodejs/src/config/configurations.ts
nand4011: I'm following the naming pattern of the other configurations in this class. We use InRegionDefault and InRegionLowLatency. The export class InRegion renames them so we can specify them as Configurations.InRegion.Default.

cprice404: what i'm saying is that in the .NET sdk, the `Lambda` configuration is entirely distinct from the `InRegion` ones.

packages/client-sdk-nodejs/src/config/transport/grpc-configuration.ts
nand4011: We should probably make it a default value. That will hide it better anyway. Technically it's still a breaking change since the interface has a new field, but I would question why anyone would make their own implementation of that.

cprice404 (approved):

PR #710, opened 2023-08-11
feat: add preview vector index client

pgautier404:
This commit adds a preview vector index client to the Web SDK. The client currently only implements data plane operations including create, list, and delete index.


examples/web/vector-index/doc-example-files/doc-examples-web-apis.ts
```diff
@@ -0,0 +1,36 @@
...
+    );
+  } else if (result instanceof ListIndexes.Error) {
+    throw new Error(`An error occurred while attempting to list caches: ${result.errorCode()}: ${result.toString()}`);
+  }
+}
+
+async function example_API_DeleteIndex(vectorClient: PreviewVectorClient) {
+  const result = await vectorClient.deleteIndex("test-index")
+  if (result instanceof DeleteIndex.Success) {
+    console.log("Index 'test-index' created");
```
anitarua, line 30: created -> deleted

pgautier404: Thank you! Fixed in e6d4e4a.

packages/client-sdk-web/src/vector-client-props.ts
```diff
@@ -0,0 +1,12 @@
+import {CredentialProvider, Configuration} from '.';
+
+export interface VectorClientProps {
+  /**
+   * Configuration settings for the vector client
+   */
+  configuration: Configuration;
```
cprice404, line 7: @pgautier404 I think this config object has a bunch of stuff like retries on it that we may want to handle differently for the Vector client. @malandis and I had this same discussion on his Python PR, and I think we are leaning towards having a separate Configuration object for Vector clients so that they can evolve independently.

We got burned a bit by copy/pasting too much of the config/wrapper stuff from CacheClient to TopicClient and AuthClient early on, and we kinda tried to unwind some of that. e.g. we have a separate config object for Topics now. so I want to make sure we're making a deliberate and forward-compatible choice here.

lmkwyt

packages/client-sdk-web/src/index.ts
```diff
@@ -119,6 +120,11 @@ import {
   TopicClientConfiguration,
 } from './config/topic-configuration';
 
+// VectorClient Response Types
+import * as CreateIndex from '@gomomento/sdk-core/dist/src/messages/responses/create-index';
+import * as ListIndexes from '@gomomento/sdk-core/dist/src/messages/responses/list-indexes';
+import * as DeleteIndex from '@gomomento/sdk-core/dist/src/messages/responses/delete-index';
```
cprice404, line 126: Michael I believe that you decided we wanted to start namespacing these a bit more in python. Do you have thoughts on what we should do here?

cprice404 (commented):
overall looking good. exciting that the tests pass. I had one question/nit about the Configuration object, and also would like Michael to weigh in with a quick API review, especially comparing/contrasting the response namespacing pattern that he did in python vs. what we should do here.

packages/client-sdk-web/src/vector-client-props.ts
pgautier404: That works for me-- I can imagine we might want to add distinct config for each type.

pgautier404: Added in 0cc099d.

packages/client-sdk-web/src/config/vector-configuration.ts
```diff
@@ -0,0 +1,48 @@
...
+   */
+  loggerFactory: MomentoLoggerFactory;
+
+  /**
+   * Configures low-level options for network interactions with the Momento service
+   */
+  transportStrategy: TransportStrategy;
+}
+
+export class VectorConfiguration implements Configuration {
```
cprice404, line 17: will we need these in nodejs SDK too? or does your PR not wire up node.js SDK yet?

pgautier404: I'm not doing anything on the node side in this PR. This is specifically for Web SDK and console.

cprice404: do you have a sense of how much extra work it would be? I presumed it would be pretty small, and that we would knock them both out together (in fact I'd kind of prefer that we establish a pattern of always adding stuff to both at the same time and not letting them drift from one another).

not a blocker for this PR but if it seems easy and you have bandwidth maybe you could file a follow-up PR before you split.

cprice404 (approved):
still would really like @malandis to skim this if possible but otherwise I'm +1

packages/core/src/messages/responses/list-indexes.ts
```diff
@@ -0,0 +1,66 @@
...
+    this.indexes = indexes;
+  }
+
+  /**
+   * An array of index names.
+   * @returns {string[]}
+   */
+  public getIndexes() {
+    return this.indexes;
+  }
```
malandis, line 38: To align with the protos

```suggestion
  public getIndexNames() {
    return this.indexNames;
  }
```

malandis (commented):
Looks good just one inconsistency vs the protos

cprice404:
@malandis what about the namespacing for the response objects? Do you have thoughts on how we should handle that here for symmetry with what you did in Python?

@pgautier404 you might want to look at the python PR to see what they are doing there. I think the goal is to stop dumping all the response classes into one big namespace and instead have a separate namespace for Vector vs Cache etc.

pgautier404:
> @malandis what about the namespacing for the response objects? Do you have thoughts on how we should handle that here for symmetry with what you did in Python?
> 
> @pgautier404 you might want to look at the python PR to see what they are doing there. I think the goal is to stop dumping all the response classes into one big namespace and instead have a separate namespace for Vector vs Cache etc.

In the Javascript SDKs, there is basically no namespacing whatsoever of classes we export. Every symbol that is exposed to developers is exposed through the root `index.ts` of each SDK. All clients, configs, responses, loggers, and so on are imported from the root. For example, the imports in `examples/nodejs/cache/advanced.ts` are:

```
import {
  CacheGet,
  ListCaches,
  CreateCache,
  CacheSet,
  CacheDelete,
  CacheClient,
  Configurations,
  MomentoLoggerFactory,
  DefaultMomentoLoggerFactory,
  DefaultMomentoLoggerLevel,
  ExperimentalMetricsCsvMiddleware,
  ExperimentalRequestLoggingMiddleware,
  CredentialProvider,
} from '@gomomento/sdk';
```

I don't really see how we could introduce a namespace for the vector index responses that wasn't inherently confusing to users who are importing every other class from the top level. Not to say there isn't a way to introduce namespaced responses in a UX-friendly way. I am not strong enough with TS any bright ideas here though.

cprice404:
discussed this with Michael; there are a few changes I want to get in here before this gets merged.

cprice404:
@malandis I'm done with this and ready to merge it if you are okay with my changes.

I tried a few things for namespacing the vector responses. I was able to separate them a little more cleanly in the file tree, and simplify the export statements in the index files, but I wasn't able to do what I really wanted to do, which was this:

https://github.com/momentohq/client-sdk-javascript/issues/784

I think that's too risky for a `1.x`. So my current thinking is:

* The control plane responses in this PR now have `Vector` in their names
* Users who want better namespacing can opt to import `vector` from the main index, and then reference these as `vector.*`, but
* The types are also available from the main index for consistency / simplicity.
* The data plane response types (coming in the next few PRs) will all be prefixed with `Vector*`, like the current Cache ones are prefixed with `Cache*`.

LMK if you are okay with that.

malandis:
That's fine with me. My only concern was some of the names can get unwieldly-long, eg `VectorAddItemBatchResponse`, `VectorAddItemBatch.Success`. If that's actually not bad then works for me.

cprice404:
> That's fine with me. My only concern was some of the names can get unwieldy/long, eg `VectorAddItemBatchResponse`, `VectorAddItemBatch.Success`. If that's actually not bad then works for me.

I don't see a better option at the moment and I'm hoping that with autocomplete those won't be too bad. I don't think they are meaningfully longer than the existing names for the Cache response classes.

malandis (approved):

PR #711, opened 2023-08-11
feat: Add eager connction client factory method.

nand4011:
Add a factory method that creates a CacheClient and attempts to eagerly establish all of its data client connections.

### DIFF
--- a/packages/client-sdk-nodejs/src/cache-client-props.ts
+++ b/packages/client-sdk-nodejs/src/cache-client-props.ts
@@ -16,6 +16,17 @@ export interface CacheClientProps {
   defaultTtlSeconds: number;
 }
 
+export interface EagerCacheClientProps extends CacheClientProps {
+  /**
+   * The time in seconds to wait for a client to establish a connection to Momento.
+   *
+   * If present, the client will eagerly create its connection to Momento at construction. It will wait until the
+   * connection is established, or until the timout runs out. It the timeout runs out, the client will be valid to use,
+   * but it may still be connecting in the background.
+   */
+  eagerConnectTimeout?: number;
+}
+
 /**
  * @deprecated use {CacheClientProps} instead
  */
--- a/packages/client-sdk-nodejs/src/cache-client.ts
+++ b/packages/client-sdk-nodejs/src/cache-client.ts
@@ -55,13 +55,16 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
    * Creates a new instance of CacheClient that eagerly creates its connections to Momento. By default, connections are
    * created lazily when the client is first used. If connections cannot be established, an error will be logged and
    * execution will resume.
+   * @param {CacheClientProps} props configuration and credentials for creating a CacheClient.
+   * @param {number} timeoutSeconds how long to wait for the connections.
    */
   static async createWithEagerConnections(
-    props: CacheClientProps
+    props: CacheClientProps,
+    timeoutSeconds = 10
   ): Promise<CacheClient> {
     const client = new CacheClient(props);
     await Promise.all(
-      client.dataClients.map(dc => (dc as DataClient).connect())
+      client.dataClients.map(dc => (dc as DataClient).connect(timeoutSeconds))
     );
     return client;
   }
--- a/packages/client-sdk-nodejs/src/cache-client.ts
+++ b/packages/client-sdk-nodejs/src/cache-client.ts
@@ -52,6 +52,21 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
   }
 
   /**
+   * Creates a new instance of CacheClient that eagerly creates its connections to Momento. By default, connections are
+   * created lazily when the client is first used. If connections cannot be established, an error will be logged and
+   * execution will resume.
+   */
+  static async createWithEagerConnections(
+    props: CacheClientProps
+  ): Promise<CacheClient> {
+    const client = new CacheClient(props);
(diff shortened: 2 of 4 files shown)


packages/client-sdk-nodejs/src/internal/data-client.ts
```diff
@@ -158,6 +159,69 @@ export class DataClient implements IDataClient {
       this.configuration.getMiddlewares()
     );
   }
+  public connect(): Promise<void> {
+    const deadline = new Date();
+    deadline.setSeconds(deadline.getSeconds() + 10);
```
pratik151192, line 164: Should this be more like 1 second since we know on average we took ~200ms?

packages/client-sdk-nodejs/src/internal/data-client.ts
```diff
@@ -158,6 +159,69 @@ export class DataClient implements IDataClient {
...
+            .getChannel()
+            .getConnectivityState(false);
+
+          if (newState === ConnectivityState.READY) {
+            this.logger.debug(`Connected! Current state: ${newState}`);
+            resolve();
+          } else if (newState === ConnectivityState.CONNECTING) {
+            // The connection goes through the CONNECTING state before becoming READY,
+            // so we must watch it twice.
+            this.connectWithinDeadline(deadline).then(resolve).catch(reject);
```
pratik151192, line 215: Does this warrant a delay before calling again? I believe there's a certain range (some 10s of millis) before the connection will transition to READY, and I'm not sure how often we may end up calling this recursively within that range. My concern is mostly with the call stack growing "too big", which ofcourse I don't know how big it can grow to be an issue in the first place.

packages/client-sdk-nodejs/test/unit/cache-client.test.ts
```diff
@@ -61,4 +62,18 @@ describe('CacheClient', () => {
...
     }
   });
+  it('createWithEagerConnections returns a client even if it cannot connect', async () => {
+    const momento = await CacheClient.createWithEagerConnections({
+      configuration: configuration,
+      credentialProvider: credentialProvider,
+      defaultTtlSeconds: 100,
+    });
+    const setResponse = await momento.set('cache', 'key', 'value');
+    expect(setResponse).toBeInstanceOf(CacheSet.Error);
```
pratik151192, line 72: I did not totally understand why this will not succeed

packages/client-sdk-nodejs/src/internal/data-client.ts
nand4011: I don't know what it would be in other environments since the customer was seeing 5 second latencies. I can make it configurable if that works better.

packages/client-sdk-nodejs/test/unit/cache-client.test.ts
nand4011: The token that this client uses is fake. It parses, so the client can be created, but it gets an error when it tries to connect.

packages/client-sdk-nodejs/src/internal/data-client.ts
nand4011: I've only seen the recursive call happen once. I think, since the watch only returns when a state change happens, it can only happen once.

packages/client-sdk-nodejs/src/internal/data-client.ts
nand4011: It has been made configurable.

packages/client-sdk-nodejs/test/unit/cache-client.test.ts
cprice404: does it make sense for this to be a unit test then? Since it is attempting a network call? should it be an int test, and would it be worth adding a happy path int test or two?

These don't have to be addressed in this PR, can be a follow-up; for this PR the most important thing is that we align on the new API.

packages/client-sdk-nodejs/src/cache-client.ts
```diff
@@ -51,6 +51,24 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
...
   }
 
+  /**
+   * Creates a new instance of CacheClient that eagerly creates its connections to Momento. By default, connections are
+   * created lazily when the client is first used. If connections cannot be established, an error will be logged and
+   * execution will resume.
+   * @param {CacheClientProps} props configuration and credentials for creating a CacheClient.
+   * @param {number} timeoutSeconds how long to wait for the connections.
+   */
+  static async createWithEagerConnections(
```
cprice404, line 61: maybe just name this singular since we hope that most users don't have to think about how many connections we have? I don't feel too strongly but I think that might fit a little better with our existing config facade.

cprice404: also we should add a plain `create` factory function to correspond to this one. That way we can start to recommend the factory functions for both use cases, rather than having one workflow require a `new` and the other one not.

packages/client-sdk-nodejs/src/cache-client-props.ts
```diff
@@ -16,6 +16,17 @@ export interface CacheClientProps {
   defaultTtlSeconds: number;
 }
 
+export interface EagerCacheClientProps extends CacheClientProps {
```
cprice404, line 19: you'll need to add this to the exports in the index file.

cprice404 (commented):
one tiny nit

cprice404 (approved):

PR #714, opened 2023-08-12
fix: correct log messages when instantiating TopicClient

pgautier404:
This commit corrects a log message that caused me great confusion.


bruuuuuuuce (approved):
:shipit:

PR #725, opened 2023-08-15
chore: tweaks to cloudflare Web SDK example docs, remove createCache call

cprice404:
Most tokens that users will have access to in a CloudFlare environment probably
won't have permissions to create a cache. Therefore this example removes the
call to `CreateCache` from the Web SDK example.

Also:
* Adds more logging to make it easier to debug in a deployed environment
* Minor tweaks to the organization of the README to prepare for alternate ways
  to set up auth in the near future.

### DIFF
--- a/examples/cloudflare-workers/README.md
+++ b/examples/cloudflare-workers/README.md
@@ -17,15 +17,7 @@
 
 This directory contains subdirectories for two types of example projects to use Momento inside a Cloudflare worker:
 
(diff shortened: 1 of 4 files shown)


pratik151192:
Thanks for doing this. Had updated docs to use fgac but never realized the code was creating a cache!

pratik151192 (approved):

PR #726, opened 2023-08-15
feat: add dictionary length api

rishtigupta:
## PR Description:
- Add dictionary length api to nodejs and web-sdk
- Add integ tests

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/data-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/data-client.ts
@@ -16,6 +16,7 @@ import {
   CacheDictionaryRemoveFields,
   CacheDictionarySetField,
   CacheDictionarySetFields,
+  CacheDictionaryLength,
   CacheGet,
   CacheIncrement,
   CacheListConcatenateBack,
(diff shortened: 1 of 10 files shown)


packages/client-sdk-nodejs/src/internal/data-client.ts
```diff
@@ -1750,6 +1751,64 @@ export class DataClient implements IDataClient {
...
+        request,
+        metadata,
+        {
+          interceptors: this.interceptors,
+        },
+        (err, resp) => {
+          if (resp?.missing) {
+            resolve(new CacheDictionaryLength.Miss());
+          } else if (resp?.found) {
+            if (!resp.found.length) {
```
pratik151192, line 1797: Is this even possible? This seems like a server side bug to me

packages/common-integration-tests/src/dictionary.ts
```diff
@@ -1466,5 +1467,45 @@ export function runDictionaryTests(
         }
       });
     });
+
+    describe('#dictionaryLength', () => {
+      itBehavesLikeItValidates((props: ValidateDictionaryProps) => {
+        return Momento.dictionaryLength(props.cacheName, props.dictionaryName);
+      });
+
+      it('returns a miss if the dictionary does not exist', async () => {
```
pratik151192, line 1476: Do you know of a reproducible case to hit the code path where the length doesn't exist even if there was a hit?

rishtigupta: Nope, afaik dictionary cannot exist with no values in it.

pratik151192: Yeah so that's the reason I asked the question above. If it's not possible for the dictionary to exist without values, is it still possible for the `length` field to be missing in the API response?

packages/client-sdk-nodejs/src/internal/data-client.ts
cprice404: @rishtigupta what is the type of `resp.found.length` here? Is it just `number`, or is it nullable (`number?`)?

cprice404 (commented):
one clarifying question related to one of @pratik151192 's comments, but otherwise LGTM! thanks for doing this!

rishtigupta:
@cprice404 the return type of `resp.found.length` is number. 
listLength implementation mentioned this - https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/internal/data-client.ts#L991, which I thought might be true for dictionary as well.

cprice404:
> @cprice404 the return type of `resp.found.length` is number. listLength implementation mentioned this - https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/internal/data-client.ts#L991, which I thought might be true for dictionary as well.

ok. if the type is `number`, then `!number` basically means `number !== 0`.

That comment seems sketchy to me. If your int tests work without this check I would remove it. (And honestly I wouldn't mind double-checking the listLength one that you linked above, but I know you have console work to do so you don't need to spend time on that for now.)

PR #728, opened 2023-08-15
feat: fastly compute integration

anitarua:
Addresses [#389](https://github.com/momentohq/dev-eco-issue-tracker/issues/389).

After much trial and error, here is a code example for using the Momento HTTP API inside of a Fastly Compute@Edge instance. The example sets, gets, and deletes an item from an existing cache. The `fastly.toml` configs handle most of the Fastly resources setup. 

The deployed instance is running at [https://solely-civil-crab.edgecompute.app](https://solely-civil-crab.edgecompute.app/).


examples/fastly-compute/.gitignore
```diff
@@ -0,0 +1,5 @@
+/node_modules
+/bin
+/build
+/pkg
+secrets.json
```
cprice404, line 5: nicely done

cprice404 (approved):
awesome

cprice404:
This is great @anitarua , thank you! Before we close this ticket out, I think we should add something to the dev docs. Check out what Pete and Pratik have put there for cloudflare/netlify etc. and lmk if you are up for filing a PR over there!

anitarua:
Huge thanks to @krispraws for helping to get this integration up and running as well!!

And yes, for sure, I'd be happy to follow up with a dev docs PR!

krispraws (commented):
@anitarua, I just saw this so sorry for the delayed comment. We should be using a [secret store](https://developer.fastly.com/reference/api/services/resources/secret-store/) for the momento token instead of a config store. I didn't use it in my [original example](https://github.com/momentohq/client-sdk-javascript/compare/main...krispraws:client-sdk-javascript:fastly_http_example) because I was just trying to get it working. But if we are publishing this as an example, we should follow good security practices.

anitarua:
Oh okay, I will put up a follow up PR once I get that working then!

krispraws:
@anitarua pointed out that secret stores are a beta feature. Let's just add a comment to the code and/or docs saying that tokens should be stored in secret stores but we are using config store because it is a beta feature and not available to all accounts

PR #730, opened 2023-08-16
chore: update fastly readme to explain secret store vs config store usage

anitarua:
Follow up to #728 to emphasize that our example uses Config Stores because Secret Stores are currently restricted to Fastly beta users, but in production environments, Secret Stores should be used to hold auth tokens instead Config Stores.


krispraws (approved):
Thank you! 🙇‍♀️

PR #735, opened 2023-08-17
chore: change examples to use cache client factory signature

pratik151192:
- feat: add a script and ci workflow to build all examples
- chore: remvoe examples build from cis for now
- chore: change all examples to use cache client factory signature


pratik151192:
this has some errors w.r.t CI - debugging step by step

PR #736, opened 2023-08-17
feat: add examples for using momento in Deno

anitarua:
Addresses [#417](https://github.com/momentohq/dev-eco-issue-tracker/issues/417). Created a new `examples` directory for using Momento Cache in Deno: one using the web SDK and one using the HTTP API.


pratik151192:
LGTM! Great work! 

I'd just add one last message after deployment succeeds to click on the URL that starts with the "project-name" which can also be easily shared with others to show the project they built. We also added our own URLs on other integrations if you want to link that part as well https://good-frog-73.deno.dev/

cprice404 (commented):
This is looking great, thanks!

Can you please add a README in the parent dir, kind of like this:

https://github.com/momentohq/client-sdk-javascript/tree/main/examples/cloudflare-workers

And then take a sentence or two about why a user would pick one of the SDKs over the other and include it in the child dir READMEs. (I think the cloudflare one probably has examples of that too.)

Other than that I think this is good to go!

cprice404 (commented):
oh! one more thing... is there anything worthwhile we can do in CI to get some coverage of these? and dependabot, idk if it has deno.json support?

anitarua:
Looks like deno dependabot support is still an [open issue](https://github.com/dependabot/dependabot-core/issues/2417), but I think I can [add deno to the CI build](https://deno.land/manual@v1.36.1/advanced/continuous_integration)

cprice404 (approved):
dance

PR #733, opened 2023-08-17
feat: add a script and ci workflow to build all examples

pratik151192:
Added a script that builds all packages. ```tsc``` is now working for every example; ```eslint``` hasn't been wired for examples that had them but were never exercised and were failing. We can double click on this later but ```npm run build``` was my min requirement. The CI workflow wasn't as straight forward as I thought it'd be for this so they have been left untouched and will be fast followed. I needed this for now as I was changing signature of client construction across almost all the examples.

The script prints all the directories that were built.. 

```
Directories Built:
/Users/pratik/sandbox/client-sdk-javascript/examples/cloudflare-workers/http-api
/Users/pratik/sandbox/client-sdk-javascript/examples/cloudflare-workers/web-sdk
/Users/pratik/sandbox/client-sdk-javascript/examples/fastly-compute
/Users/pratik/sandbox/client-sdk-javascript/examples/nodejs/access-control
/Users/pratik/sandbox/client-sdk-javascript/examples/nodejs/aws
/Users/pratik/sandbox/client-sdk-javascript/examples/nodejs/cache
/Users/pratik/sandbox/client-sdk-javascript/examples/nodejs/lambda-examples/simple-get/lambda/simple-get
/Users/pratik/sandbox/client-sdk-javascript/examples/nodejs/lambda-examples/topics-microservice
/Users/pratik/sandbox/client-sdk-javascript/examples/nodejs/load-gen
/Users/pratik/sandbox/client-sdk-javascript/examples/nodejs/mongodb-examples/simple-read-aside
/Users/pratik/sandbox/client-sdk-javascript/examples/nodejs/observability
/Users/pratik/sandbox/client-sdk-javascript/examples/nodejs/token-vending-machine/lambda/authorizer
/Users/pratik/sandbox/client-sdk-javascript/examples/nodejs/token-vending-machine/lambda/token-vending-machine
/Users/pratik/sandbox/client-sdk-javascript/examples/nodejs/topics
/Users/pratik/sandbox/client-sdk-javascript/examples/web/cache
/Users/pratik/sandbox/client-sdk-javascript/examples/web/nextjs-chat
/Users/pratik/sandbox/client-sdk-javascript/examples/web/vite-chat-app
```

PR #739, opened 2023-08-17
feat: update all examples to use the cacheclient factory

pratik151192:
Bumped all examples to `1.31.0` and updated them to use the cache client factory instead of the constructor. Only the `observability` directory wasn't updated as it has an orthogonal blocker to bump up the SDK release version being tracked under https://github.com/momentohq/dev-eco-issue-tracker/issues/435.

### Testing 

- Ran `./scripts/build-all-examples.sh` and it succeeded
    - This was added here https://github.com/momentohq/client-sdk-javascript/pull/733
- I also tested the `public-dev-doc` repo by changing the SDK branch to the one on this PR and I saw the `cheat-sheet` and `develop` pages updated with the new signatures.

### DIFF
--- a/examples/nodejs/lambda-examples/simple-get/lambda/simple-get/handler.ts
+++ b/examples/nodejs/lambda-examples/simple-get/lambda/simple-get/handler.ts
@@ -52,7 +52,8 @@ async function getCacheClient(): Promise<CacheClient> {
   if (_cacheClient === undefined) {
     const momentoAuthToken = await getSecret(authTokenSecretName);
     console.log('Retrieved secret!');
-    _cacheClient = new CacheClient({
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call,@typescript-eslint/no-unsafe-assignment
+    _cacheClient = await CacheClient.create({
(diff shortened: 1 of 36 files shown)


cprice404 (commented):
changes look fine but I think we need a 1.32.0 before we can really complete this task?

examples/nodejs/lambda-examples/simple-get/lambda/simple-get/handler.ts
```diff
@@ -52,7 +52,8 @@ async function getCacheClient(): Promise<CacheClient> {
   if (_cacheClient === undefined) {
     const momentoAuthToken = await getSecret(authTokenSecretName);
     console.log('Retrieved secret!');
-    _cacheClient = new CacheClient({
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call,@typescript-eslint/no-unsafe-assignment
+    _cacheClient = await CacheClient.create({
       configuration: Configurations.InRegion.Default.v1().addMiddleware(
```
cprice404, line 57: This should be using the lambda config

cprice404 (approved):

PR #741, opened 2023-08-18
feat: bump protos to support list caches with metadata around limits

bruuuuuuuce:
protos now contains cache limits around 

```
  // The amount of transactions per second that can be exercised
  uint32 max_traffic_rate = 1; 
  // The amount of traffic per second that can be exercised in KiB
  uint32 max_throughput_kbps = 2;
  // The maximum size of a single item in KiB
  uint32 max_item_size_kb = 3;
  // The maximum TTL allowed for a single item, in seconds
  uint64 max_ttl_seconds = 4;
```

and topic limits around

```
  // The amount of messages that can be published per second
  uint32 max_publish_rate = 1;
  // The maximum amount of active subscriptions per cache
  uint32 max_subscription_count = 2;
  // The maximum size of a single publish message, in KiB
  uint32 max_publish_message_size_kb = 3;
```

unblocks ui for https://github.com/momentohq/momento-console/issues/293

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/control-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/control-client.ts
@@ -176,15 +176,16 @@ export class ControlClient {
               const cacheName = cache.cache_name;
               const topicLimits = new TopicLimits({
                 maxPublishMessageSizeKb:
-                  cache.topic_limits.max_publish_message_size_kb,
-                maxSubscriptionCount: cache.topic_limits.max_subscription_count,
-                maxPublishRate: cache.topic_limits.max_publish_rate,
+                  cache.topic_limits?.max_publish_message_size_kb || 0,
+                maxSubscriptionCount:
+                  cache.topic_limits?.max_subscription_count || 0,
+                maxPublishRate: cache.topic_limits?.max_publish_rate || 0,
               });
               const cacheLimits = new CacheLimits({
-                maxTtlSeconds: cache.cache_limits.max_ttl_seconds,
-                maxItemSizeKb: cache.cache_limits.max_item_size_kb,
-                maxThroughputKbps: cache.cache_limits.max_throughput_kbps,
-                maxTrafficRate: cache.cache_limits.max_traffic_rate,
+                maxTtlSeconds: cache.cache_limits?.max_ttl_seconds || 0,
+                maxItemSizeKb: cache.cache_limits?.max_item_size_kb || 0,
+                maxThroughputKbps: cache.cache_limits?.max_throughput_kbps || 0,
+                maxTrafficRate: cache.cache_limits?.max_traffic_rate || 0,
               });
               return new CacheInfo(cacheName, topicLimits, cacheLimits);
             });
--- a/packages/client-sdk-nodejs/src/internal/control-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/control-client.ts
@@ -27,6 +27,10 @@ import {
 } from '@gomomento/sdk-core/dist/src/internal/utils';
 import {normalizeSdkError} from '@gomomento/sdk-core/dist/src/errors';
 import {_SigningKey} from '@gomomento/sdk-core/dist/src/messages/responses/grpc-response-types';
+import {
+  CacheLimits,
+  TopicLimits,
+} from '@gomomento/sdk-core/dist/src/messages/cache-info';
 
 export interface ControlClientProps {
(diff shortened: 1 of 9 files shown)


pratik151192:
The code LGTM but there aren't tests. Looks like we haven't implemented it on the backend yet? I think even if we haven't, we could just wire the tests to assert on 0, and when we do implement it, we can just change them?

rishtigupta (dismissed):

packages/core/src/messages/cache-info.ts
```diff
@@ -1,11 +1,94 @@
+type CacheLimitsProps = {
```
cprice404, line 1: what's the difference between `type` and `interface` here?

packages/core/src/messages/cache-info.ts
```diff
@@ -1,11 +1,94 @@
...
+  maxTrafficRate: number;
+  maxThroughputKbps: number;
+  maxItemSizeKb: number;
+  maxTtlSeconds: number;
+};
+
+export class CacheLimits {
```
cprice404, line 8: nit: I'm guessing you did these as classes because the main CacheInfo object is a cache; but is there any reason to do these as classes instead of interfaces?

cprice404 (commented):
couple of nitpicks/questions but nothing big

packages/client-sdk-nodejs/src/internal/control-client.ts
```diff
@@ -168,9 +172,23 @@ export class ControlClient {
...
+                maxSubscriptionCount:
+                  cache.topic_limits?.max_subscription_count || 0,
+                maxPublishRate: cache.topic_limits?.max_publish_rate || 0,
+              });
+              const cacheLimits = new CacheLimits({
+                maxTtlSeconds: cache.cache_limits?.max_ttl_seconds || 0,
+                maxItemSizeKb: cache.cache_limits?.max_item_size_kb || 0,
+                maxThroughputKbps: cache.cache_limits?.max_throughput_kbps || 0,
+                maxTrafficRate: cache.cache_limits?.max_traffic_rate || 0,
+              });
```
cprice404, line 189: just checking my understanding of the proto changes - `cache.cache_name` is always present but the `cache_limits` and `topics_limits` were modeled as optional?

bruuuuuuuce: they are modeled as always present, but until the change gets all the way through the backend, they will be returned as null

cprice404: but I'm surprised the TS compiler will accept this? If the type is defined as non-null?

cprice404: okay. interesting.

worth capturing a ticket to come back and remove the `?` later?

bruuuuuuuce: yea, the other weird thing about this is that the grpc web generated code compiles in ts to show that its optional. The js generated code says its required

bruuuuuuuce: so the grpc web code will always need a `?`

cprice404: why is it different for `cache_name` though?

bruuuuuuuce: I believe because it is a grpc object vs a string, but I am not entirely sure

cprice404: okay. it's probably fine like this. thanks.

bruuuuuuuce: ill make a ticket to remove the `?` for the nodejs sdk tho

bruuuuuuuce: https://github.com/momentohq/client-sdk-javascript/issues/755

packages/core/src/messages/cache-info.ts
bruuuuuuuce: yea I think an interface should work here as well

cprice404: I feel like that might be a little simpler/cleaner, wdyt?

I know it's not a huge deal either way but since this is technically going to be public API, just wanted to make sure our decision was deliberate.

cprice404 (approved):

PR #742, opened 2023-08-18
feat: Don't leak grpc-js/protobuf objects through middleware API

cprice404:
This commit changes the middleware API signatures so that they
no longer directly reference objects from the grpc-js and protobuf
libraries.

Prior to this change, any project that included a custom middleware
implementation would need to be very careful to ensure that its
grpc-js and protobuf dependency versions exactly matched those
of the Momento SDK. And when the Momento SDK updates to new versions
of those deps, consuming projects would be forced to update accordingly.
Failure to do so would result in cryptic errors about object types
not being assignable.

Now we provide our own Middleware* classes that decorate the underlying
protobuf/grpc-js libs. This means that consuming projects with custom
middleware will only depend on our API, and our transitive deps will
be hidden from them.

### DIFF
--- a/packages/client-sdk-nodejs/src/config/middleware/example-async-middleware.ts
+++ b/packages/client-sdk-nodejs/src/config/middleware/example-async-middleware.ts
@@ -1,7 +1,11 @@
-import {Middleware, MiddlewareRequestHandler} from './middleware';
+import {
+  Middleware,
+  MiddlewareMessage,
+  MiddlewareMetadata,
+  MiddlewareRequestHandler,
+  MiddlewareStatus,
+} from './middleware';
 
-import {Metadata, StatusObject} from '@grpc/grpc-js';
-import {Message} from 'google-protobuf';
 import {MomentoLogger, MomentoLoggerFactory} from '../..';
 
 class ExampleAsyncMiddlewareRequestHandler implements MiddlewareRequestHandler {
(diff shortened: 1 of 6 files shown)


pgautier404 (commented):
Nice! LGTM pending resolution of @pratik151192's question.

pgautier404 (dismissed):

cprice404:
> Nice! Can you point the `observability` package.json to your local and check if `tsc` works successfully for `ExampleMetricsMiddleware`? For me it was failing as soon as I tried to bump it up from `1.29.0` so your local should serve as the test if we'd be able to bump it after this PR is merged

sounds good, will do

cprice404:
@pratik151192 @pgautier404 I added a commit, https://github.com/momentohq/client-sdk-javascript/pull/742/commits/36723f2683a65343bf6bcaabae984d8871eb8c30 , which makes a minor tweak to the previous version such that we only instantiate the decorator classes once per event. (Previously we were instantiating them once per middleware, and continually converting back and forth between the gRPC version and the decorated version, so this way should reduce the number of allocations if there are a multiple middlewares registered.)

I also tested this with both the loadgen example (with metrics logging middleware enabled), and with the observability example. Seemed to work fine with both. (Observability example requires some minor code changes to use the new types.

pratik151192:
Thanks @cprice404. Regarding 

> Observability example requires some minor code changes to use the new types

These were probably skipped from the commit accidentally.

pratik151192 (approved):

pratik151192:
Ah nevermind @cprice404 you can only update the examples after a new release and then make those changes

pratik151192:
🛥️🚀

PR #744, opened 2023-08-18
feat: add event loop lag to metrics middleware

nand4011:
Add the event loop delay at the start of a Momento call to the metrics middleware. It measures this by calling setTimeout with no delay that executes a function that subtracts the start time of the call from the current time. If there is no lag, it executes immediately and there is no difference. I didn't use the node performance measurement APIs because its event loop delay is a histogram that measures on a regular interval, and that doesn't fit well with per-request metrics.

### DIFF
--- a/packages/client-sdk-nodejs/src/config/middleware/experimental-metrics-csv-middleware.ts
+++ b/packages/client-sdk-nodejs/src/config/middleware/experimental-metrics-csv-middleware.ts
@@ -33,6 +33,7 @@ class ExperimentalMetricsCsvMiddlewareRequestHandler extends ExperimentalMetrics
       metrics.requestSize,
       metrics.responseSize,
       metrics.connectionID,
+      metrics.eventLoopDelay,
     ].join(',');
     try {
       await fs.promises.appendFile(this.csvPath, `${csvRow}\n`);
--- a/packages/client-sdk-nodejs/src/config/middleware/impl/experimental-metrics-middleware.ts
+++ b/packages/client-sdk-nodejs/src/config/middleware/impl/experimental-metrics-middleware.ts
@@ -20,6 +20,7 @@ const FIELD_NAMES: Array<string> = [
   'requestSize',
   'responseSize',
   'connectionID',
+  'eventLoopDelay',
 ];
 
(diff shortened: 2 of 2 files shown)


nand4011:
I spent way too long thinking about how to make the histogram work before switching to the current strategy. It may still be useful to add mean or p99 event loop lag as measured by the histogram since the start of the program, or event loop utilization since the start. I can add those in if desired.

cprice404:
@nand4011 This is an interesting idea!

My first thought when seeing it though is: I'm not sure we want to add another event to the event loop by calling the setTimeout, it might compound the problem? Were you able to use this in a program and have it show you interesting results?

I was thinking that with the performance measurement API, we wouldn't necessarily have to emit metrics that were per-request. We could either:

(a) emit a separate set of metrics on a different cadence, that just told us things about the event loop. As long as there are timestamps we can correlate with the per-request metrics then it still serves our purpose. or
(b) we could include a few fields with the latest percentile metrics from the histogram in each request metrics log. those metrics wouldn't correlate with that specific request, and there'd be a bunch of requests that had the same values until the next time the histogram was updated, but it'd still at least give us a correlation with the load of the system.

I'd be curious to hear your thoughts on tradeoffs between these approaches vs what you have in this PR. We can discuss it on a call Monday if that's easier though!

nand4011:
I think that using a timer to determine lag is the way it is done internally in the performance observer: https://nodejs.org/api/perf_hooks.html#perf_hooksmonitoreventloopdelayoptions "Using a timer to detect approximate event loop delay works because the execution of timers is tied specifically to the lifecycle of the libuv event loop. That is, a delay in the loop will cause a delay in the execution of the timer, and those delays are specifically what this API is intended to detect."
The difference is that the performance observer will make a new timer every 20 ms or whatever interval we choose instead of at the start of each call.
I did see ~120ms of lag in the first call on a new lambda, so it does show something. The intermittent higher latency calls aren't always associated with a noticeable lag, however.

I have no problem with introducing outputting metrics on an interval. I avoided it because the middleware is specific to per-call metrics at the moment.

nand4011:
We're going to switch to logging event loop lag on a regular cadence instead of per request.

PR #760, opened 2023-08-22
feat: add value accessors on all relevant response classes

cprice404:
This commit includes the following two major changes:
* All relevant `Hit`/`Success` classes now have a `value()` function,
  which returns the data type that we expect to be the most commonly
  used one for the response.
* All relevant `Response` classes now have a `value()` function that
  returns the `Hit`/`Success` subclass's `value()` for the happy path,
  or `undefined` for the `Miss`/`Error` path.

This should simplify a lot of example code, so that we don't have to
call `valueRecord` or similar explicitly specified accessors, and we
can now access `value()` without doing an `instanceof` check first
(even in TypeScript).

NOTE: it is crucially important that we review the choice of data
types for each `value()` function closely and make sure that we
are all aligned on those choices before this is merged.

### DIFF
--- a/packages/core/src/messages/responses/cache-dictionary-get-field.ts
+++ b/packages/core/src/messages/responses/cache-dictionary-get-field.ts
@@ -29,7 +29,14 @@ const TEXT_DECODER = new TextDecoder();
  * }
  * ```
  */
-export abstract class Response extends ResponseBase {}
+export abstract class Response extends ResponseBase {
+  public value(): string | undefined {
+    if (this instanceof Hit) {
+      return (this as Hit).value();
+    }
+    return undefined;
+  }
+}
 
 class _Hit extends Response {
   private readonly body: Uint8Array;
@@ -37,6 +44,15 @@ class _Hit extends Response {
     super();
     this.body = body;
   }
+
+  /**
+   * Returns the data as a utf-8 string, decoded from the underlying byte array.
+   * @returns string
+   */
+  public value(): string {
+    return TEXT_DECODER.decode(this.body);
+  }
+
   /**
    * Returns the data as a utf-8 string, decoded from the underlying byte array.
    * @returns string
--- a/packages/core/src/messages/responses/cache-dictionary-get-field.ts
+++ b/packages/core/src/messages/responses/cache-dictionary-get-field.ts
@@ -50,7 +50,7 @@ class _Hit extends Response {
    * @returns string
    */
   public value(): string {
-    return TEXT_DECODER.decode(this.body);
+    return this.valueString();
   }
 
   /**
--- a/packages/core/src/messages/responses/cache-dictionary-get-fields.ts
+++ b/packages/core/src/messages/responses/cache-dictionary-get-fields.ts
@@ -34,7 +34,14 @@ type CacheDictionaryGetFieldResponseType =
  * }
  * ```
  */
-export abstract class Response extends ResponseBase {}
+export abstract class Response extends ResponseBase {
+  public value(): Record<string, string> | undefined {
+    if (this instanceof Hit) {
+      return (this as Hit).value();
+    }
+    return undefined;
+  }
+}
 
 class _Hit extends Response {
   private readonly items: _DictionaryGetResponsePart[];
@@ -141,6 +148,16 @@ class _Hit extends Response {
    * @returns {Record<string, string>}
    */
   public valueRecord(): Record<string, string> {
+    return this.valueRecordStringString();
+  }
+
+  /**
+   * Returns the data as a Record whose keys and values are utf-8 strings, decoded from the underlying byte arrays.
+   * This can be used in most places where an Object is desired.  This is a convenience alias for
+   * {valueRecordStringString}.
+   * @returns {Record<string, string>}
+   */
+  public value(): Record<string, string> {
     return this.valueRecordStringString();
   }
 
--- a/packages/core/src/messages/responses/cache-dictionary-increment.ts
+++ b/packages/core/src/messages/responses/cache-dictionary-increment.ts
@@ -20,22 +20,32 @@ import {ResponseBase, ResponseError, ResponseSuccess} from './response-base';
  * }
  * ```
  */
-export abstract class Response extends ResponseBase {}
+export abstract class Response extends ResponseBase {
+  public value(): number | undefined {
+    if (this instanceof Success) {
+      return (this as Success).value();
+    }
+    return undefined;
+  }
+}
 
 class _Success extends Response {
-  private readonly value: number;
+  private readonly _value: number;
 
   constructor(value: number) {
     super();
-    this.value = value;
+    this._value = value;
   }
 
   /**
    * The new value of the element after incrementing.
    * @returns {number}
    */
+  public value(): number {
+    return this._value;
+  }
   public valueNumber(): number {
-    return this.value;
+    return this._value;
   }
 
   public override toString(): string {
--- a/packages/core/src/messages/responses/cache-get.ts
+++ b/packages/core/src/messages/responses/cache-get.ts
@@ -6,6 +6,7 @@ import {
   ResponseMiss,
 } from './response-base';
 import {truncateString} from '../../internal/utils';
+import {CacheGet} from '../../index';
 
 const TEXT_DECODER = new TextDecoder();
 
@@ -29,7 +30,14 @@ const TEXT_DECODER = new TextDecoder();
  * }
  * ```
  */
-export abstract class Response extends ResponseBase {}
+export abstract class Response extends ResponseBase {
+  public value(): string | undefined {
+    if (this instanceof CacheGet.Hit) {
+      return (this as CacheGet.Hit).value();
+    }
+    return undefined;
+  }
+}
 
 class _Hit extends Response {
   private readonly body: Uint8Array;
@@ -37,6 +45,15 @@ class _Hit extends Response {
     super();
     this.body = body;
   }
+
+  /**
+   * Returns the data as a utf-8 string, decoded from the underlying byte array.
+   * @returns string
+   */
+  public value(): string {
+    return TEXT_DECODER.decode(this.body);
+  }
+
   /**
    * Returns the data as a utf-8 string, decoded from the underlying byte array.
    * @returns string
--- a/packages/core/src/messages/responses/cache-get.ts
+++ b/packages/core/src/messages/responses/cache-get.ts
@@ -6,7 +6,6 @@ import {
   ResponseMiss,
 } from './response-base';
 import {truncateString} from '../../internal/utils';
-import {CacheGet} from '../../index';
 
 const TEXT_DECODER = new TextDecoder();
 
@@ -32,8 +31,8 @@ const TEXT_DECODER = new TextDecoder();
  */
 export abstract class Response extends ResponseBase {
   public value(): string | undefined {
-    if (this instanceof CacheGet.Hit) {
-      return (this as CacheGet.Hit).value();
+    if (this instanceof Hit) {
+      return (this as Hit).value();
     }
     return undefined;
   }
--- a/packages/core/src/messages/responses/cache-increment.ts
+++ b/packages/core/src/messages/responses/cache-increment.ts
@@ -20,22 +20,32 @@ import {ResponseBase, ResponseError, ResponseSuccess} from './response-base';
  * }
  * ```
  */
-export abstract class Response extends ResponseBase {}
+export abstract class Response extends ResponseBase {
+  public value(): number | undefined {
+    if (this instanceof Success) {
+      return (this as Success).value();
+    }
+    return undefined;
+  }
+}
 
 class _Success extends Response {
-  private readonly value: number;
+  private readonly _value: number;
 
   constructor(value: number) {
     super();
-    this.value = value;
+    this._value = value;
   }
 
   /**
    * The new value of the element after incrementing.
    * @returns {number}
    */
+  public value(): number {
+    return this._value;
+  }
   public valueNumber(): number {
-    return this.value;
+    return this._value;
   }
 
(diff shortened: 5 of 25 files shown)


packages/core/src/messages/responses/cache-set-fetch.ts
```diff
@@ -64,6 +71,16 @@ class _Hit extends Response {
...
 
+  /**
+   * Returns the data as an Array whose values are utf-8 strings, decoded from the underlying byte arrays.
+   * This accessor is provided because Arrays are sometimes easier to work with in TypeScript/JavaScript than Sets are.
+   * This is a convenience alias for {valueArrayString}.
+   * @returns {string[]}
+   */
+  public value(): string[] {
+    return this.valueArrayString();
+  }
```
cprice404, line 82: @allenheltondev this is probably the one that I was least certain which data type we should return for the new `value()` function. `Set` might be a more obvious choice here but I have always found `Set`s to be kind of a pain to work with in JS so I opted for an `Array` instead for the first pass.

packages/core/test/unit/clients/imomentocache-interface.test.ts
nand4011, line 1: I just merged a pr that fixes this, since I was seeing it on other prs: https://github.com/momentohq/client-sdk-javascript/pull/768
I don't mind the eslint disabling vs the specific check and fail though.

packages/core/src/messages/responses/cache-dictionary-get-field.ts
```diff
@@ -29,14 +29,30 @@ const TEXT_DECODER = new TextDecoder();
...
     super();
     this.body = body;
   }
+
+  /**
+   * Returns the data as a utf-8 string, decoded from the underlying byte array.
+   * @returns string
+   */
+  public value(): string {
+    return TEXT_DECODER.decode(this.body);
```
nand4011, line 53: Why duplicate the logic in some responses and call the pre-existing value functions in others?

packages/core/src/messages/responses/cache-get.ts
```diff
@@ -29,14 +30,30 @@ const TEXT_DECODER = new TextDecoder();
  * }
  * ```
  */
-export abstract class Response extends ResponseBase {}
+export abstract class Response extends ResponseBase {
+  public value(): string | undefined {
+    if (this instanceof CacheGet.Hit) {
```
nand4011, line 35: Does this need to import CacheGet.Hit instead of referencing the class in this file?

packages/core/src/messages/responses/cache-sorted-set-get-score.ts
```diff
@@ -26,7 +26,14 @@ import {SdkError} from '../../errors';
  * }
  * ```
  */
-export abstract class Response extends ResponseBase {}
+export abstract class Response extends ResponseBase {
+  public score(): number | undefined {
```
nand4011, line 30: I understand the reasoning, but it feels a little off for this and the increment response to be the only ones without a value() method. Would it be worth having both, or do we not want to confuse users since the 'value' of the set is never returned?

packages/core/src/messages/responses/cache-dictionary-get-field.ts
cprice404: very good callout, will fix

packages/core/src/messages/responses/cache-get.ts
cprice404: agreed will fix

packages/core/src/messages/responses/cache-sorted-set-get-score.ts
cprice404: i'm just choosing to name them the same thing that they are already named on the `Hit` class. I think it would be weird for the parent class to have a different name than the accessor on the `Hit` class?

packages/core/test/unit/clients/imomentocache-interface.test.ts
cprice404: yours is better

packages/core/src/messages/responses/cache-sorted-set-get-score.ts
nand4011: Fair. I was picturing a hypothetical generic Response interface with a value() method. I suppose the question then would be whether `Hit` should have value(). Just a thought, especially since adding it would bring this out of step with the other SDKs.

cprice404: yeah I see what you're getting at for sure, but I think what makes the most sense to me is just a pattern that lets the user pretend like the parent class is the corresponding `Hit` or `Success` subclass.

packages/core/src/messages/responses/cache-get.ts
cprice404: addressed in https://github.com/momentohq/client-sdk-javascript/pull/760/commits/383af5f8b5f34a70ce4a0cf6efa6bd440dd70fb6

packages/core/src/messages/responses/cache-dictionary-get-field.ts
cprice404: addressed in https://github.com/momentohq/client-sdk-javascript/pull/760/commits/383af5f8b5f34a70ce4a0cf6efa6bd440dd70fb6

packages/core/test/unit/clients/imomentocache-interface.test.ts
cprice404: rebased onto yours

nand4011 (approved):
I don't see anything unexpected in the return types of the value() methods. We should get the ok from @allenheltondev, but it looks good to me.

packages/core/src/messages/responses/cache-dictionary-get-field.ts
```diff
@@ -29,14 +29,30 @@ const TEXT_DECODER = new TextDecoder();
...
     super();
     this.body = body;
   }
+
+  /**
+   * Returns the data as a utf-8 string, decoded from the underlying byte array.
+   * @returns string
+   */
+  public value(): string {
+    return this.valueString();
```
allenheltondev, line 53: 👍

packages/core/src/messages/responses/cache-dictionary-get-fields.ts
```diff
@@ -144,6 +151,16 @@ class _Hit extends Response {
...
   }
 
+  /**
+   * Returns the data as a Record whose keys and values are utf-8 strings, decoded from the underlying byte arrays.
+   * This can be used in most places where an Object is desired.  This is a convenience alias for
+   * {valueRecordStringString}.
+   * @returns {Record<string, string>}
+   */
+  public value(): Record<string, string> {
+    return this.valueRecordStringString();
```
allenheltondev, line 161: 👍

packages/core/src/messages/responses/cache-dictionary-increment.ts
```diff
@@ -20,22 +20,32 @@ import {ResponseBase, ResponseError, ResponseSuccess} from './response-base';
...
-    this.value = value;
+    this._value = value;
   }
 
   /**
    * The new value of the element after incrementing.
    * @returns {number}
    */
+  public value(): number {
+    return this._value;
```
allenheltondev, line 45: 👍

packages/core/src/messages/responses/cache-get.ts
```diff
@@ -29,14 +29,30 @@ const TEXT_DECODER = new TextDecoder();
...
     super();
     this.body = body;
   }
+
+  /**
+   * Returns the data as a utf-8 string, decoded from the underlying byte array.
+   * @returns string
+   */
+  public value(): string {
+    return TEXT_DECODER.decode(this.body);
```
allenheltondev, line 53: 👍 ooh i like this, great call

packages/core/src/messages/responses/cache-increment.ts
```diff
@@ -20,22 +20,32 @@ import {ResponseBase, ResponseError, ResponseSuccess} from './response-base';
...
+    this._value = value;
   }
 
   /**
    * The new value of the element after incrementing.
    * @returns {number}
    */
+  public value(): number {
+    return this._value;
+  }
```
allenheltondev, line 46: 👍

packages/core/src/messages/responses/cache-list-fetch.ts
```diff
@@ -63,6 +70,15 @@ class _Hit extends Response {
     return this.valueListString();
   }
 
+  /**
+   * Returns the data as an array of strings, decoded from the underlying byte array.  This is a convenience alias
+   * for {valueListString}
+   * @returns {string[]}
+   */
+  public value(): string[] {
+    return this.valueListString();
```
allenheltondev, line 79: 👍

packages/core/src/messages/responses/cache-list-pop-back.ts
```diff
@@ -46,6 +53,14 @@ class _Hit extends Response {
     return TEXT_DECODER.decode(this.body);
   }
 
+  /**
+   * Returns the data as a utf-8 string, decoded from the underlying byte array.
+   * @returns string
+   */
+  public value(): string {
+    return TEXT_DECODER.decode(this.body);
```
allenheltondev, line 61: 👍

packages/core/src/messages/responses/cache-list-pop-front.ts
```diff
@@ -45,6 +52,14 @@ class _Hit extends Response {
     return TEXT_DECODER.decode(this.body);
   }
 
+  /**
+   * Returns the data as a utf-8 string, decoded from the underlying byte array.
+   * @returns string
+   */
+  public value(): string {
+    return TEXT_DECODER.decode(this.body);
```
allenheltondev, line 60: 👍

packages/core/src/messages/responses/cache-sorted-set-fetch.ts
```diff
@@ -78,6 +85,16 @@ class _Hit extends Response {
...
   }
 
+  /**
+   * Returns the elements as an array of objects, each containing a `value` and `score` field.
+   * The value is a utf-8 string, decoded from the underlying byte array, and the score is a number.
+   * This is a convenience alias for {valueArrayStringNumber}.
+   * @returns {value: string; score: number}[]
+   */
+  public value(): {value: string; score: number}[] {
+    return this.valueArrayStringElements();
```
allenheltondev, line 95: 👍

packages/core/src/messages/responses/cache-sorted-set-get-score.ts
allenheltondev: I am good with score. I share the same hesitation as Nate, but since this is a `get score` api, i'm ok with this

packages/core/src/messages/responses/cache-sorted-set-get-scores.ts
```diff
@@ -126,6 +133,16 @@ class _Hit extends Response {
     return this.valueRecordString();
   }
 
+  /**
+   * Returns the data as a Record whose keys and values are utf-8 strings, decoded from the underlying byte arrays.
+   * This can be used in most places where an Object is desired.  This is a convenience alias for
+   * {valueRecordStringString}.
+   * @returns {Record<string, number>}
+   */
+  public value(): Record<string, number> {
```
allenheltondev, line 142: This is the one semantically i'm not 100% sure about. Since the `get score` one has a `score` value, i wonder if `scores` is more appropriate here?

packages/core/src/messages/responses/cache-sorted-set-increment-score.ts
```diff
@@ -20,7 +20,14 @@ import {ResponseBase, ResponseError, ResponseSuccess} from './response-base';
  * }
  * ```
  */
-export abstract class Response extends ResponseBase {}
+export abstract class Response extends ResponseBase {
+  public score(): number | undefined {
+    if (this instanceof Success) {
```
allenheltondev, line 25: Is the implementation missing from this one? it looks like you might just have the abstract class implemented.

allenheltondev (approved):
In an ideal world I could get these from a `value` property instead of a `.value()` function call. This would allow me to destructure the response in a very concise way.

```javascript
const { value } = await cacheClient.get('my-cache', 'my-key');
```

I know would not always be present like on a cache miss or error and enables devs to not practice safe coding, but a guy can dream.

Overall great choices on the defaults!

cprice404:
> In an ideal world I could get these from a `value` property instead of a `.value()` function call. This would allow me to destructure the response in a very concise way.
> 
> ```js
> const { value } = await cacheClient.get('my-cache', 'my-key');
> ```
> 
> I know would not always be present like on a cache miss or error and enables devs to not practice safe coding, but a guy can dream.
> 
> Overall great choices on the defaults!

Yeah IMO that just seems like a bit too far of a leap of inconsistency from the current API since we chose functions for all the existing ones. I'm totally happy to consider changing everything to properties full-sale in a v2.0 but I think we have to keep them as fns for consistency in 1.x.

packages/core/src/messages/responses/cache-sorted-set-increment-score.ts
cprice404: good eye - the implementation already existed though so it doesn't show up in this diff.

packages/core/src/messages/responses/cache-sorted-set-get-scores.ts
cprice404: hrm; yeah that's a fair callout. If we were starting over from scratch today I might be swayed to name the accessor on the `Hit` class `scores()`, but it's currently named `value*()`. So my response is basically the same one I gave to Nate, that I think that for consistency, these new accessors should have the same names as the existing ones.

We could definitely add `scores()` aliases to both the `Hit` class and the parent class here though in the future, if users call this one out as confusing!

PR #766, opened 2023-08-22
chore: add validation for eager connection timeout

pratik151192:
Noticed while wiring the same on Python SDK that we need a validation for a better UX


nand4011 (approved):
Looks good other than github being angry about the non-null assertions.

nand4011:
I guess github is just generally angry about that. You didn't event change it.

PR #767, opened 2023-08-23
feat: Add event loop delay logging to the metrics logging middleware

nand4011:
Add regular logging of event loop delay and utilization to the metrics logging middleware. The built-in performance monitor doesn't work well in a per-request context, so we log event loop stats on a one-second interval.

### DIFF
--- a/packages/client-sdk-nodejs/src/config/middleware/experimental-metrics-logging-middleware.ts
+++ b/packages/client-sdk-nodejs/src/config/middleware/experimental-metrics-logging-middleware.ts
@@ -5,6 +5,59 @@ import {
   ExperimentalRequestMetrics,
 } from './impl/experimental-metrics-middleware';
 import {MiddlewareRequestHandlerContext} from './middleware';
+import {
+  EventLoopDelayMonitor,
+  EventLoopUtilization,
+  monitorEventLoopDelay,
+  performance,
+} from 'perf_hooks';
+
+interface StateMetrics {
+  /**
+   * The proportion of time the event loop was busy over the last second.
+   */
+  eventLoopUtilization: number;
+
+  /**
+   * The average event loop delay over the last second, measured in 20ms increments.
+   */
+  eventLoopDelayMean: number;
+
+  /**
+   * The minimum event loop delay over the last second, measured in 20ms increments.
+   */
+  eventLoopDelayMin: number;
+
+  /**
+   * The 50th percentile event loop delay over the last second, measured in 20ms increments.
+   */
+  eventLoopDelayP50: number;
+
+  /**
+   * The 75th percentile event loop delay over the last second, measured in 20ms increments.
+   */
+  eventLoopDelayP75: number;
+
+  /**
+   * The 90th percentile event loop delay over the last second, measured in 20ms increments.
+   */
+  eventLoopDelayP90: number;
+
+  /**
+   * The 95th percentile event loop delay over the last second, measured in 20ms increments.
+   */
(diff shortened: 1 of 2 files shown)


packages/client-sdk-nodejs/src/config/middleware/experimental-metrics-logging-middleware.ts
```diff
@@ -42,6 +99,11 @@ class ExperimentalMetricsLoggingMiddlewareRequestHandler extends ExperimentalMet
  * your {Configuration} to enable this middleware.
  */
 export class ExperimentalMetricsLoggingMiddleware extends ExperimentalMetricsMiddleware {
+  private static readonly metricsLogInterval = 1000;
+  private static readonly eventLoopDelayInterval = 20;
+  private static eldMonitor: EventLoopDelayMonitor;
+  private static elu: EventLoopUtilization;
+  private static isLoggingStarted = false;
```
cprice404, line 106: i'm not seeing where you set this to `true`

packages/client-sdk-nodejs/src/config/middleware/experimental-metrics-logging-middleware.ts
```diff
@@ -50,5 +112,42 @@ export class ExperimentalMetricsLoggingMiddleware extends ExperimentalMetricsMid
...
+  }
+
+  private static startLogging(logger: MomentoLogger): void {
+    this.eldMonitor = monitorEventLoopDelay({
+      resolution: this.eventLoopDelayInterval,
+    });
+    this.eldMonitor.enable();
+    this.elu = performance.eventLoopUtilization();
+
+    setTimeout(() => {
```
cprice404, line 127: why do you need this outer `setTimeout`?

cprice404 (commented):
just a few nits

packages/client-sdk-nodejs/src/config/middleware/experimental-metrics-logging-middleware.ts
nand4011: Ha. I guess my plan to only run it once would have failed.

packages/client-sdk-nodejs/src/config/middleware/experimental-metrics-logging-middleware.ts
nand4011: I wanted the first output to occur one second in since there won't be any useful data right away.

cprice404: doesn't setInterval already do that

cprice404:
```
[Parameters](https://developer.mozilla.org/en-US/docs/Web/API/setInterval#parameters)

func

    A [function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function) to be executed every delay milliseconds. The first execution happens after delay milliseconds.
```

https://developer.mozilla.org/en-US/docs/Web/API/setInterval

nand4011: 🤦

cprice404 (approved):
I'm good with this now, thanks! @pratik151192 any comments?

great

pratik151192:
I just had one comment that I didn't press the button for but seems like you already addressed by flipping that flag to true when required. LGTM!

PR #768, opened 2023-08-23
chore: Remove non-null assertions that GitHub doesn't like

nand4011:
Replace the non-null assertions in imomentocache-interface.test.ts with explicit null checks to fix the warnings appearing in prs on GitHub.


bruuuuuuuce (approved):
:shipit:

PR #775, opened 2023-08-25
feat: add nodejs preview vector index client

pgautier404:


packages/client-sdk-nodejs/src/config/vector-configuration.ts
```diff
@@ -0,0 +1,98 @@
...
+  transportStrategy: TransportStrategy;
+  /**
+   * Configures middleware functions that will wrap each request
+   */
+  middlewares: Middleware[];
+}
+
+export class VectorConfiguration implements Configuration {
+  private readonly loggerFactory: MomentoLoggerFactory;
+  private readonly retryStrategy: RetryStrategy;
```
malandis, line 28: I left this out of the Python client since we weren't using it.

malandis (commented):
Looks good to go just one question

packages/client-sdk-nodejs/src/config/vector-configuration.ts
cprice404: yup, removing in subsequent PR, didn't catch it during this one

malandis (approved):

PR #779, opened 2023-08-29
fix: Update dependabot.yml to include _all_ example dirs

bruuuuuuuce:
https://github.com/momentohq/client-sdk-javascript/pull/778/files dependabot was not updating all example directories


anitarua (approved):
Nice!! I hadn't even realized those directories were missing from dependabot

PR #786, opened 2023-09-01
feat: Add vector data plane operations

cprice404:
This needs to be re-based when Anita's PR is merged; probably easier to review after that happens


packages/client-sdk-web/test/integration/shared/vector-data-plane.test.ts
```diff
@@ -0,0 +1,6 @@
+import {SetupVectorIntegrationTest} from '../integration-setup';
+import {runVectorDataPlaneTest} from '@gomomento/common-integration-tests';
+
+const {vectorClient} = SetupVectorIntegrationTest();
+
+runVectorDataPlaneTest(vectorClient);
```
pratik151192, line 6: I will review the rest of the bit once you rebase the PR as you mentioned on the description. Just adding a note that we should probably ignore this suite from the `integration-test` directive and add a new directive for this as well similar to how we did for `control`. You'll also need to update `runJavascriptIntegrationTests` in [canaries-kitchenshink](https://github.com/momentohq/canaries-kitchensink/blob/main/canaries/sdk-canary/index.ts#L1370) for the canaries to pick 'em up.

packages/client-sdk-nodejs/src/internal/internal-auth-client.ts
```diff
@@ -1,4 +1,4 @@
-import {auth} from '@gomomento/generated-types';
```
pratik151192, line 1: I am skipping this file for review

packages/client-sdk-nodejs/src/internal/vector-data-client.ts
```diff
@@ -0,0 +1,242 @@
...
+import {Header, HeaderInterceptorProvider} from './grpc/headers-interceptor';
+import {ClientTimeoutInterceptor} from './grpc/client-timeout-interceptor';
+import {cacheServiceErrorMapper} from '../errors/cache-service-error-mapper';
+import {
+  validateIndexName,
+  validateTopK,
+} from '@gomomento/sdk-core/dist/src/internal/utils';
+import {normalizeSdkError} from '@gomomento/sdk-core/dist/src/errors';
+
+export class VectorDataClient implements IVectorIndexDataClient {
```
pratik151192, line 27: nit: should we call this`VectorIndexDataClient` as seems like we have added the `index` term everywhere else

packages/client-sdk-nodejs/src/preview-vector-index-client.ts
```diff
@@ -0,0 +1,52 @@
...
+    configuration: new CacheConfiguration({
+      loggerFactory: props.configuration.getLoggerFactory(),
+      transportStrategy: props.configuration.getTransportStrategy(),
+      retryStrategy: defaultRetryStrategy(
+        props.configuration.getLoggerFactory()
+      ),
+      middlewares: [],
+    }),
+    credentialProvider: props.credentialProvider,
+  });
```
pratik151192, line 45: This seems like is for `cache` and we should change it to `vector`

packages/client-sdk-web/src/preview-vector-index-client.ts
```diff
@@ -0,0 +1,47 @@
...
+  props: VectorIndexClientProps
+): IVectorIndexControlClient {
+  return new ControlClient({
+    configuration: new CacheConfiguration({
+      loggerFactory: props.configuration.getLoggerFactory(),
+      transportStrategy: props.configuration.getTransportStrategy(),
+    }),
+    credentialProvider: props.credentialProvider,
+  });
+}
```
pratik151192, line 41: Same seems like this is for cache

packages/client-sdk-web/src/preview-vector-index-client.ts
```diff
@@ -0,0 +1,47 @@
...
+  constructor(props: VectorIndexClientProps) {
+    const controlClient: IVectorIndexControlClient = createControlClient(props);
+    const dataClient = createDataClient(props);
+    super(controlClient, dataClient);
+  }
+}
+
+function createControlClient(
+  props: VectorIndexClientProps
+): IVectorIndexControlClient {
```
pratik151192, line 33: Should this name also include the word `VectorIndex` as `ControlClient` is a bit high level

packages/core/src/auth/credential-provider.ts
```diff
@@ -46,6 +64,16 @@ export abstract class CredentialProvider {
    */
   abstract isControlEndpointOverridden(): boolean;
 
+  /**
+   * @returns {boolean} true if the vector endpoint was manually overridden at construction time; false otherwise
```
pratik151192, line 68: comment should say `token endpoint`

packages/core/src/auth/credential-provider.ts
```diff
@@ -169,11 +241,11 @@ export class EnvMomentoTokenProvider extends StringMomentoTokenProvider {
...
     }
-    const decodedToken = decodeAuthToken(authToken);
     super({
-      authToken: decodedToken.authToken,
-      controlEndpoint: props.controlEndpoint ?? decodedToken.controlEndpoint,
-      cacheEndpoint: props.cacheEndpoint ?? decodedToken.cacheEndpoint,
+      authToken: authToken,
+      controlEndpoint: props.controlEndpoint,
+      cacheEndpoint: props.cacheEndpoint,
+      vectorEndpoint: props.vectorEndpoint,
```
pratik151192, line 248: Hmm I am unable to find in this PR where this is actually set to `vector.blah...`

pratik151192 (commented):
Looking really good just some minor nits

packages/client-sdk-nodejs/src/preview-vector-index-client.ts
cprice404: The `ControlClient` is internal, and currently houses the control plane operations for both vector and cache stuff. That is weird and it should probably be split up, but I didn't/don't think it's worth blocking on since it doesn't impact public API. wdyt?

pratik151192: Got it. In Python we did split `_VectorIndexControlClient` which just uses the `vector_endpoint` from the `credentialProvider`; so seems like we could do the same here but I agree this is internal and can be split up at a later point. The configuration, however, I do think can be just changed to use the `VectorConfiguration` we created in this PR.

cprice404: i agree they should be split up when we have time.

The existing, internal ControlClient has a constructor signature that requires a CacheConfiguration as its first argument.

pratik151192: I see. In that case i'm ok with this paritcularly because its not exposed to users.

packages/core/src/auth/credential-provider.ts
cprice404: ty

packages/core/src/auth/credential-provider.ts
cprice404: https://github.com/momentohq/client-sdk-javascript/pull/786/files#diff-df5a3d98528b9dd05cb526fae3b1a3119deb52f12e5815a23c1867200e43f934R63

packages/client-sdk-nodejs/src/internal/vector-data-client.ts
cprice404: fair. will rename in a subsequent PR when I break up the control client.

pratik151192 (commented):
Just adding a comment again for updating `package.json` to ignore vector-data-tests from `integration-test` directive similar for `vector` as I see you adding the final bits/fixes

packages/client-sdk-nodejs/src/config/configurations.ts
```diff
@@ -18,7 +18,7 @@ const defaultLoggerFactory: MomentoLoggerFactory =
   new DefaultMomentoLoggerFactory();
 const defaultMiddlewares: Middleware[] = [];
 
-function defaultRetryStrategy(
```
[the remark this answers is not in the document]
cprice404: it is for now because I had to create a CacheClientConfiguration to pass to the ControlClient, until I spend the time to separate those two control clients.

Do you have concerns about this being exported?

packages/client-sdk-nodejs/src/preview-vector-index-client.ts
```diff
@@ -0,0 +1,52 @@
...
+function createControlClient(
+  props: VectorIndexClientProps
+): IVectorIndexControlClient {
+  return new ControlClient({
+    configuration: new CacheConfiguration({
+      loggerFactory: props.configuration.getLoggerFactory(),
+      transportStrategy: props.configuration.getTransportStrategy(),
+      retryStrategy: defaultRetryStrategy(
+        props.configuration.getLoggerFactory()
+      ),
```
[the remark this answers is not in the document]
cprice404: please see pratik's comments. This is internal and I'm just trying to get the public API out the door. I will break up the control clients later.

packages/client-sdk-nodejs/src/config/configurations.ts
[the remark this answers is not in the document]
malandis: Nope. Was just asking if it was an oversight.

packages/client-sdk-nodejs/src/internal/vector-data-client.ts
```diff
@@ -0,0 +1,242 @@
...
+            resolve(new VectorSearch.Error(cacheServiceErrorMapper(err)));
+          }
+        }
+      );
+    });
+  }
+
+  private validateRequestTimeout(timeout?: number) {
+    this.logger.debug(`Request timeout ms: ${String(timeout)}`);
+    if (timeout && timeout <= 0) {
```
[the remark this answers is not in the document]
cprice404: addressed in https://github.com/momentohq/client-sdk-javascript/pull/786/commits/d9966e65639f35beb820f1a2293f9bfc860f78e0

malandis (approved):

pratik151192 (approved):
LGTM!!!

PR #793, opened 2023-09-05
chore: Exclude vector index tests from build-and-run-tests

krispraws:


pratik151192:
Is this directive being run server-side/canaries? Canaries were running `integration-test` so had updated that only but not sure if this is used as well

krispraws:
> Is this directive being run server-side/canaries? Canaries were running `integration-test` so had updated that only but not sure if this is used as well

I don't think it is used server side or in canaries. But it is used in github workflows: https://github.com/momentohq/client-sdk-javascript/blob/9088d83a342cdcf7289a6b6c3581346b3a38a57a/.github/workflows/build.yml#L74

Looks like the CI build fails with the same error i got locally. Maybe there is some other issue with vector indexes in alpha

anitarua (approved):

krispraws:
I will withdraw this PR because we do want vector index CI tests to run on PRs.

PR #795, opened 2023-09-05
fix: set grpc byte fields to base64 strings

krispraws:
Byte fields for key and key prefix were not set correctly when generating the token.
Web sdk auth integration tests pass now.


anitarua (approved):
thanks for fixing these!!

PR #785, opened 2023-08-30
feat: add generateDisposableTokens

anitarua:
Addresses [#422](https://github.com/momentohq/dev-eco-issue-tracker/issues/422) to add generateDisposableTokens and key-level FGAC permissions in web and node SDKs.

Also includes tests for key-level FGAC and additional tests for the new WriteOnly role (I'd missed coverage there with the last PR).

PR #796, opened 2023-09-06
chore: clean up disposable token integration tests and add one for cache and topic permissions

anitarua:
Follow up to #785 to clean up the integration tests and add a new one to show disposable tokens work for the usual cache and topic permissions too.

### DIFF
--- a/packages/common-integration-tests/src/auth-client.ts
+++ b/packages/common-integration-tests/src/auth-client.ts
@@ -831,45 +831,28 @@ export function runAuthClientTests(
 
       // 2. Gets to a specific key should work for both caches, and not for any other key
       const getKey1 = await cacheClient.get(FGA_CACHE_1, 'cow');
-      if (getKey1 instanceof CacheGet.Hit) {
-        expect(getKey1.value()).toEqual('moo');
-      } else if (getKey1 instanceof CacheGet.Error) {
-        const asError = getKey1;
-        console.debug(asError.errorCode(), asError.message());
-      }
-      // expect(getKey1).toBeInstanceOf(CacheGet.Hit);
-      // expect(getKey1.value()).toEqual('moo');
+      expect(getKey1).toBeInstanceOf(CacheGet.Hit);
+      expect(getKey1.value()).toEqual('moo');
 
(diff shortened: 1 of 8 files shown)


packages/common-integration-tests/src/auth-client.ts
```diff
@@ -1340,6 +1250,98 @@ export function runAuthClientTests(
...
+      const getKey1 = await cacheClient.get(FGA_CACHE_1, 'cow');
+      expect(getKey1).toBeInstanceOf(CacheGet.Hit);
+      expect(getKey1.value()).toEqual('meow');
+
+      const setResp3 = await cacheClient.set(
+        FGA_CACHE_2,
+        FGA_CACHE_2_KEY,
+        'woof'
+      );
+      expect(setResp3).toBeInstanceOf(CacheSet.Error);
```
krispraws, line 1298: You can use `expectPermissionDeniedForSet` here

packages/common-integration-tests/src/auth-client.ts
```diff
@@ -1340,6 +1250,98 @@ export function runAuthClientTests(
...
+        'humans landed on Mars!'
+      );
+      expect(pubResp1).toBeInstanceOf(TopicPublish.Success);
+
+      const subResp2 = await topicClient.subscribe(
+        FGA_CACHE_1,
+        'breaking news!',
+        trivialHandlers
+      );
+      expect(subResp2).toBeInstanceOf(TopicSubscribe.Error);
```
krispraws, line 1329: I found we can actually [extend the jest matchers](https://jestjs.io/docs/expect#expectextendmatchers) after I wrote the wrapper functions `expectPermissionDeniedForGet` and `expectPermissionDeniedForSet`. If you could try it out for this new test, it will greatly simplify writing more of these tests. No need to refactor existing tests right now, and I understand if you don't have the time.
E.g. we could write this test like
```
expect(getKey1).toBeHit(expectedValue);
expect(setResp).toBePermissionDeniedForSet();
expect(subResp2).toBePermissionDeniedForTopicSubscribe();
```

krispraws (approved):
🎉 This looks great! Thank you so much for refactoring everything!

cprice404 (approved):
cool

PR #813, opened 2023-09-06
chore: update both TVM examples to vend disposable tokens

anitarua:
Addresses [#422](https://github.com/momentohq/dev-eco-issue-tracker/issues/422). Updates both static and dynamic token vending machine examples to use generateDisposableToken instead of generateAuthToken.

Plus a bug fix for the Vite chat app to allow ReadWrite access to users if they haven't set the auth method to "cognito".


bruuuuuuuce (approved):
:shipit:

PR #814, opened 2023-09-06
chore: rename integration test variables

cprice404:
This commit contains no functional changes; it simply renames some variables that we have been using in integration tests. Our integration test setup code, from way back at the beginning of the repo, created variables named `Momento` and `IntegrationTestCacheName` that it injected into all of the tests. The upper-case naming is not idiomatic of TS/JS, and the name `Momento` is confusing when the object is a `CacheClient`. This has been annoying me for a long time so this commit just fixes the names.


bruuuuuuuce (approved):
:shipit:

nand4011 (approved):

PR #815, opened 2023-09-06
chore: split up internal control client between vector and cache functions

cprice404:
Prior to this commit, we had a single internal ControlClient that housed the control plane APIs for both cache and vector products. This client required an instance of CacheConfiguration for its constructor, which didn't make sense for the Vector use case, so we had a bit of a hack in place to bridge it while we got the public API released.

This commit splits up the internal control client into two so that we don't have to munge the Config object. Also makes sure that all of the Vector classes are prefixed with VectorIndex for consistency.


pratik151192 (approved):
🚀

PR #816, opened 2023-09-07
fix: clean up the support for overriding endpoints

cprice404:
Our credential providers support overriding the endpoints, although we do not publicize this and I do not believe any users are using it yet.

As we have added more endpoints this has begun to get unwieldy. This commit simplifies things so that it is possible to override only the base endpoint, and allow the others to be derived from it. If you do choose to explicitly override specific endpoints you are now required to override all of them, which makes the logic simpler and allows for better compile-time checking.


bruuuuuuuce (approved):
:shipit:

anitarua (approved):

PR #836, opened 2023-09-07
fix: cancel stale workflows

bruuuuuuuce:
This commit cancels in progress jobs instead of letting them run to completion. It should save us quite a bit on gh actions minutes, and will prevent dependabot from prematurely merging prs.

This pr is currently only cancelling the `build.yml` jobs https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

If we wanted to, we could try cancelling _all_ stale jobs https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-concurrency-to-cancel-any-in-progress-job-or-run

Not sure which is the best strategy to go with, open to changing it to cancel all jobs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,6 +4,10 @@ on:
   pull_request:
     branches: [main]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   verify:
     runs-on: macos-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -4,6 +4,10 @@ on:
   pull_request:
     branches: [main]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
bruuuuuuuce, line 8: If we want to cancel _all_ jobs, this would change to

```
  group: ${{ github.ref }}
```

cprice404 (approved):

PR #837, opened 2023-09-07
chore: rename GenerateAuthToken response classes to GenerateApiToken

cprice404:
This commit is the first in a series that will refactor the name
of our `GenerateAuthToken` method to `GenerateApiKey`. In this commit
we don't do anything but change the export statements for the response
types in the index files. We retain an alias using the old name
for backward compatibility.


allenheltondev (dismissed):

ben-kugler (dismissed):

krispraws (dismissed):

bruuuuuuuce (approved):
:shipit:

PR #858, opened 2023-09-11
chore: use expectWithMessage in FGAC auth integration tests 

anitarua:
FGAC tests were failing on a separate PR but no error message was provided. Replaced instances of `expect` with `expectWithMessage` so that future CI test failures will show the error message.

Also replaced one instance of `generateAuthToken` with `generateApiKey` and changed the FGAC test cache prefix from `fga-` to `js-fga-`.


cprice404 (approved):
thank you

PR #859, opened 2023-09-12
chore: disable dependabot and add a script that updates all example deps

cprice404:
This commit disables dependabot for now; it was unsustainable for this
repo with all of the example projects. We can try to enable it again
once they address https://github.com/dependabot/dependabot-core/issues/2178 .

For now we add a script that updates all the example dependency versions.
We can run this by hand after releases, or add a new cron-type Action that
runs it once a day and generates a PR if it produces a diff.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,106 +5,123 @@
 
 version: 2
 updates:
-  - package-ecosystem: "npm" # See documentation for possible values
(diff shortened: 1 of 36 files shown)


anitarua (approved):
lgtm!

bruuuuuuuce (approved):
:shipit:

PR #875, opened 2023-09-12
chore: replace more occurrences of "auth token" with "API key"

cprice404:


anitarua (approved):

bruuuuuuuce:
looks like this branch has some conflicts @cprice404

examples/nodejs/access-control/access-control.ts
```diff
@@ -74,14 +74,14 @@ async function generateAuthToken(
 ): Promise<[string, string]> {
   const generateTokenResponse = await authClient.generateAuthToken(scope, ExpiresIn.seconds(durationSeconds));
   if (generateTokenResponse instanceof GenerateAuthToken.Success) {
-    console.log(`Generated an auth token with ${scope.toString()} scope at time ${Date.now() / 1000}!`);
+    console.log(`Generated an API key with ${scope.toString()} scope at time ${Date.now() / 1000}!`);
     console.log('Logging only a substring of the tokens, because logging security credentials is not advisable:');
-    console.log(`Auth token starts with: ${generateTokenResponse.authToken.substring(0, 10)}`);
+    console.log(`API key starts with: ${generateTokenResponse.authToken.substring(0, 10)}`);
     console.log(`Refresh token starts with: ${generateTokenResponse.refreshToken.substring(0, 10)}`);
```
rishtigupta, line 80: Are we not calling Refresh Token as Refresh Key from this change onwards?

cprice404: refresh token is a different resource.

When you generate an API Key, it comes with a refresh token that can be used to refresh the key. AFAIK no one has proposed that we change that one.

Very good question though, I had to stop and think about it for a while when I was making these changes.

bruuuuuuuce (approved):
:shipit:

examples/nodejs/access-control/access-control.ts
rishtigupta: Good to know, thanks Chris. Will make appropriate changes in the console too.

rishtigupta (approved):

PR #880, opened 2023-09-12
fix: export VectorIndexItem from core and sdks

malandis:
Previously, the type `VectorIndexItem` was not exported from sdk-core,
hence could not be re-exported from the SDKs. Because of this,
application code could not make use of the `VectorIndexItem`
type. This PR exports the type accordingly.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -68,6 +68,7 @@ import * as GenerateDisposableToken from './messages/responses/generate-disposab
 // VectorClient Response Types
 export * as vector from './messages/responses/vector';
 export * from './messages/responses/vector';
+export {VectorIndexItem} from './messages/vector-index';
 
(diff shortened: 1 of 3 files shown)


packages/core/src/index.ts
```diff
@@ -68,6 +68,7 @@ import * as GenerateDisposableToken from './messages/responses/generate-disposab
 // VectorClient Response Types
 export * as vector from './messages/responses/vector';
 export * from './messages/responses/vector';
+export {VectorIndexItem} from './messages/vector-index';
```
malandis, line 71: In some of the `index.ts` files we first import a name, then later export it. I am just directly exporting it like so. I assume this is just a stylistic difference.

anitarua (approved):

bruuuuuuuce (approved):
:shipit:

PR #916, opened 2023-09-14
feat: add batch get, set, delete functions for nodejs

anitarua:
Addresses [#460](https://github.com/momentohq/dev-eco-issue-tracker/issues/460) by implementing a `BatchUtils` namespace that offers the `batchGet`, `batchSet`, and `batchDelete` functions in the Nodejs SDK.

### DIFF
--- a/packages/client-sdk-nodejs/src/batchutils/batch-functions.ts
+++ b/packages/client-sdk-nodejs/src/batchutils/batch-functions.ts
@@ -1,91 +1,103 @@
-import {CacheDelete, CacheGet, CacheSet} from '@gomomento/sdk-core';
-import {CacheClient} from '../cache-client';
 import {
-  BatchDeleteRequest,
+  CacheDelete,
+  CacheGet,
+  CacheSet,
+  ICacheClient,
+} from '@gomomento/sdk-core';
+import {
+  BatchDeleteOptions,
   BatchDeleteResponse,
-  BatchGetRequest,
+  BatchGetOptions,
   BatchGetResponse,
-  BatchSetItem,
-  BatchSetRequest,
+  BatchSetOptions,
   BatchSetResponse,
   defaultMaxConcurrentRequests,
+  defaultTtlSeconds,
 } from './batch-props';
 import {range} from '@gomomento/sdk-core/dist/src/internal/utils';
 
 export {
-  BatchDeleteRequest,
+  BatchDeleteOptions,
   BatchDeleteResponse,
-  BatchGetRequest,
+  BatchGetOptions,
   BatchGetResponse,
-  BatchSetItem,
-  BatchSetRequest,
+  BatchSetOptions,
   BatchSetResponse,
   defaultMaxConcurrentRequests,
+  defaultTtlSeconds,
 };
 
 export async function batchGet(
-  request: BatchGetRequest
+  cacheClient: ICacheClient,
+  cacheName: string,
+  keys: Array<string | Uint8Array>,
+  options?: BatchGetOptions
 ): Promise<BatchGetResponse> {
-  const maxConcurrentGets = request.maxConcurrentGets
-    ? request.maxConcurrentGets
-    : Math.min(defaultMaxConcurrentRequests, request.keys.length);
+  const maxConcurrentGets = options?.maxConcurrentGets
+    ? options.maxConcurrentGets
+    : Math.min(defaultMaxConcurrentRequests, keys.length);
 
   const batchGetResults = range(maxConcurrentGets).map((workerId: number) =>
-    getWorker(workerId, request.cacheClient, request.cacheName, request.keys)
+    getWorker(workerId, cacheClient, cacheName, keys)
   );
   const awaitAll = await Promise.all(batchGetResults);
 
   const batchGetResponse: BatchGetResponse = {};
   awaitAll.forEach(responses => {
-    for (const key of Object.keys(responses)) {
-      batchGetResponse[key] = responses[key];
-    }
+    Object.assign(batchGetResponse, responses);
   });
   return batchGetResponse;
 }
 
 async function getWorker(
   workerId: number,
-  cacheClient: CacheClient,
+  cacheClient: ICacheClient,
   cacheName: string,
-  keys: Array<string>
+  keys: Array<string | Uint8Array>
 ): Promise<Record<string, CacheGet.Response>> {
   const responses: Record<string, CacheGet.Response> = {};
   while (keys.length) {
     const cacheKey = keys.pop();
     if (cacheKey !== undefined) {
-      responses[cacheKey] = await cacheClient.get(cacheName, cacheKey);
+      responses[String(cacheKey)] = await cacheClient.get(cacheName, cacheKey);
     }
   }
   return Promise.resolve(responses);
 }
 
 export async function batchSet(
-  request: BatchSetRequest
+  cacheClient: ICacheClient,
+  cacheName: string,
+  keys: Array<string | Uint8Array>,
+  values: Array<string | Uint8Array>,
+  options?: BatchSetOptions
 ): Promise<BatchSetResponse> {
-  const maxConcurrentSets = request.maxConcurrentSets
-    ? request.maxConcurrentSets
-    : Math.min(defaultMaxConcurrentRequests, request.items.length);
+  const maxConcurrentSets = options?.maxConcurrentSets
+    ? options.maxConcurrentSets
+    : Math.min(defaultMaxConcurrentRequests, keys.length);
+  const ttl = options?.ttl ? options.ttl : defaultTtlSeconds;
 
+  const items = keys.map((key, i) => {
+    return {key: String(key), value: String(values[i])};
+  });
   const batchSetResults = range(maxConcurrentSets).map((workerId: number) =>
-    setWorker(workerId, request.cacheClient, request.cacheName, request.items)
+    setWorker(workerId, cacheClient, cacheName, items, ttl)
   );
   const awaitAll = await Promise.all(batchSetResults);
 
   const batchSetResponse: BatchSetResponse = {};
   awaitAll.forEach(responses => {
-    for (const key of Object.keys(responses)) {
-      batchSetResponse[key] = responses[key];
-    }
+    Object.assign(batchSetResponse, responses);
   });
   return batchSetResponse;
 }
 
 async function setWorker(
   workerId: number,
-  cacheClient: CacheClient,
+  cacheClient: ICacheClient,
   cacheName: string,
-  items: Array<BatchSetItem>
+  items: Array<Record<string, string>>,
+  ttl: number
 ): Promise<Record<string, CacheSet.Response>> {
   const responses: Record<string, CacheSet.Response> = {};
   while (items.length) {
@@ -94,7 +106,8 @@ async function setWorker(
       responses[item.key] = await cacheClient.set(
         cacheName,
         item.key,
-        item.value
+        item.value,
+        {ttl}
       );
     }
   }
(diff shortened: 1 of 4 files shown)


packages/client-sdk-nodejs/src/batchutils/batch-functions.ts
```diff
@@ -0,0 +1,145 @@
...
+    }
+  });
+  return batchGetResponse;
+}
+
+async function getWorker(
+  workerId: number,
+  cacheClient: CacheClient,
+  cacheName: string,
+  keys: Array<string>
```
cprice404, line 51: you probably already know this, and I also might have mentioned it in the ticket, but this only works in TS/JS because the node runtime is single threaded. Otherwise it would be very important to use a thread-safe data structure here since multiple workers are sharing a reference to it.

packages/client-sdk-nodejs/src/batchutils/batch-functions.ts
```diff
@@ -0,0 +1,145 @@
...
+): Promise<Record<string, CacheGet.Response>> {
+  const responses: Record<string, CacheGet.Response> = {};
+  while (keys.length) {
+    const cacheKey = keys.pop();
+    if (cacheKey) {
```
cprice404, line 56: nit/thought about JS semantics - JS has a really flexible definition of `truthiness`. Lots of things like `0` and a boolean `false` would fail this condition, so it's probably best to read the docs for `pop` and explicitly test for what it returns when the list is empty, which would guess is going to either be `null` or `undefined`.

packages/client-sdk-nodejs/src/batchutils/batch-functions.ts
```diff
@@ -0,0 +1,145 @@
...
+  BatchGetRequest,
+  BatchGetResponse,
+  BatchSetItem,
+  BatchSetRequest,
+  BatchSetResponse,
+  defaultMaxConcurrentRequests,
+} from './batch-props';
+
+export async function batchGet(
+  request: BatchGetRequest
```
cprice404, line 27: This Request-based pattern for our APIs is kind of unique to our golang SDK. For TS, I think you will want the signature to look something like this:

```
export async function batchGet(cacheClient: ICacheClient, cacheName: string, keys: Array<string | UInt8Array>, options?: BatchGetOptions);
```

where `string | UInt8Array` should just be whatever we normally accept for a `get`, and the `BatchOptions` interface would look something like this:

```
export interface BatchOptions {
   maxConcurrentRequests?: number
}
```

Same comment for the other 2 APIs.

packages/client-sdk-nodejs/src/batchutils/batch-functions.ts
```diff
@@ -0,0 +1,145 @@
...
+    ? request.maxConcurrentSets
+    : Math.min(defaultMaxConcurrentRequests, request.items.length);
+
+  const batchSetResults = range(maxConcurrentSets).map((workerId: number) =>
+    setWorker(workerId, request.cacheClient, request.cacheName, request.items)
+  );
+  const awaitAll = await Promise.all(batchSetResults);
+
+  const batchSetResponse: BatchSetResponse = {};
+  awaitAll.forEach(responses => {
```
cprice404, line 76: I *think* you could simplify this down to:

```
const batchSetResponse = Obect.assign({}, ...responses);
```

packages/client-sdk-nodejs/src/batchutils/batch-functions.ts
```diff
@@ -0,0 +1,145 @@
...
+export {
+  BatchDeleteRequest,
+  BatchDeleteResponse,
+  BatchGetRequest,
+  BatchGetResponse,
+  BatchSetItem,
+  BatchSetRequest,
+  BatchSetResponse,
+  defaultMaxConcurrentRequests,
+} from './batch-props';
```
pgautier404, line 24: I'm not sure you need `from './batch-props'` here since you've already imported these above. I don't think it hurts anything but may be superfluous strictly speaking.

packages/client-sdk-nodejs/src/index.ts
```diff
@@ -330,4 +331,6 @@ export {
   MomentoLoggerFactory,
   NoopMomentoLogger,
   NoopMomentoLoggerFactory,
+  // BatchUtils
```
pgautier404, line 334: Please remove.

pgautier404 (commented):
I left a couple super picky comments, but I think this looks great!

packages/client-sdk-nodejs/src/batchutils/batch-props.ts
```diff
@@ -0,0 +1,36 @@
...
+import {CacheClient} from '../cache-client';
+
+export const defaultMaxConcurrentRequests = 5;
+
+export interface BatchGetRequest {
+  cacheClient: CacheClient;
+  cacheName: string;
+  keys: Array<string>;
+  maxConcurrentGets?: number;
+}
```
cprice404, line 11: as mentioned above I think you can get rid of all the request objects in this file. And you will want to create an Options type/interface for each one. For now all 3 options objects will probably all be the same so you can alias them, like what we do here:

https://github.com/momentohq/client-sdk-javascript/blob/d3b796575829f8406d42ae902cda04b379308487/packages/core/src/clients/IMomentoCache.ts#L62

packages/client-sdk-nodejs/src/batchutils/batch-props.ts
```diff
@@ -0,0 +1,36 @@
+import {CacheGet, CacheSet, CacheDelete} from '@gomomento/sdk-core';
+import {CacheClient} from '../cache-client';
+
+export const defaultMaxConcurrentRequests = 5;
+
+export interface BatchGetRequest {
+  cacheClient: CacheClient;
```
bruuuuuuuce, line 7: I don't see anything here that makes this node specific, why not tie this to the `ICacheClient` interface from core, and then we can implement this in both nodejs and web-sdk?

packages/client-sdk-nodejs/src/batchutils/batch-props.ts
```diff
@@ -0,0 +1,36 @@
...
+  cacheName: string;
+  keys: Array<string>;
+  maxConcurrentGets?: number;
+}
+
+export type BatchGetResponse = Record<string, CacheGet.Response>;
+
+export interface BatchSetItem {
```
cprice404, line 15: I think you can get rid of this too, you want the API for `batchSet` to feel as similar as possible to the API for `cacheClient.set`. Yours will have the extra `cacheClient` argument at the beginning of its arg list, and the options arg for the maxConcurrentRequests, but otherwise you want the signatures to feel similar.

Which raises the question of whether we should have a TTL here; I think we probably should, you might want to fold that into your `options` argument?

packages/client-sdk-nodejs/src/index.ts
cprice404, line 334: comment

packages/client-sdk-nodejs/test/integration/batchutils.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    };
+    const response = await batchSet(request);
+    for (const [key, resp] of Object.entries(response)) {
+      expectWithMessage(() => {
+        expect(resp).toBeInstanceOf(CacheSet.Success);
+      }, `expected SUCCESS for item ${key} but received ${resp.toString()}`);
+    }
+  });
+
+  it('batchGet happy path with all hits', async () => {
```
cprice404, line 53: I like your test cases! maybe add one more here for mixed hits/misses though?

packages/client-sdk-nodejs/test/integration/batchutils.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    };
+    const response = await batchGet(request);
+    for (const [key, resp] of Object.entries(response)) {
+      expectWithMessage(() => {
+        expect(resp).toBeInstanceOf(CacheGet.Hit);
+      }, `expected HIT for key ${key}, received ${resp.toString()}`);
+    }
+  });
+
+  it('batchDelete happy path', async () => {
```
cprice404, line 87: might be worth adding one more test case where you call delete on a key that doesn't exist as part of your batch

cprice404 (commented):
looking really good! just a few minor thoughts, mostly API-related

packages/client-sdk-nodejs/src/batchutils/batch-functions.ts
```diff
@@ -0,0 +1,156 @@
...
+  cacheClient: ICacheClient,
+  cacheName: string,
+  keys: Array<string | Uint8Array>,
+  values: Array<string | Uint8Array>,
+  options?: BatchSetOptions
+): Promise<BatchSetResponse> {
+  const maxConcurrentSets = options?.maxConcurrentSets
+    ? options.maxConcurrentSets
+    : Math.min(defaultMaxConcurrentRequests, keys.length);
+  const ttl = options?.ttl ? options.ttl : defaultTtlSeconds;
```
pratik151192, line 78: One of the reasons we added `BatchSetItem` as a separate interface in go-lang was because with this signature we are assuming that a user will not mess up providing `keys` and `values` out of order. Also, a user might want to provide different TTLs for the items in the batchRequest. Given these reasons, in go we have this interface:

```
type BatchSetItem struct {
	Key   momento.Key
	Value momento.Value
	Ttl   time.Duration
}
``

which is what the BatchSet accepts as a list

```
type BatchSetRequest struct {
	Client            momento.CacheClient
	CacheName         string
	Items             []BatchSetItem
	MaxConcurrentSets int
}
```

Do you think it's worth doing that for a/ consistency, and b/ giving users more flexibility and typing it to avoid ordering errors? @cprice404 would love your thoughts too.

File: packages/client-sdk-nodejs/src/batchutils/batch-functions.ts [hunk 13]
```diff
@@ -0,0 +1,156 @@
...
+
+  const items = keys.map((key, i) => {
+    return {key: String(key), value: String(values[i])};
+  });
+  const batchSetResults = range(maxConcurrentSets).map((workerId: number) =>
+    setWorker(workerId, cacheClient, cacheName, items, ttl)
+  );
+  const awaitAll = await Promise.all(batchSetResults);
```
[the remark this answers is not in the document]
anitarua: My understanding was that timeouts are baked into the cacheClient that the user provides (just looking at the [tests here](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/test/integration/client-timeout.test.ts)), and timed out requests would return a CacheSet.Error. 

Does this address your individual request timeout concern? Or are you thinking of timeouts at the worker level?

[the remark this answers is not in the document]
cprice404: I agree with your general feedback @pratik151192 , it's a great thing to keep an eye out for. In this case though I think @anitarua is correct that all of these promises will have a client-side deadline so I think it's safe. We could add a comment in the code above this line to make it clear we've considered it.

packages/client-sdk-nodejs/src/batchutils/batch-functions.ts
cprice404: Yeah that's a fair callout. Thanks for the suggestion @pratik151192 .

Sorry for the mis-direction on that part in my earlier feedback, @anitarua , but I do like @pratik151192 's suggestion here for adding a `BatchSetItem` interface. The TTL therein should be optional, and you can fall back to the default TTL for the client for any item that it's not set explicitly on. That will allow you to remove the ttl flag from `BatchSetOptions`, as well.

packages/client-sdk-nodejs/src/batchutils/batch-functions.ts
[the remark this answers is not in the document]
pratik151192: Makes sense. Thanks both!

pratik151192 (approved):
🎉👏🚀

cprice404 (approved):
party

PR #925, opened 2023-09-19
chore: add example of batch set, get, delete functions

anitarua:
Last task of [#460](https://github.com/momentohq/dev-eco-issue-tracker/issues/460): `nodejs/cache` example for batch get, set, delete functions.


cprice404 (approved):
:ty:

PR #949, opened 2023-09-22
feat: leaderboard api stubs

anitarua:
Addresses [#468](https://github.com/momentohq/dev-eco-issue-tracker/issues/468).

Just stubbed versions of the Leaderboard APIs for now. They just throw a "Not implemented" error.

Note: Will be merged into the `leaderboard-apis` branch and a follow-up PR will add some implementations for the first few APIs. Both sets of additions will be merged into `main` at the same time later.


packages/client-sdk-nodejs/src/internal/leaderboard-client.ts
```diff
@@ -0,0 +1,192 @@
...
+  validateSortedSetRanks,
+  validateSortedSetScores,
+  validateSortedSetOffset,
+  validateSortedSetCount,
+} from '@gomomento/sdk-core/dist/src/internal/utils';
+import {Configuration} from '../config/configuration';
+import {delay} from '@gomomento/common-integration-tests';
+
+export class LeaderboardClient implements InternalLeaderboardClient {
+  private readonly configuration: Configuration;
```
cprice404, line 28: nit - this is the Configuration for the CacheClient. it has stuff on it that we probably won't want to have on the leaderboard client for day 1, and their respective configs may change over time, so we probably need to create a separate LeaderBoardConfiguration class/interface.

packages/client-sdk-nodejs/src/leaderboard-client-props.ts
```diff
@@ -0,0 +1,12 @@
+import {Configuration, CredentialProvider} from '.';
+
+export interface LeaderboardClientProps {
+  /**
+   * Configuration settings for the vector client
```
cprice404, line 5: s/vector/leaderboard

packages/core/src/clients/ILeaderboardClient.ts
```diff
@@ -0,0 +1,51 @@
...
+  LeaderboardRemoveElements,
+  LeaderboardUpsert,
+} from '../messages/responses/leaderboard';
+import {
+  SortedSetFetchByRankCallOptions,
+  SortedSetFetchByScoreCallOptions,
+} from '../utils';
+
+export type LeaderboardFetchByRankOptions = SortedSetFetchByRankCallOptions;
+export type LeaderboardFetchByScoreOptions = SortedSetFetchByScoreCallOptions;
```
cprice404, line 15: nicely done

packages/core/src/clients/ILeaderboardClient.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export interface ILeaderboardClient {
+  leaderboardUpsert(
+    cacheName: string,
+    leaderboardName: string,
+    elements: Map<bigint, number>
```
cprice404, line 21: I wonder if we should consider making this `Map<bigint | number, number>`? On the output side we have to do `bigint` or we risk losing precision, but on the input side we can potentially make it more flexible for people?

packages/core/src/clients/ILeaderboardClient.ts
```diff
@@ -0,0 +1,51 @@
...
+  ): Promise<LeaderboardFetch.Response>;
+  leaderboardFetchByRank(
+    cacheName: string,
+    leaderboardName: string,
+    options?: LeaderboardFetchByRankOptions
+  ): Promise<LeaderboardFetch.Response>;
+  leaderboardGetRank(
+    cacheName: string,
+    leaderboardName: string,
+    elementId: bigint
```
cprice404, line 36: i started a thread on this in slack but we might want to make sure everyone is aligned on the name for this field. I think we might end up just shortening this to `id`, pending conversation.

Also might want to support `|number` here (unless TypeScript is kind enough to automatically cast a `number` to a bigint when you call a function, in which case you can ignore all similar feedback that I have left).

packages/core/src/clients/ILeaderboardClient.ts
```diff
@@ -0,0 +1,51 @@
...
+  ): Promise<LeaderboardGetRank.Response>;
+  leaderboardLength(
+    cacheName: string,
+    leaderboardName: string
+  ): Promise<LeaderboardLength.Response>;
+  leaderboardRemoveElements(
+    cacheName: string,
+    leaderboardName: string,
+    elementIds: Array<bigint>
```
cprice404, line 45: same comments here:
* might want to rename to `ids` pending convo
* add `|number`, unless TS will automatically cast

packages/core/src/messages/responses/leaderboard/leaderboard-length.ts
```diff
@@ -0,0 +1,66 @@
...
+  constructor(length: number) {
+    super();
+    this._length = length;
+  }
+
+  /**
+   * Returns the length of the leaderboard
+   * @returns {number}
+   */
+  public length(): number {
```
cprice404, line 36: is there a chance that this needs to be `bigint` too? might be a question for the backend team

cprice404: i see you are returning `bigint` for `rank` so it definitely seems like this one would need to be `bigint`?

packages/core/src/internal/clients/leaderboard/AbstractLeaderboardClient.ts
```diff
@@ -0,0 +1,199 @@
...
+   * The leaderboard is also created if it does not already exist.
+   *
+   * @param {string} cacheName - The cache containing the leaderboard.
+   * @param {string} leaderboardName - The leaderboard to upsert to.
+   * @param {Map<bigint, number>} elements - The ID->score pairs to add to the leaderboard.
+   * @returns {Promise<LeaderboardUpsert.Response>} -
+   * {@link LeaderboardUpsert.Success} on success.
+   * {@link LeaderboardUpsert.Error} on failure.
+   */
+  public async leaderboardUpsert(
```
cprice404, line 39: Have we had conversations with the server team about limits for this API? max number of elements and/or max payload size? If we are doing this as a unary API rather than a streaming one we will need to provide guidance to users around how to batch things.

This comment is not a blocker for this PR but I think we might want to augment this docstring with a link to a docs page or examples of how to use this for bulk ingest, so we should probably ticket that.

packages/core/src/internal/clients/leaderboard/AbstractLeaderboardClient.ts
```diff
@@ -0,0 +1,199 @@
...
+   * @param {number} [options.offset] - The number of elements to skip before
+   * returning the first element. Defaults to 0. Note: this is not the score of
+   * the first element to return, but the number of elements of the result set
+   * to skip before returning the first element.
+   * @param {number} [options.count] - The maximum number of elements to return.
+   * Defaults to undefined, which returns all elements.
+   * @returns {Promise<LeaderboardFetch.Response>} -
+   * {@link LeaderboardFetch.Success} containing the requested elements.
+   * {@link LeaderboardFetch.Error} on failure.
+   */
```
cprice404, line 72: similar comment here - this function has the potential to either blow out the memory on the client, or get throttled by the server, for large leaderboards. We will need to be able to advise customers on how to use it. If backend team doesn't have concrete answers for now then we need a ticket to circle back to this and add link to docs/examples.

packages/core/src/internal/clients/leaderboard/AbstractLeaderboardClient.ts
```diff
@@ -0,0 +1,199 @@
...
+   * @param {number} [options.endRank] - The rank of the last element to fetch.
+   * This rank is exclusive, ie the element at this rank will not be fetched.
+   * Defaults to null, which fetches up until and including the last element.
+   * @param {SortedSetOrder} [options.order] - The order to fetch the elements in.
+   * Defaults to ascending.
+   * @returns {Promise<LeaderboardFetch.Response>} -
+   * {@link LeaderboardFetch.Success} containing the requested elements.
+   * {@link LeaderboardFetch.Error} on failure.
+   */
+  public async leaderboardFetchByRank(
```
cprice404, line 107: same comment as FetchByScore

packages/core/src/messages/responses/leaderboard/leaderboard-length.ts
anitarua: Good catch, the proto returns length as `bigint` as well

https://github.com/momentohq/client-protos/blob/main/proto/leaderboard.proto#L147

packages/core/src/internal/clients/leaderboard/AbstractLeaderboardClient.ts
anitarua: @cprice404 
The API proposal draft noted that there's an upper limit of 8192 elements for the upsert, remove, fetch by score, and fetch by rank APIs. I added a new validator to check for number of elements per batch and also added that note to the docstrings. I can add another task to the github issue for creating docs/examples and linking that as well.

cprice404: let's not put validation code in the clients for this for now; if they change something on the server we don't want to have to ship a new release of the SDK. docstrings + ticket is fine

packages/client-sdk-nodejs/src/config/leaderboard-configurations.ts
```diff
@@ -0,0 +1,105 @@
...
+    });
+  }
+}
+
+/**
+ * Browser config provides defaults suitable for use in a web browser.
+ * @export
+ * @class Browser
+ */
+export class Browser extends LeaderboardClientConfiguration {
```
cprice404, line 69: let's delete this and just have the laptop config for initial pr.

packages/client-sdk-nodejs/src/internal/leaderboard-client.ts
```diff
@@ -60,11 +61,12 @@ export class LeaderboardClient implements InternalLeaderboardClient {
   public async leaderboardUpsert(
     cacheName: string,
     leaderboardName: string,
-    elements: Map<bigint, number>
+    elements: Map<bigint | number, number>
```
cprice404, line 64: thanks for doing this! two things:

1. I just realized - missed in my first review - that you are using Map here, and typically in JS it's more common/ergonomic for people to use `Record`. Ultimately, ideally, we probably want to support both. But for now, to start, we should just pick one and I think the one we probably want to pick by default is `Record`.
2. I'm assuming you tested whether the compiler would accept a `number` input for a `bigint` argument, and that that resulted in a compilation failure?

packages/client-sdk-web/src/config/leaderboard-configurations.ts
```diff
@@ -0,0 +1,98 @@
...
+    });
+  }
+}
+
+/**
+ * Browser config provides defaults suitable for use in a web browser.
+ * @export
+ * @class Browser
+ */
+export class Browser extends LeaderboardClientConfiguration {
```
cprice404, line 64: same comment here, let's start with just Laptop and add more as needed

cprice404 (commented):
did a pass on the latest commits, looks good, just a few minor comments left

packages/client-sdk-nodejs/src/internal/leaderboard-client.ts
anitarua: 1. Record doesn't accept bigint as a key type, so it would have to be `Record<string|number, number>`, is that acceptable? Could switch the Map version to be the same type as well.
2. Yeah, got the error: `Type 'number' is not assignable to type 'bigint'`

cprice404: re: 1, ah that makes sense. in that case let's keep what you've got here for now.

re: 2, thanks for confirming!

anitarua:
@cprice404 
Will this be merged to main? Or should it be merged to a separate branch so it's not accidentally included in a release?

cprice404:
> @cprice404 Will this be merged to main? Or should it be merged to a separate branch so it's not accidentally included in a release?

let's keep it in it's own branch for now, but we can talk about how you can share the artifact with raider. And then your next PR for this stuff later this week can target this branch, and we can merge them into main at the same time.

cprice404 (approved):

PR #952, opened 2023-09-25
fix: fixed plumbing of clients and added example

anitarua:
Follow up to #949 and addresses [#468](https://github.com/momentohq/dev-eco-issue-tracker/issues/468).

Realized I had set up the extends/implements structures incorrectly so I fixed that.

Also created the tarballs and wrote an example that uses the version installed in them.

### DIFF
--- a/examples/nodejs/cache/leaderboard.ts
+++ b/examples/nodejs/cache/leaderboard.ts
@@ -37,8 +37,10 @@ async function main() {
     offset: 10,
     count: 100,
   });
-  if (fetchByScore instanceof LeaderboardFetch.Success) {
-    console.log('Fetch by score success');
+  if (fetchByScore instanceof LeaderboardFetch.Hit) {
+    console.log('Fetch by score hit');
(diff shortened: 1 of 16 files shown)


examples/nodejs/cache/leaderboard.ts
```diff
@@ -0,0 +1,88 @@
...
+    [123n, 100],
+    [456n, 200],
+    [789n, 300],
+  ]);
+  const upsertResp = await client.leaderboardUpsert(
+    "my-cache", 
+    "my-leaderboard", 
+    elements
+  );
+  console.log("Upsert elements:", upsertResp.toString());
```
cprice404, line 22: we need to include some error handling stuff here, at least in the first few APIs if not all of them.

cprice404 (commented):
looks good! just need to extend the examples to include some of the `instanceof` checks to show how to differentiate hit/miss/error etc

cprice404 (approved):
looks great, thanks!

PR #954, opened 2023-09-27
feat: Leaderboard APIs

anitarua:
Starting the second feature branch (`leaderboard-apis-implemented`) where we'll stage all the new Leaderboard additions to the JS SDK ([#468](https://github.com/momentohq/dev-eco-issue-tracker/issues/468)).

Includes all the code that was shipped with the early deliverables, but also contains the updated response classes (`Found`/`NotFound`), leaderboard-specific options interfaces, adds `Order` option to the `GetRank` API, and renames `LeaderboardClient` to `PreviewLeaderboardClient`.

Let me know if there's anything else I should address here! Will be following up with implementations and integration tests soon.


examples/nodejs/cache/leaderboard.ts
```diff
@@ -0,0 +1,106 @@
...
+    credentialProvider: CredentialProvider.fromEnvironmentVariable({
+      environmentVariableName: 'MOMENTO_API_KEY',
+    }),
+  });
+
+  // upsert
+  const elements: Map<bigint, number> = new Map([
+    [123n, 100],
+    [456n, 200],
+    [789n, 300],
```
virratanasangpunth, line 23: nit: can we do `100.0` for score to indicate that we can also accept floating points as scores?

examples/nodejs/cache/leaderboard.ts
```diff
@@ -0,0 +1,106 @@
...
+    maxScore: 500,
+    order: SortedSetOrder.Descending,
+    offset: 10,
+    count: 100,
+  });
+  if (fetchByScore instanceof LeaderboardFetch.Hit) {
+    console.log('Fetch by score hit');
+  } else if (fetchByScore instanceof LeaderboardFetch.Miss) {
+    console.log('Fetch by score miss');
+  } else if (fetchByScore instanceof LeaderboardFetch.Error) {
```
virratanasangpunth, line 44: did you forget to rename these to Found/NotFound?

examples/nodejs/cache/leaderboard.ts
```diff
@@ -0,0 +1,106 @@
...
+    console.log('Upsert success');
+  } else if (upsertResp instanceof LeaderboardUpsert.Error) {
+    console.log('Upsert error:', upsertResp.message());
+  }
+
+  // fetch by score
+  const fetchByScore = await client.leaderboardFetchByScore('my-cache', 'my-leaderboard', {
+    minScore: 100,
+    maxScore: 500,
+    order: SortedSetOrder.Descending,
```
virratanasangpunth, line 36: whoops wrong order.

virratanasangpunth (commented):
LGTM.

Looks like the exapmle needs to beupdated with the new method signatures.

packages/core/src/utils/cache-call-options.ts
```diff
@@ -120,3 +120,70 @@ export interface ListFetchCallOptions {
...
+   */
+  minScore?: number;
+  /**
+   * The maximum score of the elements to return, inclusive.
+   * If the maximum score is not specified, the range extends to the highest score.
+   */
+  maxScore?: number;
+  /**
+   * The order in which to return the elements.
+   * If the order is not specified, the elements are returned in ascending order.
```
anitarua, line 174: Currently updating these comments. Negative ranks probably don't make sense, right?

virratanasangpunth (approved):
Spongebob Many Thumbs

The method signatures look like what I'm expecting. 

I can't really comment on how the classes/interfaces design, but it looks like it's following the prior pattern.

PR #956, opened 2023-09-29
feat: add all new Leaderboard apis

anitarua:
Addresses [#468](https://github.com/momentohq/dev-eco-issue-tracker/issues/468), adds all the new Leaderboard API implementations all at once.
Currently targeted against the feature branch.


packages/client-sdk-nodejs/tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
-    "target": "ES2018",
+    "target": "ES2022",
     "module": "commonjs",
-    "lib": ["es2018", "dom"],
+    "lib": ["es2022", "dom"],
```
cprice404, line 5: this is probably fine but just curious what prompted the upgrade

packages/common-integration-tests/src/leaderboard-client.ts
```diff
@@ -0,0 +1,1062 @@
...
+        {
+          id: 789n,
+          score: 300,
+          rank: 2n,
+        },
+      ];
+      expect(receivedElements).toEqual(expectedElements);
+    });
+
+    it('inserts new elements into an existing leaderboard', async () => {
```
cprice404, line 129: in general, you want any of these `it` blocks to be a self-contained test that can be run independently. it looks like you are assuming some previous state in this one. not a blocker for this PR but a good thing to fast-follow on and a good best practice for future tests.

cprice404 (commented):
looks good to me! one question about the js lib target update and a nit about test cases being self-contained (but that can be dealt with in a follow-up, not a blocker)

packages/client-sdk-nodejs/tsconfig.json
anitarua: I can't use the BigInt() constructor for targets below ES2020 apparently. I figured I'd update it to a more recent target

cprice404: we have to be careful about whether this has implications for our users, though; does this mean that people can't use us in their own projects if they are not also targeting this platform?

@bruuuuuuuce can you weigh in here?

bruuuuuuuce: I don't think changing this lib is a problem, but I think changing the `target` might be an issue. That is saying we are targeting to compile down to es2022, which I believe means it might not work in nodejs 16. We might need to add a bigint pollyfill for older node versions which don't support

bruuuuuuuce: https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping

looks like es2022 targets node version 20

anitarua: Okay. Also I was mistaken, I can still use `BigInt()` constructor with target set to ES2018, I just can't use BigInt literals like `123n`

bruuuuuuuce: That makes sense, I looked up bigint and it was released in version 10.4, which is target `es2018`. So I think we should be good with keeping it as is

anitarua: Okay I'll switch them both back to `es2018` and just switch to using the constructor

bruuuuuuuce: 💯

packages/common-integration-tests/src/leaderboard-client.ts
```diff
@@ -0,0 +1,1072 @@
...
+      //     score: 250.0,
+      //     rank: BigInt(2),
+      //   },
+      //   {
+      //     id: BigInt(456),
+      //     score: 500.0,
+      //     rank: BigInt(3),
+      //   },
+      // ];
+      // expect(receivedWithOffsetAndCount).toEqual(expectedWithOffsetAndCount);
```
anitarua, line 415: This test is currently failing

packages/common-integration-tests/src/leaderboard-client.ts
```diff
@@ -0,0 +1,1072 @@
...
+    //     BigInt(123),
+    //     {order: LeaderboardOrder.Descending}
+    //   );
+    //   expectWithMessage(() => {
+    //     expect(getRankDescending).toBeInstanceOf(LeaderboardGetRank.Found);
+    //   }, `expected Found but got ${getRankDescending.toString()}`);
+    //   const receivedRankDesc = getRankDescending as LeaderboardGetRank.Found;
+    //   expect(receivedRankDesc.rank()).toEqual(BigInt(2));
+    // });
+  });
```
anitarua, line 807: These tests were also failing and I wasn't sure why

cprice404 (approved):
@anitarua I'm good with this now. I'm also good with you merging to `main` whenever you and Vir/Tyler are on the same page, we are close enough to the finish line and we added the word `Preview` to the class name so I think it's fine.

Sync up with them and y'all can decide whether to wait for commented tests to pass or merge this sooner and fix the tests in follow-up PRs.

anitarua:
@virratanasangpunth @tylerburdsall 
Heads up about the GetByScore and GetRank tests that are failing (above). Going to open a PR to merge Leaderboard stuff into main after merging this one. Should I wait to merge that upcoming PR until all integration tests pass?

packages/common-integration-tests/src/leaderboard-client.ts
```diff
@@ -0,0 +1,1072 @@
...
+    // it('returns NotFound when leaderboard does not exist', async () => {
+    //   const response = await leaderboardClient.leaderboardGetRank(
+    //     integrationTestCacheName,
+    //     leaderboardName,
+    //     BigInt(123)
+    //   );
+    //   expectWithMessage(() => {
+    //     expect(response).toBeInstanceOf(LeaderboardGetRank.NotFound);
+    //   }, `expected NotFound but got ${response.toString()}`);
+    // });
```
tylerburdsall, line 762: Error code 14 is gRPC not found, so isn't that expected? The SDK should transform this to a `NotFound`

packages/common-integration-tests/src/leaderboard-client.ts
tylerburdsall: Dumb question, is it possible there is a collision of elements when using `upsert` in different tests that are sharing the same leaderboard name?

packages/common-integration-tests/src/leaderboard-client.ts
anitarua: Oh okay, yeah, I forgot to add that check in the SDK code then. Will fix that now

anitarua:
Should be a different leaderboard name for each `describe` suite of tests. I used `const leaderboardName = "leaderboard-${v4()}";` at the beginning of each `describe` block.

packages/common-integration-tests/src/leaderboard-client.ts
anitarua: Wait, isn't error code 5 grpc not found? https://grpc.github.io/grpc/core/md_doc_statuscodes.html

tylerburdsall: Oh you're right, sorry, I totally misread my small font on my screen 🙈 Okay if this is getting an error code 14 that means we're doing something wrong, let me confirm some behavior and see what's yielding this bug on the back end

tylerburdsall: Hmmm, local testing passed for me 🤔

tylerburdsall: Ah, I see the issue in our metrics. I can dig deeper into this

PR #957, opened 2023-09-29
feat: add PreviewLeaderboardClient and Leaderboard APIs

anitarua:
Addresses [#468](https://github.com/momentohq/dev-eco-issue-tracker/issues/468), adds the first iteration of PreviewLeaderboardClient and Leaderboard APIs.

Decided we'll use u32 for IDs and f32 for scores on the backend, which means IDs and scores are just numbers now in the SDK. 

Re-abstracted some interfaces to build in some forward compatibility if/when we support other score data types (e.g. f64). Then the things currently named just "Leaderboard" can become aliases for "FloatLeaderboard" implementations in the next iteration.


packages/core/src/messages/responses/leaderboard/leaderboard-get-rank.ts
```diff
@@ -0,0 +1,99 @@
...
+    return this._rank;
+  }
+
+  /**
+   * Returns the score of the requested element in the leaderboard
+   * @returns {number}
+   */
+  public score(): number {
+    return this._score;
+  }
```
anitarua, line 64: This is different from the sorted set GetRank response class, but from the sample integration code, it seemed expected that the user would be able to get id, score, and rank for the requested element, not just rank.

packages/common-integration-tests/src/leaderboard-client.ts
```diff
@@ -0,0 +1,1072 @@
...
+      //     score: 250.0,
+      //     rank: BigInt(2),
+      //   },
+      //   {
+      //     id: BigInt(456),
+      //     score: 500.0,
+      //     rank: BigInt(3),
+      //   },
+      // ];
+      // expect(receivedWithOffsetAndCount).toEqual(expectedWithOffsetAndCount);
```
anitarua, line 415: @tylerburdsall @virratanasangpunth 
Thanks for addressing the GetRank bugs!

This GetByScore test is still failing though. Are unbounded min and max implemented for this API? 
Tried using Postman to make the same request and I also received an empty array when I was expecting to receive 2 elements.

tylerburdsall: > Are unbounded min and max implemented for this API?

yes, this is implemented. If you set the offset to `0` and the `count` to something like `99`, what are the results?

anitarua: I still get `"elements": []`

tylerburdsall: Hmm that's odd, by default we should be handing this. But you're right @anitarua, something is happening here and I reproduced this locally. I will dig into this

tylerburdsall: Ah, found the source of the bug! I'll start pushing the fix on our back end

tylerburdsall: This should be fixed now!

examples/nodejs/cache/README.md
```diff
@@ -70,6 +70,23 @@ This example demonstrates how to use the `batchSet`, `batchGet`, and `batchDelet
 MOMENTO_API_KEY=<YOUR API KEY> npm run batchutils
 ```
 
+## Running the Leaderboard Example
```
anitarua, line 73: Will fast-follow the SDK release with a PR to update the example file and these instructions

examples/nodejs/cache/leaderboard.ts.example
```diff
@@ -0,0 +1,134 @@
...
+  });
+
+  const createCacheResponse = await cacheClient.createCache('my-cache');
+  if (createCacheResponse instanceof CreateCache.AlreadyExists) {
+    console.log('cache already exists');
+  } else if (createCacheResponse instanceof CreateCache.Error) {
+    throw createCacheResponse.innerException();
+  }
+
+  const client = new PreviewLeaderboardClient({
```
anitarua, line 34: will update the example before the bug bash

packages/client-sdk-nodejs/src/internal/leaderboard-client.ts
```diff
@@ -0,0 +1,548 @@
...
+    leaderboardName: string
+  ) {
+    try {
+      validateCacheName(cacheName);
+      validateLeaderboardName(leaderboardName);
+    } catch (err) {
+      throw new InvalidArgumentError(
+        'cache name and leaderboard name must not be empty strings'
+      );
+    }
```
anitarua, line 70: Is it reasonable to validate the cache and leaderboard names in the constructor and throw an error? Or does that make for an unexpected UX?

packages/core/src/internal/clients/leaderboard/InternalLeaderboardClient.ts
```diff
@@ -0,0 +1,126 @@
...
+  LeaderboardRemoveElements,
+  LeaderboardUpsert,
+} from '../../../messages/responses/leaderboard';
+import {
+  LeaderboardFetchByRankCallOptions,
+  LeaderboardFetchByScoreCallOptions,
+  LeaderboardGetRankCallOptions,
+} from '../../../utils/cache-call-options';
+
+export interface InternalLeaderboardClient {
```
anitarua, line 14: I'm least certain about this interface. Should it be more like the [AbstractCacheClient](https://github.com/momentohq/client-sdk-javascript/blob/1c6a68b912734aa92650ac0728df37afffbcbd28/packages/core/src/internal/clients/cache/AbstractCacheClient.ts) class setup for CacheClients?

packages/core/src/clients/ILeaderboardClient.ts
```diff
@@ -0,0 +1,20 @@
+import {InternalLeaderboardClient} from '../internal/clients';
+import {
+  LeaderboardFetchByRankCallOptions,
+  LeaderboardFetchByScoreCallOptions,
+  LeaderboardGetRankCallOptions,
+  LeaderboardOrder,
+} from '../utils';
+
+export interface ILeaderboardClient {
+  createLeaderboard(
```
cprice404, line 10: I think let's just call this `leaderboard`; it may already exist, and the word `create` implies some things.

packages/client-sdk-nodejs/src/internal/leaderboard-client.ts
cprice404: I think that's fine.

packages/client-sdk-nodejs/src/preview-leaderboard-client.ts
```diff
@@ -0,0 +1,35 @@
...
+    this.props = props;
+  }
+
+  /**
+   * Creates an instance of LeaderboardClient with 32-bit float scores.
+   */
+  public createLeaderboard(
+    cacheName: string,
+    leaderboardName: string
+  ): InternalLeaderboardClient {
```
cprice404, line 32: This class and this function are public; that means that their return types also need to be public. So you will want to create a public interface for this return type, most likely called `ILeaderboard`.

If you want to make it so that `InternalLeaderboardClient` implements that interface for now, that's fine. But that's a private type, so we shouldn't leak it across the public API boundary.

Also, one thing to keep in mind for the future - doesn't have to be addressed before bug bash or before we share with Jah - but we are currently creating the gRPC channel in the LeaderboardDataClient. That means that with your current implementation, we will create a new gRPC channel for each `leaderboard`, which probably isn't necessary. It would probably be more efficient to keep the internal leaderboard client implementation more closely aligned with the proto-generated code, where all of its signatures accept the cache name and the leaderboard name. Then in this class you would only keep one instance of that internal client around.

In that model you would have a (possibly private/internal) class called `Leaderboard` or `MomentoLeaderboard`, which would provide a concrete implementation of `ILeaderboard`. When you called its constructor you would pass in the cache name and leaderboard name, as well as the internal LeaderboardDataClient, and it would save a reference to each of them. That way we are able to share the single gRPC connection across multiple instances of `ILeaderboard`.

Let's get ready for the bug bash first - so as long as you make the public-facing API changes for now, we are good - but maybe consider a fast follow to tweak this for connection management as soon as we think the API has settled.

packages/core/src/internal/clients/leaderboard/InternalLeaderboardClient.ts
cprice404: See my comments on the concrete client class, they may be relevant here.

IMO this interface looks like exactly what I was expecting to see for `ILeaderboard`, so I would rename it to that, move it to a public folder, and then make it the return type for the public `.leaderboard()` function.

You can still make the concrete internal client class implement this for now if it makes it easier to get the API into the shape we are expecting for the bug bash.

packages/core/src/messages/responses/leaderboard/leaderboard-fetch.ts
```diff
@@ -0,0 +1,83 @@
...
+  constructor(elements: _RankedElement[]) {
+    super();
+    this._elements = elements;
+  }
+
+  /**
+   * Returns the elements as an array of objects, each containing an `id`, `score`, and `rank` field.
+   * @returns {{id: bigint, score: number, rank: number}[]}
+   */
+  public valueArray(): {id: number; score: number; rank: number}[] {
```
cprice404, line 44: I think let's just call this `values` for now; we can add aliases for it if we end up supporting different return types for this in the future.

packages/core/src/messages/responses/leaderboard/leaderboard-fetch.ts
```diff
@@ -0,0 +1,83 @@
...
+  private readonly _elements: _RankedElement[];
+
+  constructor(elements: _RankedElement[]) {
+    super();
+    this._elements = elements;
+  }
+
+  /**
+   * Returns the elements as an array of objects, each containing an `id`, `score`, and `rank` field.
+   * @returns {{id: bigint, score: number, rank: number}[]}
```
virratanasangpunth, line 42: Looks like we need to update the doc to `number` too.

cprice404 (approved):

PR #964, opened 2023-10-10
chore: run vector index doc examples

nand4011:
Run the vector index doc examples as a part of the build.

Add a vector index client instantiation example.


malandis (approved):
🖖

PR #966, opened 2023-10-12
feat: leaderboard bug bash fixes, updated response classes

anitarua:
[Ticket #468](https://github.com/momentohq/dev-eco-issue-tracker/issues/468). Addresses feedback we got from the bug bash.

Changes include:
* Removes the `Found` and `NotFound` response classes which were confusing. 
    * `Error` is returned when the cache is not found and thus the request failed. 
    * `Success` is returned when the cache was found, but does not necessarily mean that requested elements were found (e.g. fetching data from empty or non-existent leaderboards will return length=0 or elements=[] rather than `NotFound` now).
* Removes the `leaderboard` prefixes from the Leaderboard methods.
    * E.g. `leaderboardUpsert` --> `upsert`
* Upsert can now accept elements of type `Record<number, number> | Map<number, number>` instead of just `Map<number, number>`.
* `fetchByRank` now requires `startRank` and `endRank` for pagination through leaderboards.

Will fast follow to update the nodejs example and doc example snippets after this is released. Those haven't made it into the dev docs yet.

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/leaderboard-data-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/leaderboard-data-client.ts
@@ -32,7 +32,6 @@ import {
   Interceptor,
   Metadata,
   ServiceError,
-  status,
 } from '@grpc/grpc-js';
 import {version} from '../../package.json';
 import {ILeaderboardDataClient} from '@gomomento/sdk-core/dist/src/internal/clients/leaderboard/ILeaderboardDataClient';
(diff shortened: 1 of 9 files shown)


packages/client-sdk-nodejs/src/internal/leaderboard-data-client.ts
```diff
@@ -120,43 +119,51 @@ export class LeaderboardDataClient implements ILeaderboardDataClient {
...
+        convertedElements.push(
+          new _Element({id: Number(element[0]), score: element[1]})
+        )
+      );
+    }
     return convertedElements;
   }
 
-  public async leaderboardUpsert(
+  public async upsert(
```
cprice404, line 140: this class is internal, so it's fine to change the name here, but not crucial.

I don't see the public classes in this PR?

packages/core/src/internal/clients/leaderboard/AbstractLeaderboard.ts
```diff
@@ -38,10 +38,10 @@ export abstract class AbstractLeaderboard implements ILeaderboard {
    * {@link LeaderboardUpsert.Success} on success.
    * {@link LeaderboardUpsert.Error} on failure.
    */
-  public async leaderboardUpsert(
-    elements: Map<number, number>
+  public async upsert(
```
anitarua, line 41: Public classes are in AbstractLeaderboard.ts
These are all inherited by the `Leaderboard` class in each package.

packages/core/src/internal/clients/leaderboard/AbstractLeaderboard.ts
```diff
@@ -89,28 +88,31 @@ export abstract class AbstractLeaderboard implements ILeaderboard {
...
    * @returns {Promise<LeaderboardFetch.Response>} -
-   * {@link LeaderboardFetch.Found} containing the requested elements.
-   * {@link LeaderboardFetch.NotFound} when requested elements were not found.
+   * {@link LeaderboardFetch.Success} containing the requested elements.
    * {@link LeaderboardFetch.Error} on failure.
    */
-  public async leaderboardFetchByRank(
+  public async fetchByRank(
+    startRank: number,
+    endRank: number,
```
cprice404, line 108: I hadn't thought about how making these required would cause us to pull them up to positional arguments, or else make the `options` argument required.

What you did here makes sense, but the one downside is that because JS doesn't allow callers to use keyword-style call patterns for positional arguments, you can no longer say `startRank: blah` in your call to these, it'll just be `fetchByRank(12, 42)`, which isn't quite as readable.

but it is more succinct. and it seems fairly intuitive. so i think i'm on board with doing it this way, vs. keeping these inside the options interface and just making the options arg required.

@bruuuuuuuce I am curious if you have any opinion on this.

cprice404 (commented):
i think this lgtm except I think we maybe missed removing the `leaderboard` prefix in the function names on the public client classes?

packages/core/src/internal/clients/leaderboard/AbstractLeaderboard.ts
cprice404: ooohhhh right and the PreviewClient classes just have `.leaderboard()`. Okay you're right, my bad.

packages/core/src/internal/clients/leaderboard/AbstractLeaderboard.ts
bruuuuuuuce: :skeletor_fist_shake: nodejs 😠 . 

Generally speaking, I like to use objects as params to my functions in js for this exact reason. However,

- this is more concise
- this is the pattern we currently have set for our other methods
- most js sdks just use positional args like this [like the redis sdk](https://github.com/redis/node-redis#basic-example)

I think its ok to ship with this

cprice404 (approved):
i'm on board with this. i'd be fine with you merging it. i'm interested in Matt's opinion about whether startRank/endRank live in the `options` arg or not but i would be okay with fast-follow on that if we decided to change it, and I think i'm in favor of the way you have it now.

PR #968, opened 2023-10-16
feat: add token id optional field to disposable tokens

pratik151192:
Added token id as optional field (within an optional props) to disposable tokens API. The object `DisposableTokenScope` is encapsulated with all kinds of permission policies that we can provide, and it didn't seem clean to add a `tokenID` field there. Instead, added as a new props for that API.

PR #1 for https://github.com/momentohq/dev-eco-issue-tracker/issues/511

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/internal-auth-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/internal-auth-client.ts
@@ -10,6 +10,7 @@ import {
   validateDisposableTokenExpiry,
   validateValidForSeconds,
   validateCacheKeyOrPrefix,
+  validateDisposableTokenTokenID,
 } from '@gomomento/sdk-core/dist/src/internal/utils';
 import Never = grpcAuth._GenerateApiTokenRequest.Never;
 import Expires = grpcAuth._GenerateApiTokenRequest.Expires;
@@ -54,6 +55,7 @@ import {
   asDisposableTokenCachePermission,
   asDisposableTokenPermissionsObject,
   DisposableTokenCachePermission,
+  DisposableTokenProps,
   isDisposableTokenCachePermission,
   isDisposableTokenPermissionsObject,
 } from '@gomomento/sdk-core/dist/src/auth/tokens/disposable-token-scope';
@@ -186,7 +188,7 @@ export class InternalAuthClient implements IAuthClient {
   public async generateDisposableToken(
     scope: DisposableTokenScope,
     expiresIn: ExpiresIn,
-    tokenId?: string
+    disposableTokenProps?: DisposableTokenProps
   ): Promise<GenerateDisposableToken.Response> {
     const tokenClient = new token.token.TokenClient(
       this.creds.getTokenEndpoint(),
@@ -209,11 +211,22 @@ export class InternalAuthClient implements IAuthClient {
       return new GenerateDisposableToken.Error(normalizeSdkError(err as Error));
     }
 
+    const tokenID = disposableTokenProps?.tokenID;
+    if (tokenID) {
+      try {
+        validateDisposableTokenTokenID(tokenID);
+      } catch (err) {
+        return new GenerateDisposableToken.Error(
+          normalizeSdkError(err as Error)
+        );
+      }
+    }
+
     const request = new token.token._GenerateDisposableTokenRequest({
       expires: expires,
       auth_token: this.creds.getAuthToken(),
       permissions: permissions,
-      token_id: tokenId,
+      token_id: tokenID,
     });
 
     return await new Promise<GenerateDisposableToken.Response>(resolve => {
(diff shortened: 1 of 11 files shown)


pratik151192:
@cprice404 I notice that the method `generateAPIKey` too accepts a `tokenID` but I didn't club it in this PR as the ticket was scoped to the `disposableToken` API. LMK if you want me to club here though.

packages/core/src/internal/utils/validators.ts
```diff
@@ -145,6 +145,14 @@ export function validateDisposableTokenExpiry(expiresIn: ExpiresIn) {
   }
 }
 
+export function validateDisposableTokenTokenID(tokenID: string) {
+  if (tokenID.length > 64) {
+    throw new InvalidArgumentError(
+      'TokenID must be less than or equal to 64 characters.'
+    );
+  }
```
pratik151192, line 153: I had thought to `encode` the string and check the byte length in UTF-8 or something so lmk if this seems less potent that that.

packages/client-sdk-nodejs/src/internal/internal-auth-client.ts
```diff
@@ -208,10 +211,22 @@ export class InternalAuthClient implements IAuthClient {
...
+          normalizeSdkError(err as Error)
+        );
+      }
+    }
+
     const request = new token.token._GenerateDisposableTokenRequest({
       expires: expires,
       auth_token: this.creds.getAuthToken(),
       permissions: permissions,
+      token_id: tokenID,
```
anitarua, line 229: can the proto accept an undefined tokenID?

anitarua (dismissed):
lgtm!

pratik151192:
@anitarua @cprice404 unless am seeing it wrong, looks like the last 3 CIs failed on one or more tests unrelated to the PR. This particularly started happening after I added 2 new auth tests which don't seem to be related to integ tests. But would love another eye; I am afraid if this might be another Azure tipping point for something

cprice404:
> @anitarua @cprice404 unless am seeing it wrong, looks like the last 3 CIs failed on one or more tests unrelated to the PR. This particularly started happening after I added 2 new auth tests which don't seem to be related to integ tests. But would love another eye; I am afraid if this might be another Azure tipping point for something

this seems really different from prev failures in that it's data plane (afaik all the azure things happened on control plane), plus if this is running on osx i'm not sure it's even using azure. Did you spot check the server metrics?

packages/client-sdk-nodejs/src/internal/internal-auth-client.ts
```diff
@@ -208,10 +211,22 @@ export class InternalAuthClient implements IAuthClient {
       return new GenerateDisposableToken.Error(normalizeSdkError(err as Error));
     }
 
+    const tokenID = disposableTokenProps?.tokenID;
+    if (tokenID) {
```
cprice404, line 215: javascript has very liberal definitions for "truthiness", so it's always better to be very explicit about the condition you are checking for. In this case I think it would be `!== undefined`.

packages/client-sdk-web/src/internal/auth-client.ts
```diff
@@ -213,6 +216,18 @@ export class InternalWebGrpcAuthClient<
 
     request.setPermissions(permissions);
 
+    const tokenID = disposableTokenProps?.tokenID;
+    if (tokenID) {
```
cprice404, line 220: same nit about explicit condition check here

packages/client-sdk-nodejs/src/internal/internal-auth-client.ts
```diff
@@ -208,10 +211,22 @@ export class InternalAuthClient implements IAuthClient {
       return new GenerateDisposableToken.Error(normalizeSdkError(err as Error));
     }
 
+    const tokenID = disposableTokenProps?.tokenID;
```
cprice404, line 214: supernit: I think the idiomatic casing for this variable name in TS/JS would be `tokenId`. I don't really care about it here for a local var, but we want to be deliberate in the props object since that is effectively public API.

packages/core/src/auth/tokens/disposable-token-scope.ts
```diff
@@ -77,6 +77,10 @@ export type DisposableTokenScope =
   | PredefinedScope
   | DisposableTokenCachePermissions;
 
+export interface DisposableTokenProps {
+  tokenID?: string;
```
cprice404, line 81: here - I think we want `tokenId` for this public interface unless you have some style reference that suggests otherwise

packages/core/src/internal/utils/validators.ts
cprice404: I think this is fine

cprice404 (commented):
just a few nits!

anitarua (approved):
lgtm!

cprice404 (approved):

PR #970, opened 2023-10-17
chore: cloudwatch metrics example

anitarua:
Addresses [#442](https://github.com/momentohq/dev-eco-issue-tracker/issues/442). Created a new `lambda-examples` directory to show how a CDK stack can be configured to deploy a Lambda function or ECS cluster, create metric filters for the associated log groups (or a custom log group), and create a CloudWatch dashboard. 

The dashboard looks something like this when populated (updated dashboard):
cloudwatch dashboard

The dashboard uses values in the logs emitted by both the [Nodejs SDK metrics](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/config/middleware/impl/experimental-metrics-middleware.ts#L26) and [.NET SDK metrics](https://github.com/momentohq/client-sdk-dotnet/blob/2900922d8bd46ac3f91cb06dbfe070419dc9b738/src/Momento.Sdk/Config/Middleware/ExperimentalMetricsMiddleware.cs#L18) interfaces.

Also tested that this works with a .NET lambda function (that work currently exists in an [unmerged branch](https://github.com/anitarua/client-sdk-dotnet/tree/dotnet-lambda-example) but can merge into the dotnet repo if it seems useful).

Let me know if you think the dashboard should include/exclude any other metrics!

Original dashboard when PR opened:
metrics example dashboard


examples/nodejs/lambda-examples/cloudwatch-metrics/README.md
```diff
@@ -0,0 +1,105 @@
+<head>
```
bruuuuuuuce, line 1: Your readme/docs are always on point, nice job 🤜

examples/nodejs/lambda-examples/cloudwatch-metrics/infrastructure/lib/momento-metrics-stack.ts
```diff
@@ -0,0 +1,310 @@
...
+    const logGroup = new LogGroup(this, 'Logs', {
+      logGroupName: logGroupName,
+      retention: RetentionDays.ONE_DAY,
+    });
+
+    switch (stackConfig) {
+      case exampleApp.NodejsLambda: {
+        const nodejsLambda = new lambdaNodejs.NodejsFunction(this, 'MomentoMetricsMiddlewareCDKExample', {
+          functionName: 'MomentoMetricsMiddlewareCDKExample',
+          runtime: lambda.Runtime.NODEJS_16_X,
```
bruuuuuuuce, line 62: nit: we should be using nodejs 18 or 20, nodejs16 is EOL

examples/nodejs/lambda-examples/cloudwatch-metrics/infrastructure/lib/momento-metrics-stack.ts
```diff
@@ -0,0 +1,310 @@
...
+        taskDefinition.addToTaskRolePolicy(new iam.PolicyStatement({
+          effect: iam.Effect.ALLOW,
+          actions: [
+            "secretsmanager:ListSecrets",
+            "secretsmanager:GetSecretValue",
+            "logs:PutLogEvents",
+            "logs:CreateLogStream",
+            "ecs:*"
+          ],
+          resources: ["*"],
```
bruuuuuuuce, line 100: nit: we know the secret name for the momento api key above. We should restrict listing/getting secret values to only that secret

examples/nodejs/lambda-examples/cloudwatch-metrics/infrastructure/lib/momento-metrics-stack.ts
```diff
@@ -0,0 +1,310 @@
...
+        taskDefinition.addToExecutionRolePolicy(new iam.PolicyStatement({
+          effect: iam.Effect.ALLOW,
+          actions: [
+            "secretsmanager:ListSecrets",
+            "secretsmanager:GetSecretValue",
+            "logs:PutLogEvents",
+            "logs:CreateLogStream",
+            "ecs:*"
+          ],
+          resources: ["*"],
```
bruuuuuuuce, line 111: ditto

examples/nodejs/lambda-examples/cloudwatch-metrics/lambda/handler.ts
```diff
@@ -0,0 +1,86 @@
...
+export const handler = async () => {
+  try {
+    const loggerFactory = new DefaultMomentoLoggerFactory(DefaultMomentoLoggerLevel.DEBUG);
+    const logger = loggerFactory.getLogger('CloudWatchMetricsMiddlewaresExample');
+
+    const cacheClient = await getCacheClient(loggerFactory);
+    logger.info('Created a CacheClient configured with metrics middleware');
+
+    logger.info('Issuing 5 minutes of set and get requests to generate data for the dashboard example');
+    for (let i = 0; i < (60 /* seconds */ * 5 /* minutes */); i++) {
```
bruuuuuuuce, line 24: momento api calls should be fast so this should probably not matter, but just calling out that this isnt actually 5 min of gets and sets. Its 300 gets and sets plus a 300 1-second delays. So its 300 x (1 + epsilon) seconds.

examples/nodejs/lambda-examples/cloudwatch-metrics/lambda/handler.ts
```diff
@@ -0,0 +1,86 @@
...
+} from '@gomomento/sdk';
+import {ExperimentalMetricsLoggingMiddleware} from '@gomomento/sdk/dist/src/config/middleware/experimental-metrics-logging-middleware';
+
+const _secretsClient = new SecretsManagerClient({});
+const _cachedSecrets = new Map<string, string>();
+let _cacheClient: CacheClient | undefined = undefined;
+
+export const handler = async () => {
+  try {
+    const loggerFactory = new DefaultMomentoLoggerFactory(DefaultMomentoLoggerLevel.DEBUG);
```
bruuuuuuuce, line 17: do we log a ton when debug logging? If we do we might want to make this `INFO`. If not then `DEBUG` is fine

bruuuuuuuce: O is the `DEBUG` for the middleware, so we can grep the logs and make a dashboard

examples/nodejs/lambda-examples/cloudwatch-metrics/infrastructure/bin/infrastructure.ts
```diff
@@ -0,0 +1,17 @@
...
+new MomentoMetricsStack(app, 'MomentoMetricsMiddlewareExample', {
+  /* If you don't specify 'env', this stack will be environment-agnostic.
+   * Account/Region-dependent features and context lookups will not work,
+   * but a single synthesized template can be deployed anywhere. */
+  /* Uncomment the next line to specialize this stack for the AWS Account
+   * and Region that are implied by the current CLI configuration. */
+  // env: { account: process.env.CDK_DEFAULT_ACCOUNT, region: process.env.CDK_DEFAULT_REGION },
+  /* Uncomment the next line if you know exactly what Account and Region you
+   * want to deploy the stack to. */
+  // env: { account: '123456789012', region: 'us-east-1' },
```
bruuuuuuuce, line 15: nit: looks like we are hardcoding the region here to `us-west-2`, so even if a user passes in a different env, we will still have resources in west-2. I don't think it would be hard to change this, or we can just remove this text and say it will only deploy to `us-west-2`

examples/nodejs/lambda-examples/cloudwatch-metrics/infrastructure/lib/momento-metrics-stack.ts
```diff
@@ -0,0 +1,310 @@
...
+      case exampleApp.Custom: {
+        console.log('Skipping to dashboard creation');
+        break;
+      }
+      default: {
+        throw new Error('Unimplemented CDK stack application');
+      }
+    }
+
+    logGroup.addMetricFilter('ExampleMetricFilterDuration', {
```
bruuuuuuuce, line 129: This is cool I didn't know you could do this

examples/nodejs/lambda-examples/cloudwatch-metrics/docker/ecs-code/index.ts
```diff
@@ -0,0 +1,77 @@
...
+
+async function getCacheClient(loggerFactory: DefaultMomentoLoggerFactory): Promise<CacheClient> {
+  const apiKeySecretName = process.env.MOMENTO_API_KEY_SECRET_NAME;
+  if (apiKeySecretName === undefined) {
+    throw new Error("Missing required env var 'MOMENTO_API_KEY_SECRET_NAME");
+  }
+  if (_cacheClient === undefined) {
+    const momentoApiKey = await getSecret(apiKeySecretName);
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call,@typescript-eslint/no-unsafe-assignment
```
cprice404, line 44: just a nit but curious why this eslint hint is necessary here?

examples/nodejs/lambda-examples/cloudwatch-metrics/docker/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM --platform=linux/amd64 public.ecr.aws/docker/library/node:latest
+
+WORKDIR /usr/src/app
+
+COPY ecs-code/index.ts ./
+COPY ecs-code/package.json ./
+RUN npm install
+
+ENV MOMENTO_API_KEY_SECRET_NAME="MomentoMetricsApiKey"
+ENV AWS_REGION="us-west-2"
```
cprice404, line 10: what happens if you leave this out?

examples/nodejs/lambda-examples/cloudwatch-metrics/infrastructure/bin/infrastructure.ts
cprice404: good catch @bruuuuuuuce . let's see if we can get rid of the hard-coded us-west-2 refs

examples/nodejs/lambda-examples/cloudwatch-metrics/infrastructure/lib/momento-metrics-stack.ts
```diff
@@ -0,0 +1,310 @@
...
+      [exampleApp.Custom, customLogGroupName]
+    ])
+    const logGroupName = configToLogGroupName.get(stackConfig);
+
+    const logGroup = new LogGroup(this, 'Logs', {
+      logGroupName: logGroupName,
+      retention: RetentionDays.ONE_DAY,
+    });
+
+    switch (stackConfig) {
```
cprice404, line 58: I would love to move all of the stuff that has to do with deploying the lambda/ecs examples to a few smaller, purpose-built functions so that they don't distract from the flow of readability for this main code.

examples/nodejs/lambda-examples/cloudwatch-metrics/infrastructure/lib/momento-metrics-stack.ts
cprice404, line 129: would probably be useful to add a code comment above this part, explaining what it is doing. "The Momento experimental metrics middleware produces JSON logs with metrics about each Momento request.  An example log entry looks like this: ...

In this section we use CloudWatch Metric Filters (maybe link to docs) to extract metrics from these log messages."

examples/nodejs/lambda-examples/cloudwatch-metrics/infrastructure/lib/momento-metrics-stack.ts
```diff
@@ -0,0 +1,310 @@
...
+      dashboardName: 'MomentoMetricsCDKExampleDashboard',
+      defaultInterval: cdk.Duration.hours(1),
+    });
+
+    const latency = new GraphWidget({
+      left: [
+        new Metric({
+          metricName: 'Duration (Latency)',
+          namespace: 'MomentoMetricsCDKExample',
+          region: 'us-west-2',
```
cprice404, line 170: the `cdk.Stack` object should have a variable that you can read to get the region string. I can't remember if it's `this.region` or `this.env.region` or what, but it should be available.

examples/nodejs/lambda-examples/cloudwatch-metrics/infrastructure/lib/momento-metrics-stack.ts
```diff
@@ -0,0 +1,310 @@
...
+      left: [
+        new Metric({
+          metricName: 'GRPC Status Code',
+          namespace: 'MomentoMetricsCDKExample',
+          region: 'us-west-2',
+        }),
+      ],
+      stacked: false,
+      height: 6,
+      width: 6,
```
cprice404, line 191: I suspect that some of the layout properties you are setting on some of these widgets may align with the defaults.

In the past I have found that it's easier to manage programmatic dashboards if you keep the number of layout properties that you explicitly specify to a minimum. You will definitely need some height/width/position attributes, and you may also want to keep some of your annotation stuff. Heck, maybe you need everything here :) but just wanted to call out - i would experiment with omitting anything you don't think is absolutely necessary and see if it still looks okay, and if so, omit what you can.

totally your call on which things clear the bar of being necessary though!

cprice404 (commented):
this is looking awesome!

i want to test it in my account before we merge it, and not sure if i will get to that today. but i left a handful of minor feedback that I think you could go ahead and move forward on when you have bandwidth.

examples/nodejs/lambda-examples/cloudwatch-metrics/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+The source code for the CDK application lives in the `infrastructure` directory.
+To build and deploy it you will first need to install the dependencies:
+
+```bash
+cd infrastructure
+npm install
+```
+
+You will also need to set the `stackConfig` variable in the [`momento-metrics-stack.ts` file](./infrastructure/lib/momento-metrics-stack.ts) to `exampleApp.NodejsLambda` to specify that you want to deploy the Lambda metrics example.
```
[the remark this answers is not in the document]
anitarua: Couldn't get the `stackConfig` variable working with CFN stack parameters (they're lazily-loaded tokens but the constructor needs to know the value while creating the stack). I instead moved all the config related stuff to a `.env` file and updated the README with instructions for those.

examples/nodejs/lambda-examples/cloudwatch-metrics/docker/ecs-code/index.ts
anitarua: can't recall why anymore, but I've removed the hints

examples/nodejs/lambda-examples/cloudwatch-metrics/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+This directory contains an AWS CDK stack that will create and populate a custom CloudWatch dashboard for Momento client-side metrics. You can deploy an example Lambda function or ECS cluster that uses the Momento Node.js SDK's Experimental Metrics Middleware to generate example dashboard data for 5 minutes at a time. Or you can deploy just the CloudWatch dashboard and use your own application to emit Momento metrics.
+
+## Prerequisites
+
+- Node version 18 or higher is required
+- To get started with Momento, you will need a Super User Momento API key. You can get one from the [Momento Console](https://console.gomomento.com). Check out the [getting started](https://docs.momentohq.com/getting-started) guide for more information on obtaining an API key.
+- To deploy the CDK app, you will need to have [configured your AWS credentials](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-authentication.html#cli-chap-authentication-precedence).
+
+## Deploying the Lambda Function
```
cprice404, line 21:
```suggestion
## Deploying the CloudWatch Dashboard (along with an example Momento Lambda Function)
```

examples/nodejs/lambda-examples/cloudwatch-metrics/README.md
```diff
@@ -0,0 +1,119 @@
...
+cd infrastructure
+npm install
+```
+
+Then create a `.env` file in the `infrastructure` directory, which is where you will define all required configuration variables. 
+You will need the Super User API key you generated from the [Momento Console](https://console.gomomento.com). You will also set the `STACK_CONFIG` variable to `lambda` like so:
+
+```bash
+MOMENTO_API_KEY=<YOUR_MOMENTO_API_KEY>
+STACK_CONFIG="lambda"
```
cprice404, line 45: nit - can we rename this var to something like `EXAMPLE_MOMENTO_APPLICATION` or `ADDITIONAL_APPLICATION` or something along those lines?

cprice404 (commented):
anita i deployed this locally and started playing with it, but ran out of time. I will finish reviewing this evening!

examples/nodejs/lambda-examples/cloudwatch-metrics/docker/ecs-code/index.ts
```diff
@@ -0,0 +1,73 @@
...
+async function main() {
+  try {
+    const loggerFactory = new DefaultMomentoLoggerFactory(DefaultMomentoLoggerLevel.INFO);
+    const logger = loggerFactory.getLogger('CloudWatchMetricsMiddlewaresExample');
+
+    const cacheClient = await getCacheClient(loggerFactory);
+    logger.info('Created a CacheClient configured with metrics middleware');
+
+    logger.info('Issuing 5 minutes of set and get requests to generate data for the dashboard example');
+    for (let i = 0; i < 60 /* seconds */ * 5 /* minutes */; i++) {
```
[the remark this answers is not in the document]
cprice404: actually ended up with this:

```
const delayBetweenRequestsMillis = 100;
    const delaysPerSecond = 1000 / delayBetweenRequestsMillis;
    for (let i = 0; i < 60 /* seconds */ * 60 /* minutes */ * delaysPerSecond; i++) {
      await cacheClient.set('cache', 'metrics-example-{i}', 'VALUE');
      await cacheClient.get('cache', 'metrics-example-{i}');
      await delay(delayBetweenRequestsMillis);
    }
```

examples/nodejs/lambda-examples/cloudwatch-metrics/infrastructure/lib/momento-metrics-stack.ts
```diff
@@ -0,0 +1,337 @@
...
+      }
+    }
+
+    /* Step 5:
+    The Momento experimental metrics middleware produces JSON logs with metrics about each Momento request. An example log entry looks like this:
+        (Momento: _ExperimentalMetricsLoggingMiddleware): 
+        {
+            "numActiveRequestsAtStart": 1,
+            "numActiveRequestsAtFinish": 1,
+            "requestType": "MiddlewareMessage",
```
cprice404, line 95: woof, is this what really shows up for the request type now? that seems like a bug in my last refactor. We want this field to show something like `Get`, `Set`, `ListPushBack`, etc. If all of the requests are showing up with just this, then we should make a quick tweak on `main` to fix that and then rebase your branch. I can help with that if you need another pair of eyes.

examples/nodejs/lambda-examples/cloudwatch-metrics/infrastructure/lib/momento-metrics-stack.ts
```diff
@@ -0,0 +1,337 @@
...
+
+    logGroup.addMetricFilter('ExampleMetricFilterRequestSize', {
+      metricNamespace: 'MomentoMetricsCDKExample',
+      metricName: 'Request Size (bytes)',
+      filterPattern: FilterPattern.literal('{$.requestSize >= 0}'),
+      metricValue: '$.requestSize',
+      unit: Unit.BYTES,
+    });
+
+    logGroup.addMetricFilter('ExampleMetricFilterResponseSize', {
```
cprice404, line 192: see my comments below about `dimensions` before reading this.

For this metric, adding a dimension for the request type would be extremely useful. Then we could graph p50/p99 request size / response size by request type, for a few of the most common request types.

This also would rely on fixing the bug I mentioned above about how the request type is currently showing up as 'middleware' or whatever.

examples/nodejs/lambda-examples/cloudwatch-metrics/infrastructure/lib/momento-metrics-stack.ts
```diff
@@ -0,0 +1,337 @@
...
+          metricName: 'Response Size (bytes)',
+          namespace: 'MomentoMetricsCDKExample',
+          label: 'Count of responses per second',
+          statistic: Stats.SAMPLE_COUNT,
+        }),
+      ],
+      period: cdk.Duration.seconds(1),
+      height: 6,
+      sparkline: true,
+      title: 'Transactions per second',
```
cprice404, line 274: Let's call this "requests per second".

examples/nodejs/lambda-examples/cloudwatch-metrics/infrastructure/lib/momento-metrics-stack.ts
```diff
@@ -0,0 +1,337 @@
...
+    });
+
+    const numberOfRequests = new GraphWidget({
```
cprice404, line 277: i'm curious what is the intended difference between this widget and the previous? They are both basically graphing requests per second and I'm seeing the same values on the two graphs in my dashboard.

I also don't completely understand why the values are showing up as the same, when on one you are graphing sample count over the last minute and one you are graphing sample count over the last second :) something fishy about that that i don't know the answer to. So we need to sort that out, but probably also only need one of the two of these.

Also... I'm sleeping 100ms between requests, so I would expect to see about 10tps * 2 (one get and one set) = 20 tps, but I'm seeing about 1.1k, which seems way too high, and I don't have an answer for that yet either :)

cprice404: when i look at the lightstep metrics for this I see about 9.5 tps for get and 9.5 tps for set, which is much closer to what i was expecting: https://app.lightstep.com/momento-preprod/dashboard/user-experience/TJXd7Dq4?time_window=minutes_60&chart_id=XYbgq5zQ

cprice404: I think maybe what is happening here is that the maximum resolution for metricfilter metrics is 1 minute, so even when you set the period to something < 1 minute, you still get the resolution of 1 minute. (we should google this and see if this is documented for cloudwatch metric filters.)

if that's the case, then setting the period to 1m and dividing the value by 60 via cloudwatch metricmath may be the way to go. 1100 / 60 = 18.3, which is approximately what I'm seeing in lightstep for get + set combined. (with my own sleep between requests set to 100ms.)

examples/nodejs/lambda-examples/cloudwatch-metrics/infrastructure/lib/momento-metrics-stack.ts
```diff
@@ -0,0 +1,337 @@
...
+          label: 'Number of requests per second',
+        }),
+      ],
+      title: 'Number of requests',
+      period: cdk.Duration.minutes(1),
+      statistic: Stats.SAMPLE_COUNT,
+    });
+
+    const totalBytesSentReceived = new GraphWidget({
+      left: [
```
cprice404, line 291: let's do p50 and p99 on this one. and if we are able to get the request type dimension added then I think we should do "get p50", "get p99", "set p50", "set p99".

cprice404 (commented):
this is looking really awesome. I left some comments for us to fine tune the dashboard widgets in this last mile, to make them more realistic for an operator dashboard, but this is really great and I can see the finish line!

examples/nodejs/lambda-examples/cloudwatch-metrics/infrastructure/lib/momento-metrics-stack.ts
anitarua: Good catch! Got a PR for this: #993

anitarua:
Lots of great feedback, thanks!! Here's what the dashboard looks like now

cloudwatch dashboard

cprice404 (approved):
dance

PR #971, opened 2023-10-19
chore: delete leaderboard at end of each leaderboard test

anitarua:
Each `describe` block creates a new leaderboard and I belatedly realized that each should also delete them when done testing. Also renamed the leaderboards so they all start with `test-leaderboard-` and are hopefully easier to find and manually delete, if necessary.

### DIFF
--- a/packages/common-integration-tests/src/leaderboard-client.ts
+++ b/packages/common-integration-tests/src/leaderboard-client.ts
@@ -17,7 +17,7 @@ export function runLeaderboardClientTests(
   integrationTestCacheName: string
 ) {
   describe('#Creates leaderboard client', () => {
-    const leaderboardName = `leaderboard-${v4()}`;
+    const leaderboardName = `test-leaderboard-${v4()}`;
 
     it('validates the cache name', () => {
       expect(() =>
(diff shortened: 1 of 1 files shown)


virratanasangpunth (approved):

packages/common-integration-tests/src/leaderboard-client.ts
```diff
@@ -851,5 +893,12 @@ export function runLeaderboardClientTests(
         expect(deletedLength).toEqual(0);
       }, `expected array of size 0 but got ${deletedLength}`);
     });
+
+    it('deletes the leaderboard when done testing', async () => {
+      const response = await leaderboard.delete();
+      expectWithMessage(() => {
+        expect(response).toBeInstanceOf(LeaderboardDelete.Success);
+      }, `expected Success but got ${response.toString()}`);
+    });
```
cprice404, line 902: this is the kind of thing where (in the future) it might be worth writing a higher-order function to encapsulate it... something like `withTemporaryLeaderboard(leaderboardName, () => { test code })`. something like, in addition to DRY, could help with error handling (e.g. if the test throws an exception you can still make sure the leaderboard gets deleted in the `catch` block before re-throwing exception.)

I have given the guidance to be careful about not going too crazy with DRY in cases where it makes the tests less readable/maintainable, but this is one where I think it would be a net win.

No need to change this PR, just commenting for the future.

cprice404 (approved):

PR #972, opened 2023-10-20
feat: support more vector item metadata value types

malandis:
In the previous version of the service, we only supported
string-valued metadata. In the latest version, we support the
following value types:

- string
- int
- double
- bool
- list of strings

We update the `VectorItem` and `SearchHit` interfaces to support
metadata of these types (where int and float are both represented as
`number`).

We update `sendUpsertItemBatch` to serialize the metadata
appropriately. We distinguish numbers that are integers vs
floating point.

Similarly we update `sendSearch` to deserialize the metadata
appropriately.

We add an extra integration test to ensure each of the metadata value
types are stored and retrieved properly.


packages/client-sdk-web/src/internal/vector-index-data-client.ts
```diff
@@ -70,19 +80,60 @@ export class VectorIndexDataClient implements IVectorIndexDataClient {
...
+    }
+    return Object.entries(item.metadata).map(([key, value]) => {
+      const metadata = new vectorindex._Metadata();
+      metadata.setField(key);
+      if (typeof value === 'string') {
+        metadata.setStringValue(value);
+      } else if (typeof value === 'number') {
+        if (Number.isInteger(value)) {
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+          metadata.setIntegerValue(value);
```
malandis, line 107: I am unclear why some of these modifiers have unsafe call warnings and others do not.

malandis:
I regenerated the `package-lock.json` files in https://github.com/momentohq/client-sdk-javascript/pull/972/commits/a6683b651bc7406a9f7e3616e524b9629b344391. The patches are hefty.

packages/client-sdk-nodejs/src/internal/vector-index-data-client.ts
```diff
@@ -77,37 +80,92 @@ export class VectorIndexDataClient implements IVectorIndexDataClient {
...
+      return [];
+    }
+    return Object.entries(item.metadata).map(([key, value]) => {
+      if (typeof value === 'string') {
+        return new vectorindex._Metadata({
+          field: key,
+          string_value: value,
+        });
+      } else if (typeof value === 'number') {
+        if (Number.isInteger(value)) {
```
cprice404, line 128: lol @ javascript

packages/client-sdk-web/src/internal/vector-index-data-client.ts
```diff
@@ -201,9 +252,43 @@ export class VectorIndexDataClient implements IVectorIndexDataClient {
                   id: hit.getId(),
                   distance: hit.getDistance(),
                   metadata: hit.getMetadataList().reduce((acc, metadata) => {
-                    acc[metadata.getField()] = metadata.getStringValue();
+                    const field = metadata.getField();
+                    switch (metadata.getValueCase()) {
+                      case vectorindex._Metadata.ValueCase.STRING_VALUE:
```
cprice404, line 257: curious if there is a reason you are able to use constants here but you had to use string literals for the earlier case statement

packages/client-sdk-nodejs/src/internal/vector-index-data-client.ts
```diff
@@ -212,9 +270,35 @@ export class VectorIndexDataClient implements IVectorIndexDataClient {
...
+                            new UnknownError(
+                              'Search responded with an unknown result'
+                            )
+                          )
+                        );
+                        break;
+                    }
                     return acc;
-                  }, {} as Record<string, string>),
+                  }, {} as Record<string, string | number | boolean | Array<string>>),
```
cprice404, line 301: just curious if you tried a type alias for this in JS too, and observed the same issue where the IDEs didn't expand it in a helpful way

cprice404 (approved):
minor questions / nits, no blockers

packages/client-sdk-web/src/internal/vector-index-data-client.ts
bruuuuuuuce: I pulled this down locally to test this. I don't think that you need any of these `eslint-disable` lines actually. The js protobuf dependecies trip up some IDEs. I suspect that your IDE is operating off of some stale data, and that it thinks that `item` is of type `any` or `unknown`. I think if you remove all of the `eslint-disable` lines, and rebuild in your terminal it will pass. (Verified this :onmybox: ). Sometimes if you rebuild on your terminal, _and then_ open up the project inside your IDE, it will pickup the latest changes. :skeletor_fist_shake: _caching issues_ :skeletor_fist_shake:

packages/client-sdk-web/src/internal/vector-index-data-client.ts
malandis: That appears to be a grpc-js vs grpc-web thing. grpc-js uses string literals and grpc-web constants. 🤸‍♂️

packages/client-sdk-nodejs/src/internal/vector-index-data-client.ts
malandis: Same thing in the sense that if we make a type alias `T` that is used in a function signature for `foo`, then if you type `foo` you will only see `T`, unless you manually click the function, go to definition, and then hover over `T`.

packages/client-sdk-web/src/internal/vector-index-data-client.ts
cprice404: cool. bonus points for defining our own constants for it in the near future but definitely not a blocker.

PR #975, opened 2023-10-20
chore: refactor leaderboard tests to use temporary leaderboards that get deleted

anitarua:
Follow up to #971, but uses a higher order function.


cprice404 (approved):
fantastic

PR #979, opened 2023-10-25
chore: use username as tokenId for nextjs chat examples

pratik151192:
Instead of passing a user's name involve in chat through a topic message, we are inferring it through the `tokenId` in `TopicItem`. This will be demonstrated in a blog where we can say that users can save $$ going forward by following this route.

The username is being passed around through a query string.

### DIFF
--- a/examples/web/nextjs-chat/src/app/api/momento/token/route.ts
+++ b/examples/web/nextjs-chat/src/app/api/momento/token/route.ts
@@ -11,6 +11,7 @@ import {
 } from "./config";
 import { authOptions } from "../../auth/[...nextauth]/route";
 import { getServerSession } from "next-auth";
+import { type NextRequest } from "next/server";
 
 const authClient = new AuthClient({
   credentialProvider: CredentialProvider.fromString({
@@ -20,10 +21,12 @@ const authClient = new AuthClient({
 
 export const revalidate = 0;
 
-export async function GET(_request: Request) {
+export async function GET(_request: NextRequest) {
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore
-  const usernameValue = _request.cookies._parsed.get("username").value;
+  let usernameValue: undefined | string =
+    _request.nextUrl.searchParams.get("username");
+  usernameValue = usernameValue === null ? undefined : usernameValue;
 
   if (usernameValue === undefined) {
     console.error(`Username is undefined`);
@@ -61,7 +64,7 @@ export async function GET(_request: Request) {
   throw new Error("Unable to get token from momento");
 }
 
-async function fetchTokenWithOpenAuth(username: string) {
+async function fetchTokenWithOpenAuth(username: string | undefined) {
   return await authClient.generateDisposableToken(
     tokenPermissions,
     tokenExpiresIn,
@@ -69,7 +72,7 @@ async function fetchTokenWithOpenAuth(username: string) {
   );
 }
 
-async function fetchTokenWithAuthCredentials(username: string) {
+async function fetchTokenWithAuthCredentials(username: string | undefined) {
   const session = await getServerSession(authOptions);
 
   if (!session) {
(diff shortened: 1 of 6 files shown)


examples/web/nextjs-chat/src/app/api/momento/token/route.ts
```diff
@@ -19,14 +19,27 @@ const authClient = new AuthClient({
 });
 
 export const revalidate = 0;
+
 export async function GET(_request: Request) {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  const usernameValue = _request.cookies._parsed.get("username").value;
```
pratik151192, line 26: for some reason TS doesn't think `cookies` is a part of `_request` but i tested this and it works

examples/web/nextjs-chat/src/utils/momento-web.ts
```diff
@@ -15,7 +15,7 @@ export enum EventTypes {
 
 export type ChatMessageEvent = {
   event: EventTypes.MESSAGE;
-  username: string;
+  username?: string;
```
pratik151192, line 18: This entire demo is wired around `ChatMessageEvent` so i made this optional here even though that wasn't the cleanest thing to do. Would have needed more overhaul if we wanted to change the entire wiring.

allenheltondev (approved):

examples/web/nextjs-chat/src/app/api/momento/token/route.ts
cprice404: I suspect that you can handle this via next-auth's session stuff but @bruuuuuuuce can probably provide a more clear answer.

examples/web/nextjs-chat/src/utils/momento-web.ts
pratik151192: I dont like the way i've done this at all but was the quickest without a overhaul of the entire example (i maybe wrong in that though).. right now this `type` is constructed when a user is sending a message, so at that point it will be undefined; when a user receives the message from the sender, we popualte this through the `tokenId`.

This type `ChatMessageEvent` is used across the app and the [HTML bodies](https://github.com/momentohq/client-sdk-javascript/blob/main/examples/web/nextjs-chat/src/app/pages/chat-room.tsx#L102) too expect this type when displaying the message.. so i wasn't able to remove it from here without changing the HTML part.

examples/web/nextjs-chat/src/utils/momento-web.ts
```diff
@@ -35,17 +35,20 @@ let onErrorCb: (
...
 async function getNewWebClients(): Promise<MomentoClients> {
   webTopicClient = undefined;
   // we don't want to cache the token, since it will expire in 5 min
-  const fetchResp = await fetch(window.location.origin + "/api/momento/token", {
-    cache: "no-store",
-  });
+  const fetchResp = await fetch(
+    window.location.origin + `/api/momento/token?username=${username}`,
+    {
+      cache: "no-store",
```
allenheltondev, line 49: doesn't this header need to come back in the response?

allenheltondev (approved):

examples/web/nextjs-chat/src/utils/momento-web.ts
pratik151192: Oh you mean for us to store it? This part is simply sending the `username` while generating the disposable token and gets back the `token` in the `fetchResp`.

examples/web/nextjs-chat/src/app/page.tsx
```diff
@@ -85,6 +86,12 @@
   }
 
   if (!usernameSelected) {
+    const handleUsernameChange = (e) => {
+      const newUsername = e.target.value;
+      setUsername(newUsername);
+      Cookies.set('username', newUsername, { expires: 7 }); // Store username in a cookie
```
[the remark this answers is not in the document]
allenheltondev: I think it's fine

examples/web/nextjs-chat/src/utils/momento-web.ts
allenheltondev: I'm talking about the `cache` header. That's a response header that browsers look at to determine if it can cache responses or not. It looks like you're providing it in the request, which i don't think will do anything.

pratik151192: Oh lol yea that was as it is from before i am not sure at all what that does. there's a comment on this which i think it signals it to not cache?

`  // we don't want to cache the token, since it will expire in 5 min`

PR #981, opened 2023-10-25
fix: import/export permission roles from web-sdk

rishtigupta:
## PR Description:
- import and export TopicRole, CacheRole from web-sdk such that they can be consumed from the sdk-web rather than sd-core in the console

### DIFF
--- a/packages/client-sdk-web/src/index.ts
+++ b/packages/client-sdk-web/src/index.ts
@@ -122,6 +122,7 @@ import {
   TopicSelector,
   AllCaches,
   AllTopics,
+  CacheRole, TopicRole
 } from '@gomomento/sdk-core';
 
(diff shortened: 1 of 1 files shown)


packages/client-sdk-web/src/index.ts
```diff
@@ -122,6 +122,7 @@ import {
   TopicSelector,
   AllCaches,
   AllTopics,
+  CacheRole, TopicRole
```
bruuuuuuuce, line 125: nit:

```suggestion
  CacheRole, 
  TopicRole,
```

bruuuuuuuce (approved):
:shipit:

cprice404 (commented):
:ty:

PR #989, opened 2023-10-27
chore: remove dependency on sdk-core from examples

pratik151192:
Added more exports to `index.ts` to remove examples dependency on `sdk-core` after a release. Tested with local installation that was failing on the first PR and now its succeeding after these imports.

### DIFF
--- a/packages/client-sdk-nodejs/src/index.ts
+++ b/packages/client-sdk-nodejs/src/index.ts
@@ -143,6 +143,8 @@ import {
   TopicSelector,
   AllCaches,
   AllTopics,
+  LeaderboardOrder,
+  ILeaderboard,
 } from '@gomomento/sdk-core';
 
 import {Configuration, CacheConfiguration} from './config/configuration';
@@ -329,6 +331,8 @@ export {
   LeaderboardConfiguration,
   LeaderboardClientConfiguration,
   PreviewLeaderboardClient,
+  LeaderboardOrder,
+  ILeaderboard,
   // Errors
   MomentoErrorCode,
   SdkError,
--- a/examples/nodejs/cache/doc-example-files/doc-examples-js-apis.ts
+++ b/examples/nodejs/cache/doc-example-files/doc-examples-js-apis.ts
@@ -74,6 +74,8 @@ import {
   TopicConfigurations,
   PreviewLeaderboardClient,
   LeaderboardConfigurations,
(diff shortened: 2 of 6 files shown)


pratik151192:
Hmm looks like this was added for node version 14 to work

```
Error:    8:1  error  '@gomomento/sdk-core' should be listed in the project's dependencies. Run 'npm i -S @gomomento/sdk-core' to add it  import/no-extraneous-dependencies
Error:   16:8  error  "@gomomento/sdk-core" is extraneous                                                                                 node/no-extraneous-import
```

cprice404:
> Hmm looks like this was added for node version 14 to work
> 
> ```
> Error:    8:1  error  '@gomomento/sdk-core' should be listed in the project's dependencies. Run 'npm i -S @gomomento/sdk-core' to add it  import/no-extraneous-dependencies
> Error:   16:8  error  "@gomomento/sdk-core" is extraneous                                                                                 node/no-extraneous-import
> ```

I believe that error is telling you about another place where core is explicitly used in an import, I don't think it has to do with the node version. We should fix the imports.

cprice404 (commented):
more imports we need to fix I think

pratik151192:
> I believe that error is telling you about another place where core is explicitly used in an import, I don't think it has to do with the node version. We should fix the imports.

ah yeah will fix

bruuuuuuuce (commented):
we should also update the package.json and remove the dep

pratik151192:
@bruuuuuuuce 

> we should also update the package.json and remove the dep

yep i can only successfully do that after a release with the changes in this PR

packages/client-sdk-nodejs/src/index.ts
```diff
@@ -143,6 +143,8 @@ import {
   TopicSelector,
   AllCaches,
   AllTopics,
+  LeaderboardOrder,
+  ILeaderboard,
```
cprice404, line 147: thank you! do we need to do this in the web sdk too?

cprice404 (approved):

cprice404 (approved):

packages/client-sdk-nodejs/src/index.ts
pratik151192: yep did that and mistakenly committed my local `npm i` installation.. will update examples once released!

PR #991, opened 2023-10-28
nodejsfunction excludes aws-sdk by default

martzcodes:
The topics example with CDK uses aws-sdk version 3 (`@aws-sdk/*` pattern).  This is excluded by default https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda_nodejs.BundlingOptions.html#externalmodules 

By specifying `aws-sdk` (version 2) as an external module in the bundling options you're overriding the default external modules which actually causes esbuild to bundle aws-sdk version 3 into the lambda.

This PR removes the incorrect `aws-sdk` (version 2) SDK from external modules so that CDK/esbuild will exclude the version of aws-sdk being used in the project.

cc: CDK fan @allenheltondev 😈

### DIFF
--- a/examples/nodejs/lambda-examples/topics-microservice/lib/topics-microservice-stack.ts
+++ b/examples/nodejs/lambda-examples/topics-microservice/lib/topics-microservice-stack.ts
@@ -21,11 +21,6 @@ export class TopicsMicroserviceStack extends Stack {
         "SECRETNAME": secretName,
         "CACHENAME": cachName
       },
-      bundling: {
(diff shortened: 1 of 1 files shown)


cprice404 (approved):
this looks right to me, thanks @martzcodes !

@bruuuuuuuce you've been messing with stuff like this lately, can you take a quick peek and then merge this? :ty:

bruuuuuuuce (approved):
:shipit:

PR #994, opened 2023-10-31
feat: add rate-limiter using momento cache

pratik151192:


examples/nodejs/rate-limiter/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+All tasks complete!
+
+```
+
+## Key Mechanism
+
+At the heart of our rate-limiter is a key mechanism that allows us to perform rate limiting based on user-per-minute granularity. The key is constructed using a combination of the user ID and the current minute. This key plays a pivotal role in tracking and limiting the number of transactions a user can make in a given minute.
+
+## Approaches
```
cprice404, line 59: I would like to hear what @allenheltondev has to say about this, but my 2 cents:

The story about how you tested the two different implementations and came to the conclusion of which one to is great for a blog post. But here in the examples section of the repo, I suspect that most users don't want to think about all of that, they just heard that we have some code they can use as a rate limiter and they want to copy/paste it and consume it in their app. They want to be able to trust us that we thought about the implementation and came up with the best one.

So IMO we should trim the content in the repo down to just saying "here's a few use cases you might use a rate limiter for, and here's an example snippet showing how to use this rate limiter".

I'd do the same thing for dev docs. And the compare/contrast stuff I'd just save for the blog post.

@allenheltondev WDYT? if you think we should keep both versions here too I don't have an issue with it, just not necessarily what I'm guessing a user will be looking for here.

examples/nodejs/rate-limiter/README.md
```diff
@@ -0,0 +1,133 @@
+<head>
+  <meta name="Momento Node.js Client Library Documentation" content="Node.js client software development kit for Momento Cache">
+</head>
+<img src="https://docs.momentohq.com/img/logo.svg" alt="logo" width="400"/>
+
+[![project status](https://momentohq.github.io/standards-and-practices/badges/project-status-official.svg)](https://github.com/momentohq/standards-and-practices/blob/main/docs/momento-on-github.md)
+[![project stability](https://momentohq.github.io/standards-and-practices/badges/project-stability-stable.svg)](https://github.com/momentohq/standards-and-practices/blob/main/docs/momento-on-github.md)
+
+## Rate-limiter Requirements
```
cprice404, line 9: I would add a section above this describing what a rate limiter is and why someone might want to use one.

examples/nodejs/rate-limiter/README.md
```diff
@@ -0,0 +1,133 @@
...
+To run any of the examples you will need to install the dependencies once first:
+
+```bash
+npm install
+```
+
+## Running the rate-limiter
+
+```bash
+npm install
```
anitarua, line 23: duplicate `npm install` command?

examples/nodejs/rate-limiter/worker.ts
```diff
@@ -0,0 +1,133 @@
...
+    totalRequests = parseInt(process.argv[2], 10);
+  }
+
+  if (process.argv[3]) {
+    randomDelayUpperBound = parseInt(process.argv[3], 10);
+  }
+
+  if (process.argv[4]) {
+    tpmLimit = parseInt(process.argv[4], 10);
+  }
```
anitarua, line 42: would it be helpful to include examples of how to pass different CLI args in the readme? might be helpful for users who might want to write a script that tests different rate limiting scenarios

examples/nodejs/rate-limiter/get-increment-rate-limiter.ts
```diff
@@ -0,0 +1,48 @@
...
+  _client: CacheClient;
+  _limit: number;
+
+  constructor(client: CacheClient, limit: number) {
+    super();
+    this._client = client;
+    this._limit = limit;
+  }
+
+  public async acquire(id: string): Promise<boolean> {
```
cprice404, line 18: nit: I wonder if a name like `isLimitExceeded` would be a more natural name for a class that provides rate limiting functionality?

examples/nodejs/rate-limiter/README.md
cprice404: I also don't see anywhere in here where you are talking about what units of time you are able to limit for, or constraints of this class.

cprice404 (commented):
code looks good other than maybe tweaking the function name.

We should align with align on what the main message that we are trying to communicate via the README should be, though.

anitarua (commented):
cool example application!

examples/nodejs/rate-limiter/README.md
pratik151192: That makes sense to me. Do you want me to remove the `GetIncrementRateLimiter` that's not recommended as well in that case?

examples/nodejs/rate-limiter/README.md
pratik151192: Will add. Great feedback!

examples/nodejs/rate-limiter/get-increment-rate-limiter.ts
pratik151192: I actually took this terminology from [guava](https://guava.dev/releases/19.0/api/docs/index.html?com/google/common/util/concurrent/RateLimiter.html), but that's actually a blocking call and waits until a permit is available. So in lieu of that I like your suggestion!

examples/nodejs/rate-limiter/README.md
pratik151192: Ooo nice catch!

examples/nodejs/rate-limiter/worker.ts
pratik151192: yes can add; I actually just added it for my blog demo so not totally applicable but I think I can let this as is and add info about it.

examples/nodejs/rate-limiter/README.md
cprice404: that would be my suggestion, yes, but I'd love to hear what Allen thinks.

allenheltondev: Yeah totally agree, keep it technical in here but definitely link to the blog post, there's no reason not to.

pratik151192: Sure makes sense; will do that once that's out!

anitarua (approved):
lgtm!

pratik151192:
@cprice404 I have on approval will go ahead and merge to link from blog but lmk if README needs a tweak happy to fast follow. Have changed API names and the README essence as per your feedback.

examples/nodejs/rate-limiter/README.md
```diff
@@ -0,0 +1,109 @@
...
+const momento = await CacheClient.create({
+  configuration: Configurations.Laptop.v1(),
+  credentialProvider: CredentialProvider.fromEnvironmentVariable({
+    environmentVariableName: "MOMENTO_API_KEY",
+  }),
+  defaultTtlSeconds: 6000,
+});
+
+const tpmLimit = 10;
+const rateLimiter = new MomentoRateLimiter(momento, tpmLimit);
```
cprice404, line 42: would be nice to accept cache name as an argument here rather than hard-coding in the lib, but not a huge deal.

examples/nodejs/rate-limiter/README.md
```diff
@@ -0,0 +1,109 @@
...
+Momento Rate Limiter - p50 Latency: 2
+Momento Rate Limiter - p90 Latency: 3
+Momento Rate Limiter - p99 Latency: 5
+Momento Rate Limiter - p99.9 Latency: 34
+
+All tasks complete!
+
+```
+
+There are three additional arguments that you can provide to the rate-limiter if you want to experiment with different configurations:
```
cprice404, line 91: these are arguments that you are passing to the example/simulator, not to the rate limiter itself, correct?

examples/nodejs/rate-limiter/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+- totalRequests: The total number of requests that the example will simulate, defaulted to 1000.
+- randomDelayUpperBound: The simulation adds a random delay between 0 and the randomDelayUpperBound, defaulted to 60 seconds.
+- tpmLimit: The rate per minute at which a user or entity will be allowed by the rate-limiter, defaulted to 500.
+
+To override totalRequests to 10, randomDelayUpperBound to 60, and tpmLimit to 1, the command will look like:
+
+```bash
+MOMENTO_API_KEY="yourApiKey" npm run rate-limiter 10 60 1
```
cprice404, line 100: maybe add one more sentence after this that says "if you want to see some throttles, try these settings". And optionally paste another example of your output where throttle count is not zero.

cprice404 (commented):
this looks great, left a few minor suggestions that you can decide whether you want to follow up on.

examples/nodejs/rate-limiter/README.md
pratik151192: When we spoke to Jah I got the feeling that most customers like to decouple control and data plane so left it as is

cprice404: that's different from being locked in to a hard coded cache name though :)

this is okay for example code where we expect users to copy/paste/modify. this would not be okay in a library though :)

pratik151192: ah yeah gotcha now!

PR #998, opened 2023-11-01
chore: wrap momento metrics json in `momento` object for cloudwatch metric filters

anitarua:
So that #970 can incorporate MomentoMetricsMiddleware into the metric filters (e.g. `{$.MomentoMetricsMiddleware.status >= 0}`) to avoid clashing with other metrics customers may stream to a log group.

### DIFF
--- a/packages/client-sdk-nodejs/src/config/middleware/impl/experimental-metrics-middleware.ts
+++ b/packages/client-sdk-nodejs/src/config/middleware/impl/experimental-metrics-middleware.ts
@@ -24,50 +24,52 @@ const FIELD_NAMES: Array<string> = [
 ];
 
 export interface ExperimentalRequestMetrics {
-  /**
-   * number of requests active at the start of the request
-   */
-  numActiveRequestsAtStart: number;
-  /**
-   * number of requests active at the finish of the request (including the request itself)
-   */
-  numActiveRequestsAtFinish: number;
-  /**
-   * The generated grpc object type of the request
-   */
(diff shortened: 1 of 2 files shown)


packages/client-sdk-nodejs/src/config/middleware/impl/experimental-metrics-middleware.ts
```diff
@@ -24,50 +24,52 @@ const FIELD_NAMES: Array<string> = [
...
-  requestSize: number;
-  /**
-   * The size of the response body in bytes
-   */
-  responseSize: number;
-  /**
-   * The ID of the specific connection that made the request
-   */
-  connectionID: string;
+  MomentoMetricsMiddleware: {
```
pratik151192, line 27: not a strong opinion but I think `momento. numActiveRequestsAtStart` or `momento.duration` will be a cleaner UX than `MomentoMetricsMiddleware.duration`

pratik151192 (dismissed):

packages/client-sdk-nodejs/src/config/middleware/impl/experimental-metrics-middleware.ts
anitarua: I like that better too, updated

pratik151192 (approved):

cprice404 (approved):
oh yeah

anitarua:
good call, the .NET update seems to be simpler

PR #1001, opened 2023-11-02
fix: fix error messages to include required limits

pgautier404:


packages/common-integration-tests/src/create-delete-list-cache.ts
```diff
@@ -60,17 +60,35 @@ export function runCreateDeleteListCacheTests(cacheClient: ICacheClient) {
...
+          }, `invalid throughput_throttling_limit. ${limitsMessage}`);
+          expectWithMessage(() => {
+            expect(cache.getCacheLimits().maxItemSizeKb).toEqual(1024);
+          }, `invalid item_size_limit. ${limitsMessage}`);
+          expectWithMessage(() => {
+            expect(cache.getCacheLimits().maxTrafficRate).toEqual(100);
+          }, `invalid throttling_limit. ${limitsMessage}`);
+          expectWithMessage(() => {
+            expect(cache.getCacheLimits().maxTtlSeconds).toEqual(86400);
+          }, `invalid max_ttl. ${limitsMessage}`);
```
cprice404, line 80: supernit, would you mind including the current value in these? or will the inner jest assertion actually log that part?

pgautier404: The assertion failure message should contain the expected and received values.

cprice404 (approved):

PR #1010, opened 2023-11-07
chore: cleanup code causing warning errors on prs

bruuuuuuuce:
On every pr we see a bunch of these eslint warnings, which causes noise and makes the pr harder to grok. This commit fixes all these eslint warnings so that they will no longer appear on every pr

Screen Shot 2023-11-06 at 4 34 39 PM


pratik151192:
Awesome. Sometimes my intellij complains about the exact opposite of the git warnings which makes me confused. I guess I need to configure my intellij properly though

pratik151192 (approved):

anitarua (approved):
💯

PR #1011, opened 2023-11-07
chore: add `deploy` target for TVM to make it easier to deploy

cprice404:
This commit adds a `deploy` target to the token vending machine
and updates the README to use that for deploying instead of `cdk`.
The main motivation here is to automate the `npm install` step for
the lambda project, to prevent users from accidentally skipping it
and running into an error or deploying the lambda with an outdated
version of the dependencies.


pratik151192 (approved):
Woot!

anitarua (approved):
nice!

PR #1013, opened 2023-11-07
fix: set maxWorkers to 1 for jest integration tests to see if it helps flakiness

cprice404:
We are seeing a lot of timeouts in CI test runs, and the only guess
that I have is that there is a lot of CPU contention on the tiny runner
machines due to running too many tests concurrently. This commit sets
the max number of workers to 1 to see if that cuts down on timeouts.


pratik151192 (approved):
LGTM

PR #1014, opened 2023-11-08
feat: webhooks

bruuuuuuuce:
Adds support for our new webhooks apis, putWebhook, deleteWebhook, listWebhooks, and getWebhookSecret


packages/client-sdk-nodejs/src/errors/cache-service-error-mapper.ts
```diff
@@ -30,6 +30,7 @@ export function cacheServiceErrorMapper(err: ServiceError | null): SdkError {
     err?.metadata,
     err?.stack,
   ];
+  console.log('err code', err);
```
bruuuuuuuce, line 33:
```suggestion
```

packages/client-sdk-nodejs/jest.config.ts
```diff
@@ -5,6 +5,7 @@ const config: Config = {
   testEnvironment: 'node',
   roots: ['<rootDir>/test'],
   testMatch: ['**/*.test.ts'],
+  // testMatch: ['./test/integration/shared/webhooks.test.ts'],
```
bruuuuuuuce, line 8:
```suggestion
```

packages/client-sdk-nodejs/src/internal/webhook-client.ts
```diff
@@ -0,0 +1,159 @@
...
+        }
+      );
+    });
+  }
+
+  async listWebhooks(cache: string): Promise<ListWebhooks.Response> {
+    try {
+      validateCacheName(cache);
+    } catch (err) {
+      console.log('herer', err, err instanceof SdkError);
```
bruuuuuuuce, line 88:
```suggestion
```

packages/client-sdk-nodejs/src/internal/webhook-client.ts
```diff
@@ -0,0 +1,158 @@
...
+import {
+  CredentialProvider,
+  MomentoLogger,
+  Webhook,
+  WebhookId,
+  DeleteWebhook,
+  PutWebhook,
+  ListWebhooks,
+  PostUrlWebhookDestination,
+  SdkError,
```
bruuuuuuuce, line 13:
```suggestion
```

packages/common-integration-tests/src/webhooks.ts
```diff
@@ -0,0 +1,84 @@
...
+          const webhookWeAreLookingFor = resp
+            .getWebhooks()
+            .find(
+              wh =>
+                wh.id.webhookName === webhook.id.webhookName &&
+                wh.id.cacheName === webhook.id.cacheName
+            );
+          expect(webhookWeAreLookingFor).toBeTruthy();
+        } else {
+          console.log(resp);
```
bruuuuuuuce, line 74:
```suggestion
```

packages/client-sdk-nodejs/package.json
```diff
@@ -51,10 +52,9 @@
     "uuid": "8.3.2"
   },
   "dependencies": {
-    "@gomomento/generated-types": "0.91.1",
+    "@gomomento/generated-types": "0.94.0",
     "@gomomento/sdk-core": "file:../core",
     "@grpc/grpc-js": "1.9.0",
-    "@types/google-protobuf": "3.15.6",
```
cprice404, line 57: nice catch

packages/core/src/internal/clients/pubsub/AbstractTopicClient.ts
```diff
@@ -46,6 +54,56 @@ export abstract class AbstractTopicClient implements ITopicClient {
...
+
+  /**
+   * Deletes a webhook
+   *
+   * @param {WebhookId} id - The webhook id to be deleted
+   * @returns {Promise<DeleteWebhook.Response>} -
+   * {@link DeleteWebhook.Success} on success.
+   * {@link DeleteWebhook.Error} on failure.
+   */
+  public async deleteWebhook(id: WebhookId): Promise<DeleteWebhook.Response> {
```
cprice404, line 68: food for thought comment here. I'm trying to focus extra attention on the public APIs in this PR since this is the first SDK where we will be exposing them.

When I first saw what you did for `WebHookId` I thought it was nice to encapsulate that into a single object. But then looking over it further I'm wondering if it would be more consistent with the rest of our SDK APIs to just keep it as two args: cacheName, webhookName.

e.g., with the various Topic APIs, we don't encapsulate a `TopicId`, we just expose both things as args. Kinda true with collections too.

And, like, in the CacheClient, we made this `.cache()` interface that provides all of the CacheClient APIs with the cache name curried. If we wanted to do that in the future for TopicClient as well, these signatures would be a little out of place.

Open to discussion on this, I don't have an overwhelmingly strong opinion about it, but it is feeling to me like it would be a bit more consistent to break the two things into separate args.

packages/common-integration-tests/src/webhooks.ts
```diff
@@ -0,0 +1,113 @@
...
+        destination: new PostUrlWebhookDestination('someurl.com'),
+      });
+    });
+
+    ItBehavesLikeItValidatesTopicName((props: ValidateTopicProps) => {
+      return topicClient.putWebhook({
+        id: {
+          cacheName: 'a cache',
+          webhookName: 'ItBehavesLikeItValidatesCacheName',
+        },
```
cprice404, line 41: yeah, like, this looks a little odd that these two params aren't just at the same level as topicName and destination

packages/core/src/utils/webhook-destination.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  type: WebhookDestinationType.PostUrl;
+
+  getType(): WebhookDestinationType.PostUrl {
+    return WebhookDestinationType.PostUrl;
+  }
+}
+
+export type WebhookDestination = InstanceType<typeof PostUrlWebhookDestination>;
```
cprice404, line 28: can we union in `string` here so that it is valid for them do just do `destination: "https://foo.com/webhook"`?

cprice404 (commented):
i've looked over this, mostly focusing on APIs. overall it looks really good, I like what you did for the destinations for the most part. just a few nit / food-for-thought comments on a couple things in the API

packages/core/src/utils/webhook-destination.ts
bruuuuuuuce: We can, but then we will loose the type safety of the `WebhookDestination.type`

cprice404: hrm. that's a bummer, I do like the idea of being able to `switch` on that, but also since we don't know if/when we are going to even support other destinations, i feel like the 90% use case API for right now is gonna be people just wanting to specify a URL.

Maybe instead of changing this type definition, you could just change the signature of `putWebhook` so that it would accept either a WebhookDestination or a `string`?

bruuuuuuuce: yea that will work

packages/common-integration-tests/src/webhooks.ts
bruuuuuuuce: we can put them at the same level, but do you want to take them out of the object completely?

So sig becomes

`putWebhook(cacheName: string, topicName: string, webhookName: string, destination: string | WebhookDestination)`

bruuuuuuuce: when there are that many strings i just worry that its too easy to swap like topicName and cacheName since typescript doesnt have named parameters

cprice404: yeah, i hear you - and if we were starting over from scratch you could probably sell me on making every API just have a single `options` or `props` argument to solve that. but since the existing APIs don't have that I feel like it would be inconsistent to do that just for this one / these, don't you?

cprice404: also I think all of the existing APIs in the current client start with `cacheName` and `topicName`, but after that, it's fair game. so e.g. look at the sig of `subscribe`. you could follow that pattern for webhooks and I think it would be consistent with the existing APIs.

packages/core/src/clients/ITopicClient.ts
```diff
@@ -24,7 +23,17 @@ export interface ITopicClient {
   ): Promise<TopicSubscribe.Response>;
 
   listWebhooks(cache: string): Promise<ListWebhooks.Response>;
-  putWebhook(webhook: Webhook): Promise<PutWebhook.Response>;
-  deleteWebhook(id: WebhookId): Promise<DeleteWebhook.Response>;
-  getWebhookSecret(id: WebhookId): Promise<GetWebhookSecret.Response>;
+  putWebhook(
+    cacheName: string,
+    webhookName: string,
+    options: PutWebhookCallOptions
```
cprice404, line 29: I'm saying that all the other functions have `cacheName` and `topicName` as their first two arguments.

Those are the two arguments I was advocating for consistency on.

If you want to put the `webhookName` and friends into `Options` on any of these PRs I'm okay with it.

bruuuuuuuce: for the other methods which have `cacheName` and `topicName`, they are for `topics`, so `topicName` is a first class citizen. For `webhooks`, `webhookName` is a first class citizen, so I think the pattern is consistent

cprice404: i disagree. this re-ordering of the arguments is the thing that would be the most likely to cause users to accidentally put the wrong string values in the wrong order.

there is a hierarchy to these things:

* Cache is the top level
* Topics are next - you can't have a topic without a cache
* Webhooks are next - you can't have a webhook without a topic

the arguments should appear in that order. whether you want to make the webhookName argument a string or part of an `Options` is up to you, but the order should be consistent and reflect the hierarchy.

bruuuuuuuce: `DeleteWebhook` and `getWebhookSecret` both have `cache: String, webhookName: string` since that is what we use to identify the webhook. If we have `putWebhook` be `cache: string, topic: string, options: Options` then it will be inconsistent with `deleteWebhook` and `getWebhookSecret`

cprice404: if you don't agree with what i'm saying here, feel free to start a notion doc or slack thread with the proposed 5 signatures in it and we can solicit some additional opinions.

cprice404 (dismissed):
okay

rishtigupta (approved):

PR #1023, opened 2023-11-09
chore: build in the same way on PRs and releases

nand4011:
Run build-and-test-package.sh for each package being released instead of manually calling the build, lint, and test commands. This builds the release packages in the same way that pr packages are built.

Part of https://github.com/momentohq/dev-eco-issue-tracker/issues/546
Both scripts ran `npm run lint` already, so I don't know how the release job was failing on lint if a pr didn't, but this makes them use the same code at least.


anitarua (approved):
lgtm!

PR #1029, opened 2023-11-10
chore: add example snippets for CollectionTTL

cprice404:
This is based on user feedback that our docs didn't provide enough
clarity around CollectionTtl. In a subsequent PR we will consume
these examples from the docs.


bruuuuuuuce (approved):
:shipit:

PR #1034, opened 2023-11-13
feat: mvi search and fetch vectors

malandis:
Adds `searchAndFetchVectors` method to both web and Node.js clients, including:
- method definitions
- response types (+`SearchAndFetchVectorsHit`)
- integration tests

For the integration tests I parameterized the existing search tests to work on both `search` and `searchAndFetchVectors`.


packages/core/src/internal/clients/vector/AbstractVectorIndexClient.ts
```diff
@@ -133,6 +135,34 @@ export abstract class AbstractVectorIndexClient
     return await this.dataClient.search(indexName, queryVector, options);
   }
 
+  /**
+   * Searches for the most similar vectors to the query vector in the index.
+   *
+   * Ranks the vectors according to the similarity metric specified when the
+   * index was created.
+   * Also retursn the vectors associated with each result.
```
nand4011, line 143: Minor typo.

malandis, line 143:
```suggestion
   * Also returns the vectors associated with each result.
```

malandis: Good catch. Fixed in [297d469](https://github.com/momentohq/client-sdk-javascript/pull/1034/commits/297d469e0351c00ac425082d52a40dfdace5b447)

nand4011 (approved):
LGTM. The python sdk keeps SearchHit separate from the search vector. Do we want to change that to combine them like you do here?

malandis:
> LGTM. The python sdk keeps SearchHit separate from the search vector. Do we want to change that to combine them like you do here?

I'm not sure what you mean?

nand4011:
Ignore me. That was a poor reading of the python search hit.

PR #1039, opened 2023-11-14
feat: add initial support for middleware in web sdk

cprice404:
Posting this in case you all want to try to finish it and merge it to use to help debug flakey CI while I'm gone.

It's working for me locally but before this PR can be merged you will want to at least revert the changes to the `basic` example, where I did an `npm install ../../../packages/client-sdk-web`. (I included this in the PR to illustrate how I was testing the middleware.)

This implementation is not ideal because all of the middleware APIs are synchronous. I would much prefer them to be asynchronous. However, to support that I believe we would need to be using the promise-based grpc-web client, which seemed like too large/risky a change to take on right now.

https://github.com/grpc/grpc-web/issues/877


bruuuuuuuce (approved):
:shipit:

PR #1045, opened 2023-11-15
fix: import structure so dist folder has correct folder heirarchy

bruuuuuuuce:
We are seeing issues installing the lastest nodejs `1.50.4` sdk. On main, when building the nodejs sdk, the dist output looks like

```
dist
├── client-sdk-web
│   ├── package.json
│   ├── src
│   └── test
└── core
    └── src
```

It is expecting 

```
dist/src/index.js
```
This is resulting in an error `Unable to resolve path to module '@gomomento/sdk'` when trying to install the sdk.

It looks like this has to do with some of our imports for the webhooks client coming from `../core/...` vs `@gomoment-core/dist/...`. 

I have tested this change locally on my box, `npm pack` the sdks, installing them in examples, and making sure they will work. I will fast follow this with a pr that will add some steps to our build to ensure that the `dist` folder has the correct folder structure. But I want to get this out first

### DIFF
--- a/packages/client-sdk-nodejs/package.json
+++ b/packages/client-sdk-nodejs/package.json
@@ -31,7 +31,6 @@
   "license": "Apache-2.0",
   "devDependencies": {
     "@gomomento/common-integration-tests": "file:../common-integration-tests",
-    "@types/google-protobuf": "3.15.10",
     "@types/jest": "27.5.2",
     "@types/node": "14.18.3",
     "@types/uuid": "8.3.4",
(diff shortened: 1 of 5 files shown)


packages/client-sdk-nodejs/package.json
```diff
@@ -55,6 +54,7 @@
     "@gomomento/generated-types": "0.96.0",
     "@gomomento/sdk-core": "file:../core",
     "@grpc/grpc-js": "1.9.0",
+    "@types/google-protobuf": "3.15.10",
```
bruuuuuuuce, line 57: I was getting some weird `@types/google-protobuf` is missing error when installing our sdk. So adding this back here. This probably means that we are leaking some grpc types somewhere, and this pr is not aiming at fixing that.

rishtigupta (approved):

pratik151192 (approved):

PR #1046, opened 2023-11-15
chore: tweak vector index info to string to quote string fields

malandis:
We add quoting on the to VectorIndexInfo string fields on `toString`
for clarity.


bruuuuuuuce (approved):

bruuuuuuuce:
:shipit:

PR #1051, opened 2023-11-17
fix: don't validate limits outside of CI

bruuuuuuuce:
Inside of github actions, the env var `CI` is always set to true

https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

We can use this to key off of, and have a certain subset of integration tests only run in the CI environment, such as ones that verify limits against a cache

### DIFF
--- a/packages/common-integration-tests/src/common-int-test-utils.ts
+++ b/packages/common-integration-tests/src/common-int-test-utils.ts
@@ -21,6 +21,18 @@ import {
 import {v4} from 'uuid';
 import {ITopicClient} from '@gomomento/sdk-core/dist/src/clients/ITopicClient';
 
+const isInsideGithubCI = process.env.CI === 'true';
+const itif = (condition: boolean) => (condition ? it : it.skip);
+const maybe = (condition: boolean) => (condition ? describe : describe.skip);
+/**
+ * Only runs the tests inside the 'it' block if inside a continuous integration environment.
+ */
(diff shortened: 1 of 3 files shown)


packages/common-integration-tests/src/common-int-test-utils.ts
```diff
@@ -21,6 +21,18 @@ import {
 import {v4} from 'uuid';
 import {ITopicClient} from '@gomomento/sdk-core/dist/src/clients/ITopicClient';
 
+const isInsideGithubCI = process.env.CI === 'true';
```
malandis, line 24: Does GH set this?

pratik151192 (approved):

packages/common-integration-tests/src/common-int-test-utils.ts
```diff
@@ -21,6 +21,18 @@ import {
...
+const itif = (condition: boolean) => (condition ? it : it.skip);
```
malandis, line 25: Nice

packages/common-integration-tests/src/common-int-test-utils.ts
bruuuuuuuce: yea they set it here https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

malandis (approved):

PR #1054, opened 2023-11-17
fix: use jest-ci-spec-reporter for reporting integration test output

bruuuuuuuce:
We got some feedback that the dotnet sdk canaries logging was very easy to grok, while the javascript sdk canaries logging was still a bit confusing to understand. This pr attempts to log the javascript sdk integration tests in as similar manner as possible to the dotnet integration tests

Using the [jest-ci-spec-reporter](https://github.com/robertbradleyux/jest-ci-spec-reporter), we can log a pass/fail for each individual test on a single line, which is the same behavior as the dotnet integration tests.

This is what reports look like when all tests pass
Screen Shot 2023-11-17 at 12 32 49 PM

And this is what it looks like when there is a failing test
Screen Shot 2023-11-17 at 1 06 31 PM


pratik151192 (approved):
🎉

ben-kugler (approved):

PR #1058, opened 2023-12-06
feat: add mvi methods for get item (metadata)? batch

malandis:
Implements MVI `get item batch` and `get item metadata batch`, adds tests, and extracts `VectorIndexMetadata` to separate type.

### DIFF
--- a/packages/core/src/internal/clients/vector/IVectorIndexDataClient.ts
+++ b/packages/core/src/internal/clients/vector/IVectorIndexDataClient.ts
@@ -3,6 +3,8 @@ import {
   VectorDeleteItemBatch,
   VectorSearch,
   VectorSearchAndFetchVectors,
+  VectorGetItemBatch,
+  VectorGetItemMetadataBatch,
 } from '../../../messages/responses/vector';
 import {VectorIndexItem} from '../../../messages/vector-index';
 import {SearchOptions} from '../../../clients/IVectorIndexClient';
@@ -26,4 +28,12 @@ export interface IVectorIndexDataClient {
     indexName: string,
     ids: Array<string>
   ): Promise<VectorDeleteItemBatch.Response>;
+  getItemBatch(
+    indexName: string,
+    ids: Array<string>
+  ): Promise<VectorGetItemBatch.Response>;
+  getItemMetadataBatch(
+    indexName: string,
+    ids: Array<string>
+  ): Promise<VectorGetItemMetadataBatch.Response>;
 }
--- a/packages/core/src/messages/responses/vector/vector-get-item-batch.ts
+++ b/packages/core/src/messages/responses/vector/vector-get-item-batch.ts
@@ -1,5 +1,5 @@
 import {SdkError} from '../../../errors';
-import {VectorIndexItem} from '../../vector-index';
+import {VectorIndexStoredItem} from '../../vector-index';
 import {ResponseBase, ResponseError, ResponseSuccess} from '../response-base';
 
 /**
@@ -22,7 +22,7 @@ import {ResponseBase, ResponseError, ResponseSuccess} from '../response-base';
  * ```
  */
 export abstract class Response extends ResponseBase {
-  hits(): Record<string, VectorIndexItem> | undefined {
+  hits(): Record<string, VectorIndexStoredItem> | undefined {
     if (this instanceof Success) {
       return this.hits();
     }
@@ -36,8 +36,8 @@ class _Success extends Response {}
  * Indicates a Successful VectorGetItemBatch request.
  */
 export class Success extends ResponseSuccess(_Success) {
-  private readonly _hits: Record<string, VectorIndexItem>;
-  constructor(hits: Record<string, VectorIndexItem>) {
+  private readonly _hits: Record<string, VectorIndexStoredItem>;
+  constructor(hits: Record<string, VectorIndexStoredItem>) {
     super();
     this._hits = hits;
   }
@@ -46,9 +46,9 @@ export class Success extends ResponseSuccess(_Success) {
    *
    * Items that were not found will not be included in the
    * returned object.
-   * @returns {Record<string, VectorIndexItem>} The items that were found in the index.
+   * @returns {Record<string, VectorIndexStoredItem>} The items that were found in the index.
    */
-  hits(): Record<string, VectorIndexItem> {
+  hits(): Record<string, VectorIndexStoredItem> {
     return this._hits;
   }
 }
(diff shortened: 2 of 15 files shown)


packages/core/src/messages/responses/vector/vector-get-item-batch.ts
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * Returns the found items from the VectorGetItemBatch request.
+   *
+   * Items that were not found will not be included in the
+   * returned object.
+   * @returns {Record<string, VectorIndexItem>} The items that were found in the index.
+   */
+  hits(): Record<string, VectorIndexItem> {
+    return this._hits;
+  }
```
malandis, line 53: As in the python client, I opted to return a record for the hits. If someone wants a list of responses in identical order to the id's requested, we can add that.

packages/core/src/messages/responses/vector/vector-get-item-metadata-batch.ts
```diff
@@ -0,0 +1,84 @@
...
+ * if (response instanceof VectorGetItemMetadataBatch.Error) {
+ *   // Handle error as appropriate.  The compiler will smart-cast `response` to type
+ *   // `VectorGetItemMetadataBatch.Error` in this block, so you will have access to the properties
+ *   // of the Error class; e.g. `response.errorCode()`.
+ * }
+ * ```
+ */
+export abstract class Response extends ResponseBase {
+  hits():
+    | Record<string, Record<string, string | number | boolean | Array<string>>>
```
malandis, line 25: It's becoming tedious to copy the metadata `Record` type signature. I prefer extracting this to a type alias in a single place. If we did this though then the type would not expand in IDE tooltips; users would have to go to the method definition then hover over the metadata alias to understand what it was.

Though that's less than ideal for users, the duplication is becoming a bit much; should we add a new metadata type alternative in the future, we will have to update it in many places. Thus I recommend extracting the type

malandis (commented):
Leaving this here for early review. Thoughts @cprice404 ?

packages/core/src/messages/responses/vector/vector-get-item-metadata-batch.ts
cprice404: agree

packages/core/src/messages/vector-index.ts
```diff
@@ -1,5 +1,39 @@
...
+   */
   metadata?: Record<string, string | number | boolean | Array<string>>;
 }
+
+/**
+ * An item stored in the vector index.
+ *
+ * @remarks Specifying metadata is optional when upserting with
+ * `VectorIndexItem` but it is always returned when fetching items.
+ * Ie we know with certainty that `metadata` is not undefined.
```
cprice404, line 24: 👍

packages/core/src/messages/responses/vector/vector-get-item-metadata-batch.ts
```diff
@@ -0,0 +1,84 @@
...
+ * ```
+ * if (response instanceof VectorGetItemMetadataBatch.Error) {
+ *   // Handle error as appropriate.  The compiler will smart-cast `response` to type
+ *   // `VectorGetItemMetadataBatch.Error` in this block, so you will have access to the properties
+ *   // of the Error class; e.g. `response.errorCode()`.
+ * }
+ * ```
+ */
+export abstract class Response extends ResponseBase {
+  hits():
```
cprice404, line 24: just curious about the choice of the word `hit` for these. Is that a term of art in this space? I'd probably have just expected `values` or `items`.

packages/core/src/internal/clients/vector/IVectorIndexDataClient.ts
```diff
@@ -26,4 +28,12 @@ export interface IVectorIndexDataClient {
...
     ids: Array<string>
   ): Promise<VectorDeleteItemBatch.Response>;
+  getItemBatch(
+    indexName: string,
+    ids: Array<string>
+  ): Promise<VectorGetItemBatch.Response>;
+  getItemMetadataBatch(
+    indexName: string,
+    ids: Array<string>
+  ): Promise<VectorGetItemMetadataBatch.Response>;
```
cprice404, line 38: just out of curiosity - is it common for users to only want one or the other of these two types of data? (i.e. they either want the item, or they want the metadata, but they don't want both?)

I can imagine that there might be some server-side implementation / performance details that make it desirable to have these as two separate APIs but I'm not familiar enough with use cases to know if users are going to end up usually only calling one of the 2 of these, or if they will frequently be calling both of them. If the latter, just curious about thoughts about possibly returning both from a single api.

cprice404 (dismissed):

packages/core/src/messages/responses/vector/vector-get-item-metadata-batch.ts
malandis: In search, `hits` are the results of a search operation. While this is more of a lookup than a search, I was thinking about that and how we have "cache hits" and "cache misses"; each item is either a hit or a miss in the index. Thus an object with only the hits I called `hits`. That said I was gnashing my teeth over what to call it.

cprice404: ah, so I guess these APIs are not guaranteed to return the same number of results as the number of input IDs? that makes some sense. I feel like I still would have expected `items` or `rows` or `values`, before `hits` but I don't have a huge problem with `hits`, just offering up my unfiltered reaction :)

packages/core/src/internal/clients/vector/IVectorIndexDataClient.ts
malandis: A customer asked for these. I am unclear on the particular use outside of the search.

That said, there are two RPCs under the hood: `get item batch` and `get item metadata batch`. Each one has a `metadata selector` that allows us (users) to choose all metadata, some metadata, or no metadata. For the first cut I kept it simple and did not plumb that through just yet.

So in the future if someone only wants the vectors, we can plumb through the `metadata selector` and they can invoke `get item batch with no metadata`. Or we can furnish a new method that is just "get vector".

packages/core/src/messages/responses/vector/vector-get-item-metadata-batch.ts
malandis: Refactored in [025f3c0](https://github.com/momentohq/client-sdk-javascript/pull/1058/commits/025f3c0387cf541f89395942e50db92cf3307247)

packages/core/src/messages/responses/vector/vector-get-item-metadata-batch.ts
malandis: > ah, so I guess these APIs are not guaranteed to return the same number of results as the number of input IDs?

That's right. The methods are akin to the dictionary get fields operations in that way.

> I feel like I still would have expected items or rows or values, before hits but I don't have a huge problem with hits, just offering up my unfiltered reaction :)

cprice404 (approved):

PR #1059, opened 2023-12-06
feat: add new rotate webhook secret api

bruuuuuuuce:


packages/client-sdk-web/src/internal/webhook-client.ts
```diff
@@ -207,4 +209,44 @@ export class WebhookClient implements IWebhookClient {
...
+        (err, resp) => {
+          if (err || !resp) {
+            resolve(
+              new RotateWebhookSecret.Error(cacheServiceErrorMapper(err))
+            );
+          } else {
+            resolve(
+              new RotateWebhookSecret.Success({
+                secret: resp.getSecretString(),
+                webhookName: id.webhookName,
```
bruuuuuuuce, line 243: the protos don't include this, but I think it might be useful to include in the response. If we want to keep these 1:1 with the protos, I can remove it here.

cprice404: nah i'm on board

malandis (approved):

PR #1062, opened 2023-12-06
chore: add scheduled job to update momento deps in all examples

cprice404:
This is the first step of replacing the dependabot functionality;
for now all this does is run the script. Will add more commits
to make it file a PR once we ensure this is working.


bruuuuuuuce (approved):
:shipit:

PR #1094, opened 2023-12-27
feat: support configuring multiple grpc connections for leaderboard client

cprice404:
Prior to this commit, the node.js leaderboard client did not honor the
our grpc config setting for "numClients", which is intended to control the
number of TCP connections we make to the server. This commit adds support
for that setting, meaning that users who may exceed 100 concurrent requests
can increase the value in order to get higher throughput.

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/leaderboard-data-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/leaderboard-data-client.ts
@@ -18,6 +18,7 @@ import {
   validateLeaderboardOffset,
   validateLeaderboardCount,
   validateLeaderboardRanks,
+  range,
 } from '@gomomento/sdk-core/dist/src/internal/utils';
 import {LeaderboardConfiguration} from '../config/leaderboard-configuration';
 import {leaderboard} from '@gomomento/generated-types/dist/leaderboard';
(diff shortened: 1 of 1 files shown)


packages/client-sdk-nodejs/src/internal/leaderboard-data-client.ts
```diff
@@ -58,33 +60,44 @@ export class LeaderboardDataClient implements ILeaderboardDataClient {
       `Creating leaderboard client using endpoint: '${this.credentialProvider.getCacheEndpoint()}'`
     );
 
-    this.clientWrapper = new IdleGrpcClientWrapper({
```
cprice404, line 61: this part is mostly a whitespace change; check github's "hide whitespace" to make this diff more obvious

eaddingtonwhite (approved):
👌

packages/client-sdk-nodejs/src/internal/leaderboard-data-client.ts
```diff
@@ -522,4 +535,11 @@ export class LeaderboardDataClient implements ILeaderboardDataClient {
       );
     });
   }
+
+  protected getNextDataClient(): leaderboard.LeaderboardClient {
```
cprice404, line 539: this method is stolen from the existing code for the CacheClient where we are already supporting multiple grpc connections.

PDXKimani (approved):

anp13 (approved):
🚢

PR #1098, opened 2023-12-30
feat: new configuration setting to opt-in to throwing on errors

cprice404:
This commit adds a new configuration setting to the various Configuration objects: `ThrowOnErrors`.

The default behavior for all of the Momento clients is to return errors as part of the function's return value, to try to make error handling feel more like a first-class part of the API. However there are some cases where it is easier to allow errors to bubble up the call stack and handle them via a try/catch, and some users may prefer that style of error handling since it is more prevalent in JS.

If the new ThrowOnErrors config option can be set to `true`, then the clients will throw exceptions rather than returning them as errors.

### DIFF
--- a/packages/client-sdk-nodejs/src/errors/cache-service-error-mapper.ts
+++ b/packages/client-sdk-nodejs/src/errors/cache-service-error-mapper.ts
@@ -18,25 +18,27 @@ import {
   FailedPreconditionError,
 } from '../../src';
 
+export interface HandleCacheServiceErrorOptions {
+  err: ServiceError | null;
+  errorResponseFactoryFn: (err: SdkError) => unknown;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  resolveFn: (result: any) => void;
+  rejectFn: (err: SdkError) => void;
+}
+
 export class CacheServiceErrorMapper {
   private readonly throwOnError: boolean;
 
   constructor(throwOnError: boolean) {
     this.throwOnError = throwOnError;
   }
-  handleError(
-    err: ServiceError | null,
-    errorResponseFactoryFn: (err: SdkError) => unknown,
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    resolveFn: (result: any) => void,
-    rejectFn: (err: SdkError) => void
-  ): void {
-    const error = this.convertError(err);
+  handleError(opts: HandleCacheServiceErrorOptions): void {
+    const error = this.convertError(opts.err);
 
     if (this.throwOnError) {
-      rejectFn(error);
+      opts.rejectFn(error);
     } else {
-      resolveFn(errorResponseFactoryFn(error));
+      opts.resolveFn(opts.errorResponseFactoryFn(error));
     }
   }
 
--- a/packages/client-sdk-nodejs/src/errors/cache-service-error-mapper.ts
+++ b/packages/client-sdk-nodejs/src/errors/cache-service-error-mapper.ts
@@ -18,7 +18,7 @@ import {
   FailedPreconditionError,
 } from '../../src';
 
-export interface HandleCacheServiceErrorOptions {
+export interface ResolveOrRejectErrorOptions {
   err: ServiceError | null;
   errorResponseFactoryFn: (err: SdkError) => unknown;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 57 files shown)


packages/client-sdk-nodejs/src/errors/cache-service-error-mapper.ts
```diff
@@ -18,49 +18,74 @@ import {
...
-    case Status.ALREADY_EXISTS:
-      return new AlreadyExistsError(...errParams);
-    default:
-      return new UnknownError(...errParams);
+export class CacheServiceErrorMapper {
+  private readonly throwOnError: boolean;
+
+  constructor(throwOnError: boolean) {
+    // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+    console.log(`CONSTRUCTING CACHE SERVICE ERROR MAPPER: ${throwOnError}`);
```
bruuuuuuuce, line 26:
```suggestion
```

packages/client-sdk-nodejs/src/errors/cache-service-error-mapper.ts
```diff
@@ -18,49 +18,74 @@ import {
...
-      return new LimitExceededError(...errParams);
-    case Status.ALREADY_EXISTS:
-      return new AlreadyExistsError(...errParams);
-    default:
-      return new UnknownError(...errParams);
+export class CacheServiceErrorMapper {
+  private readonly throwOnError: boolean;
+
+  constructor(throwOnError: boolean) {
+    // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
```
bruuuuuuuce, line 25:
```suggestion
```

packages/client-sdk-nodejs/src/internal/cache-control-client.ts
```diff
@@ -113,15 +122,20 @@ export class CacheControlClient {
...
         .DeleteCache(
           request,
           {interceptors: this.interceptors},
           (err, _resp) => {
             if (err) {
-              resolve(new DeleteCache.Error(cacheServiceErrorMapper(err)));
+              this.cacheServiceErrorMapper.handleError(
+                err,
+                e => new DeleteCache.Error(e),
+                resolve,
```
bruuuuuuuce, line 136: is there a compilation failure if you accidentally swap the resolve and reject functions?

packages/client-sdk-nodejs/src/internal/pubsub-client.ts
```diff
@@ -244,7 +253,7 @@ export class PubsubClient extends AbstractPubsubClient {
         serviceError.code === Status.INTERNAL &&
         serviceError.details === PubsubClient.RST_STREAM_NO_ERROR_MESSAGE;
       const momentoError = new TopicSubscribe.Error(
-        cacheServiceErrorMapper(serviceError)
+        this.cacheServiceErrorMapper.convertError(serviceError)
```
bruuuuuuuce, line 256: We are not throwing here because we might retry the subscription? Do we want to throw when its the first subscription message?

packages/client-sdk-nodejs/src/internal/cache-control-client.ts
cprice404: I don't think so. Presumably you would get some int test failures though because nothing would succeed?

Also open to ideas on how to improve this.

packages/client-sdk-nodejs/src/internal/pubsub-client.ts
cprice404: this is a good question, I probably need to look into this part a little more closely.

cprice404: The thing here is that our JS topic subscriptions are callback-based already. So what that means is that there is no thread of execution to try to throw the exception into; if an error occurs during a subscription then the onError callback gets called.

I can't immediately think of a behavior change that the ThrowOnErrors flag could reasonably introduce for a callback-based subscription model. In other languages where the subscriptions are iterator-style, there's a thread to throw the execption into, but not for callback-style.

Open to ideas/discussion, though!

packages/client-sdk-nodejs/src/errors/cache-service-error-mapper.ts
cprice404: fixed

packages/client-sdk-nodejs/src/internal/cache-control-client.ts
bruuuuuuuce: the idea that comes to mind first is having the function take in an object with named arguments, but I don't think that its a dealbreaker,  and that would require changing basically every line in this pr 😅 . I think how it currently is is good

bruuuuuuuce (dismissed):
:shipit:

packages/client-sdk-nodejs/src/internal/cache-control-client.ts
cprice404: this is done

bruuuuuuuce (approved):
:shipit:

PR #1100, opened 2024-01-12
feat: support default configuration for client objects

cprice404:
Based on some feedback about example code and related content,
this commit makes the Configuration arguments optional for all
Client constructors, and provides default configurations for
each one. This will allow us to simplify our code examples
so that it is easier to highlight the more interesting and
important parts of the code.


packages/client-sdk-nodejs/src/config/vector-index-configurations.ts
```diff
@@ -61,47 +61,3 @@ export class Laptop extends VectorIndexClientConfiguration {
     });
   }
 }
-
-/**
- * Browser config provides defaults suitable for use in a web browser.
- * @export
- * @class Browser
- */
-export class Browser extends VectorIndexClientConfiguration {
```
malandis, line 70: Nice catch

packages/client-sdk-nodejs/src/internal/cache-client-props-with-config.ts
```diff
@@ -0,0 +1,6 @@
+import {CacheClientProps} from '../cache-client-props';
+import {Configuration} from '../config/configuration';
+
+export interface CacheClientPropsWithConfig extends CacheClientProps {
+  configuration: Configuration;
+}
```
malandis, line 6: I wasn't aware you could override a property's optionality (`configuration` is optional now in `CacheClientProps`) this way. That's convenient.

malandis (approved):
👍

bruuuuuuuce (approved):
:shipit:

PR #1101, opened 2024-01-13
feat: support array inputs for DictionarySetFields

cprice404:
Prior to this commit the DictionarySetFields API would accept a Map or a Record containing the elements that the caller wished to add to the dictionary.

A user reported an error when writing JS (not TS) code where they had attempted to pass in an array of tuples (equivalent to what the map entries would have looked like) rather than an actual Map.

This commit adds support for passing an array of tuples without wrapping it in a Map first.


bruuuuuuuce (dismissed):
:shipit:

PR #1105, opened 2024-01-17
feat: leaderboard should work with doubles instead of floats

anitarua:
addresses https://github.com/momentohq/dev-eco-issue-tracker/issues/624

### After service changes

New test should succeed along with the other existing leaderboard tests.

### Before service changes

New test should currently fail with the old version of the protos. Should pass once the protos and backend service are updated to handle doubles instead of floats.

I see the precision discrepancies when testing locally: 
```
 -     "score": 27.004309862363737,
 +     "score": 27.004310607910156,

 -     "score": 16777217,
 +     "score": 16777218,
```


cprice404 (dismissed):
yes

packages/client-sdk-nodejs/package.json
```diff
@@ -18,7 +18,8 @@
     "integration-test-auth": "jest auth-client.test.ts --maxWorkers 1",
     "integration-test-vector": "jest vector-control-plane.test.ts vector-data-plane.test.ts --maxWorkers 1",
     "unit-test": "jest unit",
-    "integration-test": "jest integration --testPathIgnorePatterns \"auth-client.test.ts|vector-control-plane.test.ts|vector-data-plane.test.ts\" --maxWorkers 1",
+    "integration-test-leaderboard": "jest leaderboard --maxWorkers 1",
+    "integration-test": "jest integration --testPathIgnorePatterns \"auth-client.test.ts|vector-control-plane.test.ts|vector-data-plane.test.ts|leaderboard.test.ts\" --maxWorkers 1",
```
cprice404, line 22: @anitarua I think if we still want the leaderboard tests to run in github CI then we will have to update one or two of those github yaml files too?

anitarua: oh yeah you're right, I'll add that

cprice404 (approved):
nod

PR #1106, opened 2024-01-17
chore: add examples for node.js config and error handling

cprice404:
This adds some new examples for configuration and error handling,
including how to construct the client without specifying an explicit
Configuration, and how to use the `WithThrowOnErrors(true)` config
option.

### DIFF
--- a/examples/nodejs/cache/doc-example-files/config-and-error-handling.ts
+++ b/examples/nodejs/cache/doc-example-files/config-and-error-handling.ts
@@ -1,0 +1,173 @@
+/**
+ *
+ * This file contains examples of consuming the JavaScript APIs, for use as examples
+ * in the public dev docs.  Each function name that begins with `example_` is available
+ * to the dev docs to inject into the code snippets widget for the specified API.
+ *
+ * These examples should all be JavaScript; we can add TypeScript-specific examples in
+ * a second file in the future if desired.
+ *
+ */
+import {
+  CacheClient,
+  Configurations,
+  CredentialProvider,
+  CreateCache,
+  DeleteCache,
+  CacheFlush,
+  ListCaches,
+  CacheSet,
+  CacheGet,
+  CacheDelete,
+  AuthClient,
+  AllDataReadWrite,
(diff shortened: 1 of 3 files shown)


examples/nodejs/cache/doc-example-files/doc-examples-js-apis.ts
```diff
@@ -1455,6 +1457,10 @@ async function main() {
   await example_API_LeaderboardLength(leaderboard);
   await example_API_LeaderboardRemoveElements(leaderboard);
   await example_API_LeaderboardDelete(leaderboard);
+  await example_API_LeaderboardFetchByRankPagination(leaderboard);
```
cprice404, line 1460: @anitarua I noticed that none of these paginated ones were getting called, idk if that was intentional? If so I can revert this part.

examples/nodejs/cache/doc-example-files/config-and-error-handling.ts
```diff
@@ -0,0 +1,173 @@
...
+  DeleteWebhook,
+  PutWebhook,
+  RotateWebhookSecret,
+  GetWebhookSecret,
+  SdkError,
+  MomentoErrorCode,
+} from '@gomomento/sdk';
+
+/* eslint-disable @typescript-eslint/no-unused-vars */
+async function configuration_ConstructWithNoConfig() {
```
cprice404, line 94: @allenheltondev everything in this file is intended to show the difference between explicitly passing in a Configuration vs. using the default, plus how to use the ThrowOnErrors stuff. I plan to pull these snippets into code blocks in the dev docs in another PR momentarily so if there is anything in here you don't like lmk.

examples/nodejs/cache/doc-example-files/doc-examples-js-apis.ts
anitarua: whoops, definitely didn't mean to leave those out, thanks for catching that!

anitarua (dismissed):

anitarua (approved):

PR #1108, opened 2024-01-18
chore: use default accessor in mvi list indexes response base

malandis:
We have adopted a pattern where users can call accessors on the
response base class to avoid an instanceof guard. That is, suppose a
user calls `client.listIndexes()`. This returns a
`ListVectorIndexes.Response` class, which will either be a `Success`
or `Error`. A user should then use `instanceof` to check the appropriate
subtype.

In order to simplify access we move the accessor to the base class and
do an instanceof check there. In the event the response is not a
success we return undefined.


cprice404 (approved):
eyebrow

PR #1118, opened 2024-02-01
fix: implement default topk behavior for mvi search methods

malandis:
- chore: extract vector search topk to constant
- test: add integration test to exercise default topk
- fix: use default topk in nodejs vector client
- fix: use default topk in web client


cprice404 (approved):
complete

PR #1119, opened 2024-02-02
feat: implement mvi search filters

malandis:
Adds filter expression data types and implementation for `search` and
`searchAndFetchVectors`. Includes search filters for equality checks and logical expressions. A followup PR will include the exhaustive set of comparison operators.

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/vector-index-data-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/vector-index-data-client.ts
@@ -331,7 +331,7 @@ export class VectorIndexDataClient implements IVectorIndexDataClient {
   ): vectorindex._FilterExpression | undefined {
     if (filterExpression === undefined) {
       return undefined;
-    } else if (filterExpression instanceof F.VectorAndExpression) {
+    } else if (filterExpression instanceof F.VectorFilterAndExpression) {
       return new vectorindex._FilterExpression({
         and_expression: new vectorindex._AndExpression({
           first_expression: VectorIndexDataClient.buildFilterExpression(
@@ -342,7 +342,7 @@ export class VectorIndexDataClient implements IVectorIndexDataClient {
           ),
         }),
       });
-    } else if (filterExpression instanceof F.VectorOrExpression) {
+    } else if (filterExpression instanceof F.VectorFilterOrExpression) {
       return new vectorindex._FilterExpression({
         or_expression: new vectorindex._OrExpression({
           first_expression: VectorIndexDataClient.buildFilterExpression(
@@ -353,7 +353,7 @@ export class VectorIndexDataClient implements IVectorIndexDataClient {
           ),
         }),
       });
-    } else if (filterExpression instanceof F.VectorNotExpression) {
+    } else if (filterExpression instanceof F.VectorFilterNotExpression) {
       return new vectorindex._FilterExpression({
         not_expression: new vectorindex._NotExpression({
           expression_to_negate: VectorIndexDataClient.buildFilterExpression(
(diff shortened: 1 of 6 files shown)


packages/client-sdk-nodejs/src/internal/vector-index-data-client.ts
```diff
@@ -325,6 +326,85 @@ export class VectorIndexDataClient implements IVectorIndexDataClient {
     }
   }
 
+  private static buildFilterExpression(
```
malandis, line 329: Would have been nice if this logic could be in one place, but the instantiating the proto classes are completely different between grpc nodejs and web.

packages/core/src/messages/vector-index-filters.ts
```diff
@@ -0,0 +1,194 @@
...
+    return VectorFilterExpressionFactory.or(this, other);
+  }
+
+  /**
+   * Negates this expression.
+   * @returns The negated expression.
+   */
+  public not(): VectorFilterNotExpression {
+    return VectorFilterExpressionFactory.not(this);
+  }
```
malandis, line 31: These methods allow us to chain expressions with logical operators easier:

```typescript
const e1: VectorFilterExpression;
const e2: VectorFilterExpression;

e1.and(e2);
```

packages/core/src/messages/vector-index-filters.ts
```diff
@@ -0,0 +1,194 @@
...
+}
+
+/**
+ * Factory for creating vector filter expressions.
+ */
+export class VectorFilterExpressionFactory {
```
malandis, line 37: I added this factory class to ease the pain building a nontrivial expression. The hard way (long names and lots of `new`s):

```typescript
new VectorFilterAndExpression(new VectorFilterEqualsExpression('field1', 'value1'), new VectorFilterEqualsExpression('field2', 42))
```

And with these conveniences:
```typescript
import { VectorFilterExpressionFactory as F } from package;

F.equals('field1', 'value1').and(F.equals('field2', 42));
// ...or...
F.and(F.equals('field1', 'value1'), F.equals('field2', 42));
```

malandis (commented):
Some comments on design choices for discussion.

packages/client-sdk-nodejs/src/internal/vector-index-data-client.ts
cprice404: yep.

packages/common-integration-tests/src/vector-data-plane.ts
```diff
@@ -1023,6 +1024,127 @@ export function runVectorDataPlaneTest(
...
+          await sleep(2_000);
+
+          for (const {filterExpression, expectedIds, testCaseName} of [
+            {
+              filterExpression: F.equals('str', 'value1'),
+              expectedIds: ['test_item_1'],
+              testCaseName: 'string equality',
+            },
+            {
+              filterExpression: F.not(F.equals('str', 'value1')),
```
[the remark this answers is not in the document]
cprice404: this path seems great to me

packages/core/src/messages/vector-index-filters.ts
cprice404: I'm 💯 on having this factory; I think in modern languages/stdlibs, I've more frequently seen stuff like this exposed by just adding an `s` on the name of the thing, so like we did with `Configurations`, I'd probably name this `VectorFilterExpressions`.

cprice404 (dismissed):
lgtm, left one minor thought on naming that you can take or leave.

lmk if there was any other specific thing in the PR you were hoping for focus on.

packages/core/src/messages/vector-index-filters.ts
malandis: Done in [c4aeb19](https://github.com/momentohq/client-sdk-javascript/pull/1119/commits/c4aeb19f4b45d8c0343a7f98277c64dc9cfb058c)

cprice404 (approved):
filter

PR #1131, opened 2024-02-07
feat: add channel configuration based on environment

pratik151192:


packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -136,32 +138,54 @@ export class CacheDataClient implements IDataClient {
...
+
+      // Default values for keepalive options can be set here, if desired
+      channelOptions['grpc.keepalive_permit_without_calls'] =
+        channelConfig.keepAlivePermitWithoutCalls;
+      channelOptions['grpc.keepalive_timeout_ms'] =
+        channelConfig.keepAliveTimeoutMs;
+      channelOptions['grpc.keepalive_time_ms'] = channelConfig.keepAliveTimeMs;
+    }
+
+    let reinitializeClientEveryInterval = false;
```
pratik151192, line 172: we can move this to channel config too to avoid the type check below

packages/client-sdk-nodejs/src/config/configurations.ts
```diff
@@ -70,6 +70,14 @@ export class Laptop extends CacheConfiguration {
       transportStrategy: transportStrategy,
       middlewares: defaultMiddlewares,
       throwOnErrors: false,
+      channelConfiguration: {
+        maxSessionMemoryMB: grpcConfig.getMaxSessionMemoryMb(),
```
cprice404, line 74: mentioned this in slack but I'd rather put all the new configs into the existing `grpcConfig` instead of adding a new object/layer.

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -136,32 +138,54 @@ export class CacheDataClient implements IDataClient {
       `Creating cache client using endpoint: '${this.credentialProvider.getCacheEndpoint()}'`
     );
 
+    const channelConfig: ChannelConfiguration =
+      this.configuration.getChannelConfiguration();
+
+    const channelOptions: Record<string, number> = {
+      'grpc-node.max_session_memory': channelConfig.maxSessionMemoryMB,
+      'grpc.use_local_subchannel_pool': channelConfig.useLocalSubchannelPool
```
cprice404, line 146: i don't think there is a good reason to make this configurable right now.

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -136,32 +138,54 @@ export class CacheDataClient implements IDataClient {
...
+    if (channelConfig.useKeepAlive) {
+      // Assert that keepalive configurations are not undefined
+      if (
+        channelConfig.keepAlivePermitWithoutCalls === undefined ||
+        channelConfig.keepAliveTimeoutMs === undefined ||
+        channelConfig.keepAliveTimeMs === undefined
+      ) {
+        throw new Error(
+          'Keepalive settings must be defined when useKeepAlive is enabled.'
+        );
```
cprice404, line 161: tbh i'd rather make these grpc settings basically just pass-through; i don't think you need an additional `useKeepAlive`. If you really want to do that, I would make it an interface that contains the nested settings.

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
cprice404: yeah let's just keep this out of this PR for now I think?

packages/client-sdk-nodejs/src/internal/grpc/idle-grpc-client-wrapper.ts
```diff
@@ -53,6 +57,23 @@ export class IdleGrpcClientWrapper<T extends CloseableGrpcClient>
...
+          `Checking to see if client was created more than ${this.maxIdleMillis} ms`
+        );
+        this.logger.info(
+          'Client was created more than 4 minutes ago; reconnecting.'
+        );
+        this.client.close();
+        this.client = this.clientFactoryFn();
+        this.clientCreatedTime = Date.now();
+      }
+    }
```
cprice404, line 76: would prefer to add this stuff in a separate PR when its configuration options are added

packages/client-sdk-nodejs/src/config/configurations.ts
```diff
@@ -130,6 +135,11 @@ class InRegionDefault extends CacheConfiguration {
     const grpcConfig: GrpcConfiguration = new StaticGrpcConfiguration({
       deadlineMillis: deadlineMillis,
       maxSessionMemoryMb: defaultMaxSessionMemoryMb,
+      channelConfiguration: {
```
pratik151192, line 138: These were added for non-lambda configs

cprice404 (approved):

PR #1135, opened 2024-02-08
chore: add more logs to eager connection setup

pratik151192:


packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -242,10 +243,12 @@ export class CacheDataClient implements IDataClient {
...
             // so we must watch it twice.
+            this.logger.debug(`Connecting! Current state: ${newState}`);
             this.connectWithinDeadline(deadline).then(resolve).catch(reject);
           } else {
             this.logger.error(
-              `Unable to connect to Momento: Unexpected connection state: ${newState}. Please contact Momento if this persists.`
+              `Unable to connect to Momento: Unexpected connection state: ${newState}., oldState: ${currentState}
+              Please contact Momento if this persists.`
             );
             resolve();
```
cprice404, line 253: these look good; any reason not to add a log in the clientfactoryfn line 168? or in the connect fn line 193?

pratik151192: the one at 168 gets logged as a gRPC trace saying `Channel constructed` etc. but either way added at both places.

cprice404: thanks! i don't necessarily want to always have to rely on grpc trace logs so it's good to have as much as possible covered in our own.

cprice404 (approved):

PR #1136, opened 2024-02-09
refactor: use mvi filter fields for get/delete items

malandis:
We are in the process of migrating MVI delete item batch, get item batch, and get item metadata batch to use a filter expression to select items to delete (get) as opposed to a list of id's. For convenience, we will still allow users to select items with a list of id's from the top level methods. This PR migrates the backend.

As part of the migration we are also moving away from the hit/miss pattern for get item batch and get item metadata batch response messages. While the hit/miss pattern made sense for when we only selected items by id, it no longer makes sense for when a user supplies a general filter expression. What's more, previously when we read the hit/miss responses, we built an object containing only the hits, so the distinction wasn't as important.

In future PRs we will add overloads to also use a filter expression from delete and get item methods.

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/vector-index-data-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/vector-index-data-client.ts
@@ -810,7 +810,7 @@ export class VectorIndexDataClient implements IVectorIndexDataClient {
   ): Promise<VectorGetItemMetadataBatch.Response> {
     const request = new vectorindex._GetItemMetadataBatchRequest({
       index_name: indexName,
-      ids: ids,
+      filter: VectorIndexDataClient.idsToFilterExpression(ids),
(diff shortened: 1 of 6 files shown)


packages/client-sdk-nodejs/src/internal/vector-index-data-client.ts
```diff
@@ -713,7 +728,7 @@ export class VectorIndexDataClient implements IVectorIndexDataClient {
   ): Promise<VectorGetItemBatch.Response> {
     const request = new vectorindex._GetItemBatchRequest({
       index_name: indexName,
-      ids: ids,
+      filter: VectorIndexDataClient.idsToFilterExpression(ids),
```
malandis, line 731: Previously we allowed to select items to delete with a list of ids. We now have an id in set filter expression which is equivalent.

anitarua (approved):

PR #1138, opened 2024-02-09
chore: DRY up nodejs grpc channel configuration

cprice404:
Prior to this commit we had several different places in the code where
we were manually constructing an instance of grpc-js's `ChannelOptions`.

This commit creates a helper function to consolidate the construction
in a single spot; we accept an instance of our `GrpcConfiguraion` as input.
As we add support for more of the different grpc settings over time, this
will become increasingly more important.


pratik151192 (approved):
Nice!

PR #1139, opened 2024-02-09
fix: add support for item sizes up to 5 mb.

cprice404:
Prior to this commit, we were not explicitly setting the max size for grpc requests and responses. The default was 4mb, which meant that if a user had requested a limit increase to support 5mb item sizes, the SDK was not actually capable of serving those items back to them.

This commit adds configuration options to support setting different values for the max request and response size, and sets the defaults to 5mb to support those customers.


rishtigupta (approved):

pratik151192 (approved):

virratanasangpunth (dismissed):

cprice404:
had to rebase

pratik151192 (approved):

rishtigupta (approved):

PR #1141, opened 2024-02-12
refactor: rename mvi parameter ids to filter in get/delete

malandis:
To standardize how we select items in get item batch, get item metadata batch, and delete item batch, we use the term filter in methods, whether selecting by ids or filter expression.


bruuuuuuuce (approved):
:shipit:

bruuuuuuuce (approved):
:shipit:

PR #1148, opened 2024-02-14
chore: increase jest test timeout to reduce flaky tests

anitarua:
Saw a lot of jest test timeouts while collecting test failure data over the last couple of weeks, particularly in the web SDK. Doubled the timeout for both just in case


bruuuuuuuce (approved):
:shipit:

Issue #1147, opened 2024-02-14
TTL and CollectionTtl should validate to be a positive integer


pratik151192:
Unfortunately since TS/JS has a generic number type, customers can pass decimal in their TTL that surfaces as:

```
MomentoError: {\"_innerException\":{\"_transportDetails\":{\"grpc\":{\"code\":13,\"details\":\"13 INTERNAL: Request message serialization failure: Assertion failed\",\"metadata\":{}}},\"_errorCode\":\"INTERNAL_SERVER_ERROR\",\"_messageWrapper\":\"An unexpected error occurred while trying to fulfill the request; please contact us at support@momentohq.com\"}}" }
```

We should add client-side logic to validate using `Number.isInteger()` or something and wrap appropriate messaging around it.

We saw this with one customer who were programmatically computing the TTL which turned out to be in decimal.

Acceptance Criteria:

- [x] Scalar APis should call a common `validateTTL` that checks for positive integer.
- [x] CollectionTTL can be changed to add this check avoiding the need to add checks at each collection API

PR #1149, opened 2024-02-15
fix: ttl validator should check for positive integers

anitarua:
addresses https://github.com/momentohq/client-sdk-javascript/issues/1147

### DIFF
--- a/packages/core/src/internal/utils/validators.ts
+++ b/packages/core/src/internal/utils/validators.ts
@@ -126,7 +126,9 @@ export function validateTtlMinutes(ttlMinutes: number) {
 
 export function validateTtl(ttl: number) {
   if (ttl < 0 || !Number.isInteger(ttl)) {
-    throw new InvalidArgumentError('ttl must be a positive integer');
+    throw new InvalidArgumentError(
+      `ttl must be a positive integer, received ${ttl}`
+    );
   }
 }
 
--- a/packages/core/src/internal/utils/validators.ts
+++ b/packages/core/src/internal/utils/validators.ts
@@ -124,6 +124,8 @@ export function validateTtlMinutes(ttlMinutes: number) {
   }
 }
 
+// Currently accepting ttl in seconds, but if someone requests support for millis,
+// the validator will need to check for Number.isSafeInteger(ttl * 1000).
 export function validateTtlSeconds(ttl: number) {
   if (ttl < 0 || !Number.isSafeInteger(ttl)) {
     throw new InvalidArgumentError(
--- a/packages/core/src/internal/utils/validators.ts
+++ b/packages/core/src/internal/utils/validators.ts
@@ -119,7 +119,7 @@ export function validateTopK(topK: number) {
 }
 
 export function validateTtlMinutes(ttlMinutes: number) {
-  if (ttlMinutes < 0 || !Number.isInteger(ttlMinutes)) {
+  if (ttlMinutes < 0 || !Number.isSafeInteger(ttlMinutes)) {
     throw new InvalidArgumentError('ttlMinutes must be a positive integer');
   }
 }
--- a/packages/core/src/internal/utils/validators.ts
+++ b/packages/core/src/internal/utils/validators.ts
@@ -119,8 +119,14 @@ export function validateTopK(topK: number) {
 }
 
 export function validateTtlMinutes(ttlMinutes: number) {
-  if (ttlMinutes < 0) {
-    throw new InvalidArgumentError('ttlMinutes must be positive');
+  if (ttlMinutes < 0 || !Number.isInteger(ttlMinutes)) {
+    throw new InvalidArgumentError('ttlMinutes must be a positive integer');
+  }
+}
+
+export function validateTtl(ttl: number) {
+  if (ttl < 0 || !Number.isInteger(ttl)) {
+    throw new InvalidArgumentError('ttl must be a positive integer');
   }
 }
 
--- a/packages/core/src/internal/utils/validators.ts
+++ b/packages/core/src/internal/utils/validators.ts
@@ -124,8 +124,8 @@ export function validateTtlMinutes(ttlMinutes: number) {
   }
 }
 
-export function validateTtl(ttl: number) {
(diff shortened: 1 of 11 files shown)


packages/core/test/unit/utils/collection-ttl.test.ts
```diff
@@ -59,9 +71,9 @@ describe('CollectionTtl', () => {
 
   describe('#ttlMilliseconds', () => {
     test('converts to milliseconds', () => {
-      const cttl = new CollectionTtl(1.23);
+      const cttl = new CollectionTtl(2);
 
-      expect(cttl.ttlMilliseconds()).toEqual(1230);
+      expect(cttl.ttlMilliseconds()).toEqual(2000);
     });
```
anitarua, line 77: not sure if this was the right move, should we still allow collectionTtl to be specified as a float?

packages/core/test/unit/utils/collection-ttl.test.ts
```diff
@@ -19,6 +19,18 @@ describe('CollectionTtl', () => {
...
+      expect(() => {
+        CollectionTtl.of(-1);
+      }).toThrow('ttl must be a positive integer');
+    });
+
+    it('does not allow float ttl', () => {
+      expect(() => {
+        CollectionTtl.of(1.5);
+      }).toThrow('ttl must be a positive integer');
+    });
```
anitarua, line 33: previously allowed floating point collectionTtl, should that still be the case?

packages/core/src/internal/utils/validators.ts
```diff
@@ -119,8 +119,16 @@ export function validateTopK(topK: number) {
...
 export function validateTtlMinutes(ttlMinutes: number) {
-  if (ttlMinutes < 0) {
-    throw new InvalidArgumentError('ttlMinutes must be positive');
+  if (ttlMinutes < 0 || !Number.isInteger(ttlMinutes)) {
+    throw new InvalidArgumentError('ttlMinutes must be a positive integer');
+  }
+}
+
+export function validateTtl(ttl: number) {
+  if (ttl < 0 || !Number.isInteger(ttl)) {
```
pratik151192, line 128: Should we use `Number.isSafeInteger` instead? I am not sure if a `ttl` beyond Int.MAX_VALUE is acceptable and if not `isSafeInteger` may be a better check. cc @cprice404

packages/core/test/unit/utils/collection-ttl.test.ts
pratik151192: Yeah previously the SDK allowed it but it fails at the gRPC serialization layer which surfaces as an `INTERNAL_ERROR`. There's an interesting case though for which I'll tag @cprice404 again as I am not sure why its happening.

A floating point upto three decimal places is being accepted. So

```
await momento.set('cache', 'foo', 'FOO', {ttl: 30.411});
```

succeeds, but

```
await momento.set('cache', 'foo', 'FOO', {ttl: 30.4115});
```

fails with an `Request message serialization failure: Assertion failed` error. 

Same happens with `CollectionTtl` as well. Either way, I don't think a decimal ttl makes seconds and we call it as `ttlSeconds` in a bunch of places so I'd go ahead with what you have in the PR.

packages/core/src/internal/utils/validators.ts
```diff
@@ -119,8 +119,16 @@ export function validateTopK(topK: number) {
...
 
 export function validateTtlMinutes(ttlMinutes: number) {
-  if (ttlMinutes < 0) {
-    throw new InvalidArgumentError('ttlMinutes must be positive');
+  if (ttlMinutes < 0 || !Number.isInteger(ttlMinutes)) {
+    throw new InvalidArgumentError('ttlMinutes must be a positive integer');
+  }
+}
+
+export function validateTtl(ttl: number) {
```
pratik151192, line 127: Since we have a methodName called `validateTtlMinutes`, I'd call this `validateTtlSeconds` just for parity with units being present in the name.

pratik151192 (commented):
LGTM! Just a couple minor comments

packages/core/src/internal/utils/validators.ts
```diff
@@ -119,8 +119,16 @@ export function validateTopK(topK: number) {
 }
 
 export function validateTtlMinutes(ttlMinutes: number) {
-  if (ttlMinutes < 0) {
-    throw new InvalidArgumentError('ttlMinutes must be positive');
+  if (ttlMinutes < 0 || !Number.isInteger(ttlMinutes)) {
```
pratik151192, line 122: I think we missed this for `safeInteger`.. dont want to nitpick but just to be consistent

pratik151192 (commented):
I forgot about this during the first review but should we also add this check in the `cache-client` constructor. I don't see a validation there for `defaultTtlSeconds` in `CacheClientProps`. So in theory we could give the cache default Ttl to be float and all cache requests will fail with the serialization error (I repro'd this)

packages/core/src/internal/utils/validators.ts
anitarua: oh yes, good catch, thanks!

packages/core/test/unit/utils/collection-ttl.test.ts
cprice404: oh..... I think I understand the reason why this is failing after 3 decimal places. I think the wire format transports the value as a number of milliseconds, not a number of seconds. So there is probably a `* 1000` in the code before we serialize.

We may have a user ask us to support millis. But I think for now let's just do what you are doing in this PR @anitarua , and force an integer. If someone requests support for millis we'll just have to change your validation function to do `* 1000` and then the integer check afterward.

cprice404 (approved):

packages/core/test/unit/utils/collection-ttl.test.ts
cprice404: maybe we should leave a comment to this effect in the validateTtl function to make it obvious to the next developer if this does come up again.

pratik151192 (approved):

cprice404 (approved):

PR #1151, opened 2024-02-20
chore: increase jest and test cache client timeouts

anitarua:
Was still seeing occasional test failures related to jest test timeouts, mostly for vector tests in the web sdk, but doubling the timeout again here in both just in case. 

Also increasing the cache client timeout by 30 seconds after seeing timeout errors from different tests, figured there might be enough tests added recently that increasing the timeout a bit might be worth a shot.

### DIFF
--- a/packages/client-sdk-nodejs/jest.config.ts
+++ b/packages/client-sdk-nodejs/jest.config.ts
@@ -8,7 +8,7 @@ const config: Config = {
   transform: {
     '^.+\\.tsx?$': 'ts-jest',
   },
-  testTimeout: 240000,
+  testTimeout: 480000,
   reporters: ["jest-ci-spec-reporter"]
 };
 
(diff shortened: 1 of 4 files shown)


malandis (approved):
There are two changes here: (1) the jest test timeouts and (2) the client timeouts.

Regarding (1), we have bumped the jest timeouts [before](https://github.com/momentohq/client-sdk-javascript/pull/1005), so I suspect this is a downstream problem and not the root cause.

Do you think (2) is related or a different cause? 90s is pretty long.

Not blocking comments.

anitarua:
(1) Very possible it's a different problem instead, though I interpreted it to mean I didn't increase the timeout enough the first time. When running the tests locally, the vector tests take about 232000ms for node and 254500ms for web. I figured doubling it again to 480000 will be more than sufficient and reduce the likelihood of having to increase that timeout again soon. 

(2) The cache client timeout errors were from non-vector tests, so I think unrelated. I agree 90s seems pretty long but thought increasing it was worth a shot, just in case it helps out the webhooks and fgac tests (the ones that more often fail with timeout errors).

anitarua:
Oh huh, I didn't think it was per test, but it's true the jest docs are pretty vague about that setting lol

> `testTimeout` -- Default timeout of a test in milliseconds

Issue #349, opened 2023-03-16
pubsub followup: middleware investigation


malandis:
The unary middlewares do not work with the subscribe streaming response. Figure out what exactly is incompatible, then decide on the correct set of middlewares.

PR #1153, opened 2024-02-22
feat: setBatch and getBatch

anitarua:
Adds batch get and set APIs for [#674](https://github.com/momentohq/dev-eco-issue-tracker/issues/674)

Note: the streaming APIs use a different set of interceptors, this issue is still open: https://github.com/momentohq/client-sdk-javascript/issues/349 

Another note: this PR doesn't remove the [nodejs batchutils functions](https://github.com/momentohq/client-sdk-javascript/tree/main/packages/client-sdk-nodejs/src/batchutils)

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/cache-data-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/cache-data-client.ts
@@ -65,6 +65,8 @@ import {
   MomentoLoggerFactory,
   SortedSetOrder,
   UnknownError,
+  GetBatch,
+  SetBatch,
 } from '..';
 import {version} from '../../package.json';
 import {IdleGrpcClientWrapper} from './grpc/idle-grpc-client-wrapper';
@@ -104,7 +106,6 @@ import _Unbounded = cache_client._Unbounded;
 import ECacheResult = cache_client.ECacheResult;
 import _ItemGetTypeResponse = cache_client._ItemGetTypeResponse;
 import {grpcChannelOptionsFromGrpcConfig} from './grpc/grpc-channel-options';
-import {GetBatch, SetBatch} from '@gomomento/sdk-core';
 
 export const CONNECTION_ID_KEY = Symbol('connectionID');
 
@@ -119,6 +120,7 @@ export class CacheDataClient implements IDataClient {
   private readonly logger: MomentoLogger;
   private readonly cacheServiceErrorMapper: CacheServiceErrorMapper;
   private readonly interceptors: Interceptor[];
+  private readonly streamingInterceptors: Interceptor[];
 
   /**
    * @param {CacheClientProps} props
@@ -171,11 +173,19 @@ export class CacheDataClient implements IDataClient {
     // to be able to set it.
     const context: MiddlewareRequestHandlerContext = {};
     context[CONNECTION_ID_KEY] = dataClientID;
+
+    const headers = [
+      new Header('Authorization', this.credentialProvider.getAuthToken()),
+      new Header('Agent', `nodejs:${version}`),
+    ];
+
     this.interceptors = this.initializeInterceptors(
+      headers,
       this.configuration.getLoggerFactory(),
       this.configuration.getMiddlewares(),
       context
     );
+    this.streamingInterceptors = this.initializeStreamingInterceptors(headers);
   }
   public connect(timeoutSeconds = 10): Promise<void> {
     this.logger.debug('Attempting to eagerly connect to channel');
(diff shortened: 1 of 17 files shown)


packages/core/src/clients/IMomentoCache.ts
```diff
@@ -96,6 +99,13 @@ export interface IMomentoCache {
     field: string | Uint8Array,
     options?: SetIfNotExistsOptions
   ): Promise<CacheSetIfNotExists.Response>;
+  getBatch(keys: string[] | Uint8Array[]): Promise<GetBatch.Response>;
+  setBatch(
+    items:
+      | Record<string, string | Uint8Array>
+      | Map<string | Uint8Array, string | Uint8Array>,
```
cprice404, line 106: i like how you did the types here to support both Record and Map!

packages/core/src/internal/clients/cache/AbstractCacheClient.ts
```diff
@@ -209,6 +212,46 @@ export abstract class AbstractCacheClient implements ICacheClient {
...
+   *
+   * @param {string} cacheName - The cache to perform the lookup in.
+   * @param {string[] | Uint8Array[]} keys - The list of keys to look up.
+   * @returns {Promise<GetBatch.Response>} -
+   * {@link GetBatch.Success} containing the values if they were found.
+   * {@link GetBatch.Error} on failure.
+   */
+  public async getBatch(
+    cacheName: string,
+    keys: string[] | Uint8Array[]
```
cprice404, line 226: I think that theoretically this could support `List<string | Uint8Array>` (mixed key types).

I don't think it's super important for us to support this but if it's easy it'd be nice.

packages/core/src/messages/responses/cache-batch-get.ts
```diff
@@ -0,0 +1,97 @@
...
+    super();
+    this.body = body;
+  }
+
+  /**
+   * Returns the data as a list of utf-8 strings, decoded from the underlying byte array.
+   * May contain nulls corresponding to cache misses.
+   * @returns {(string | undefined)[]}
+   */
+  public values(): (string | undefined)[] {
```
cprice404, line 53: I wonder if we should consider making the return type for this be a `Record<string, string | uint>`, or a `Map`.

I can definitely imagine cases where a user would just want them in order, but they could maybe achieve that sufficiently through `results`. It feels like the more natural / common / simple use case (which is what we want to prioritize for this `.values` function) might be to get back a structure that they can index into by key.

would really love to get other folks thoughts on this though, especially since is the first major SDK we're releasing this in and we will be treating it as a canonical example.

cprice404 (commented):
Overall this looks great! I just had a few minor questions / thoughts on the API. We should ask for a few more 👀  / opinions.

packages/core/src/messages/responses/cache-batch-get.ts
anitarua: yeah I like that idea considering they provided a Record or Map, they're probably expecting a similar structure back.  

I had missed this before but it seems like [dictionaryGetFields](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/core/src/messages/responses/cache-dictionary-get-fields.ts) returns items as Record or Map as well

malandis: 💯 on the record/map as the default. In the past when we have exposed an array (eg in earlier iterations of `dictionaryGetFields`), I believe we found the record/map easier to use. My intuition is users reach for the record/map first as well.

pgautier404 (approved):
LGTM!

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -720,6 +736,165 @@ export class CacheDataClient implements IDataClient {
...
+      });
+
+      call.on('error', (err: ServiceError | null) => {
+        this.cacheServiceErrorMapper.resolveOrRejectError({
+          err: err,
+          errorResponseFactoryFn: e => new GetBatch.Error(e),
+          resolveFn: resolve,
+          rejectFn: reject,
+        });
+      });
```
malandis, line 809: For the backlog: (1) the three callbacks -- data, end and error -- have boilerplate we can abstract away; and (2) we should require the developer to implement all three.

Otherwise 👍

malandis (approved):

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
cprice404: @malandis wdym? is your comment targeted at future streaming APIs?

cprice404: or just consolidating the get/set ones I guess?

cprice404 (approved):
batch

PR #1155, opened 2024-02-27
chore: update log level during timeout to warn

pratik151192:
This commit will allow us to look at the log for the gRPC connection status when a deadline/timeout is observed


packages/client-sdk-nodejs/src/internal/grpc/middlewares-interceptor.ts
```diff
@@ -86,8 +86,8 @@ export function middlewaresInterceptor(
               // also try to connect if it's idle, false will just get the status
               const connectionStatus =
                 grpcClient?.getChannel()?.getConnectivityState(false) ?? null;
-              logger.debug(
-                `Received status: ${status.code} ${
+              logger.warn(
+                `Deadline Exceeded! Received status: ${status.code} ${
```
pratik151192, line 90: verified that the interceptor is consumed by the leaderboard data client

anitarua (approved):

PR #1168, opened 2024-03-06
feat: add close method to clients

pratik151192:
Adds close method to cache clients and middlewares. This allows for proper cleanup of resources, and also doesn't leave any background intervals lurking.

### DIFF
--- a/packages/client-sdk-nodejs/src/config/middleware/experimental-metrics-logging-middleware.ts
+++ b/packages/client-sdk-nodejs/src/config/middleware/experimental-metrics-logging-middleware.ts
@@ -105,6 +105,9 @@ export class ExperimentalMetricsLoggingMiddleware extends ExperimentalMetricsMid
   private static elu: EventLoopUtilization;
   private static isLoggingStarted = false;
   static numActiveRequests = 0;
+  // this is typed as any because JS returns a number for intervalId but
+  // TS returns a NodeJS.Timeout.
+  private static intervalId: any | null = null; // Store the interval ID
 
   constructor(loggerFactory: MomentoLoggerFactory) {
     super(
@@ -125,7 +128,7 @@ export class ExperimentalMetricsLoggingMiddleware extends ExperimentalMetricsMid
     this.eldMonitor.enable();
     this.elu = performance.eventLoopUtilization();
 
-    setInterval(() => {
+    this.intervalId = setInterval(() => {
       this.elu = performance.eventLoopUtilization(this.elu);
       const metrics: StateMetrics = {
         eventLoopUtilization: this.elu.utilization,
@@ -141,5 +144,16 @@ export class ExperimentalMetricsLoggingMiddleware extends ExperimentalMetricsMid
       logger.info(JSON.stringify(metrics));
       this.eldMonitor.reset();
     }, this.metricsLogInterval);
+  }
+
+  close() {
+    if (ExperimentalMetricsLoggingMiddleware.intervalId) {
+      this.logger.debug('Stopping Metrics logging.');
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+      clearInterval(ExperimentalMetricsLoggingMiddleware.intervalId);
+      ExperimentalMetricsLoggingMiddleware.intervalId = null;
+      ExperimentalMetricsLoggingMiddleware.isLoggingStarted = false;
+      this.logger.debug('Metrics logging stopped.');
+    }
   }
 }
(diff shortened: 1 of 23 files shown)


packages/client-sdk-nodejs/src/preview-leaderboard-client.ts
```diff
@@ -30,12 +31,22 @@ export class PreviewLeaderboardClient implements ILeaderboardClient {
...
+    this.dataClient.close();
+    this.configuration.getMiddlewares().forEach(m => {
+      if (m.close) {
+        m.close();
+      }
+    });
+  }
+
   /**
    * Creates an instance of LeaderboardClient with 32-bit float scores.
```
anitarua, line 51: lol could you correct this comment while you're in here? should be 64-bit float scores now

> Creates an instance of LeaderboardClient with 64-bit float scores.

packages/client-sdk-nodejs/src/config/middleware/experimental-metrics-logging-middleware.ts
```diff
@@ -105,6 +105,9 @@ export class ExperimentalMetricsLoggingMiddleware extends ExperimentalMetricsMid
   private static elu: EventLoopUtilization;
   private static isLoggingStarted = false;
   static numActiveRequests = 0;
+  // this is typed as any because JS returns a number for intervalId but
+  // TS returns a NodeJS.Timeout.
+  private static intervalId: any | null = null; // Store the interval ID
```
anitarua, line 110: I think you can turn off the eslint warning if you know you need to use `any` for this to work, something like so?

```
private static intervalId: any | null = null; // eslint-disable no-explicit-any
```

anitarua (approved):
lgtm!

packages/client-sdk-nodejs/src/config/middleware/experimental-metrics-logging-middleware.ts
```diff
@@ -105,6 +105,8 @@ export class ExperimentalMetricsLoggingMiddleware extends ExperimentalMetricsMid
   private static elu: EventLoopUtilization;
   private static isLoggingStarted = false;
   static numActiveRequests = 0;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private static intervalId: any | null = null; // Store the interval ID
```
cprice404, line 109: looking at this code, i'm not totally sure why all of these variables are static, but i guess we don't need to deal with that now.

packages/client-sdk-nodejs/test/integration/cache-client-close.test.ts
```diff
@@ -0,0 +1,23 @@
+import {expectWithMessage} from '@gomomento/common-integration-tests';
+
+import {GetBatch, ICacheClient} from '@gomomento/sdk-core';
+import {SetupIntegrationTestWithMiddleware} from './integration-setup';
+
+const {cacheClient, cacheName} = SetupIntegrationTestWithMiddleware();
+
+runTestWithClientCloseToMakeSureTestDoesntRunForever(cacheClient);
+export function runTestWithClientCloseToMakeSureTestDoesntRunForever(
```
cprice404, line 9: not necessarily a problem but I'm curious what the value is of having a function here rather than just having `describe` at the top level like it is for most tests?

packages/common-integration-tests/src/create-delete-list-cache.ts
```diff
@@ -62,9 +62,9 @@ export function runCreateDeleteListCacheTests(cacheClient: ICacheClient) {
           expect(knownCaches.length === 1).toBeTrue();
           const cache = knownCaches[0];
 
-          const expectedThroughputLimit = 1024;
-          const expectedItemSizeLimit = 1024;
-          const expectedThrottlingLimit = 100;
+          const expectedThroughputLimit = 10240;
+          const expectedItemSizeLimit = 4883;
+          const expectedThrottlingLimit = 500;
```
cprice404, line 67: I think you probably don't want to commit these?

cprice404 (commented):
just a few nits. thanks for doing this!

packages/client-sdk-nodejs/src/config/middleware/experimental-metrics-logging-middleware.ts
pratik151192: I am going to move this to another middleware so will address it for sure on that PR!

packages/client-sdk-nodejs/test/integration/cache-client-close.test.ts
pratik151192: ah yeah i didn't put a lot of thought into that honestly will address

packages/common-integration-tests/src/create-delete-list-cache.ts
pratik151192: I do!! We have test failures otherwise as we recently bumped up these limits.

anitarua (approved):

PR #1169, opened 2024-03-06
chore: make sure nodejs can set and get 5mb messages

anitarua:
Sets and gets a string with 5million bytes (~5mb)


bruuuuuuuce (approved):
:shipit:

PR #1165, opened 2024-03-05
chore: temporarily allow >= for expected throttling limits

cprice404:
We have a test that makes assertions about the cache limits configuration,
both to provide test coverage of the control plane API and to ensure that
we have a sufficient, uniform configuration across the canary environments.

We are currently in the process of raising the limits to allow more parallelization
of tests. This PR temporarily relaxes the constraint for the TPS limit from
equality to `>=` to allow us some flexibility until we know the final value.

PR #1170, opened 2024-03-06
chore: add endpoint prefix

pgautier404:
Although the Web SDK contains logic to automatically prefix its endpoints with `web.` when computing them from an API key, it declines to apply that logic to endpoint overrides passed in by the user. This is problematic when using a `BaseEndpointOverride`, where endpoint URIs are dynamically computed from a common base string. This commit extends the `BaseEndpointOverride` to include an `endpointPrefix`, allowing users to supply a prefix top be prepended to each computed URI.

Similar to #1165 this PR relaxes a test assertion about throttling limits. This is intended to be a temporary change while the limits in the test environment are in flux.

### DIFF
--- a/packages/common-integration-tests/src/create-delete-list-cache.ts
+++ b/packages/common-integration-tests/src/create-delete-list-cache.ts
@@ -74,9 +74,9 @@ export function runCreateDeleteListCacheTests(cacheClient: ICacheClient) {
 
           // checking that cache limits are equal to or greater than default limits
           expectWithMessage(() => {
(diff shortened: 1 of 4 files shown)


pgautier404:
> @pgautier404 just to make sure I understand - is it the case that the console is always passing in BaseEndpointOverride?

Yes. Corresponding console PR is [here](https://github.com/momentohq/momento-console/pull/636).

cprice404 (dismissed):
lgtm if it looks good to matt/rishti

bruuuuuuuce (approved):
:shipit:

PR #1171, opened 2024-03-07
chore: decouple event loop metrics into its own middleware

pratik151192:
This PR extracts the event loop metrics from the metrics logging middleware. The metrics logging middleware is per request and can take a toll on customer's CW bills. We'd want to sometimes measure event loop alone without needing request level metrics so a separate middleware is better for that. A factory method is provided for different options to enable the kind of metrics the user wants, to avoid initializing the middleware for each one.

Usage:

```
const middlewares: Middleware[] = MiddlewareFactory.createMetricsMiddleware(middlewaresExampleloggerFactory, {
    eventLoopMetrics: true,
    activeRequestCountMetrics: true,
    perRequestMetrics: true,
  });
  const middlewaresExampleClient = await CacheClient.create({
    configuration: Configurations.Laptop.v1(middlewaresExampleloggerFactory)
      .withMiddlewares(middlewares),
    credentialProvider: CredentialProvider.fromEnvironmentVariable({
      environmentVariableName: 'MOMENTO_API_KEY',
    }),
    defaultTtlSeconds: 60,
  });
```

### DIFF
--- a/packages/client-sdk-nodejs/src/config/middleware/experimental-event-loop-perf-middleware.ts
+++ b/packages/client-sdk-nodejs/src/config/middleware/experimental-event-loop-perf-middleware.ts
@@ -1,0 +1,163 @@
+import {
+  Middleware,
+  MiddlewareMessage,
+  MiddlewareMetadata,
+  MiddlewareRequestHandler,
+  MiddlewareStatus,
+} from './middleware';
+import {
+  EventLoopDelayMonitor,
+  EventLoopUtilization,
+  monitorEventLoopDelay,
+  performance,
+} from 'perf_hooks';
+import {MomentoLogger, MomentoLoggerFactory} from '@gomomento/sdk-core';
+
+interface StateMetrics {
+  /**
+   * The proportion of time the event loop was busy over the last second.
+   */
+  eventLoopUtilization: number;
+
+  /**
+   * The average event loop delay over the last second, measured in 20ms increments.
+   */
+  eventLoopDelayMean: number;
+
+  /**
+   * The minimum event loop delay over the last second, measured in 20ms increments.
+   */
+  eventLoopDelayMin: number;
+
+  /**
+   * The 50th percentile event loop delay over the last second, measured in 20ms increments.
+   */
+  eventLoopDelayP50: number;
+
+  /**
+   * The 75th percentile event loop delay over the last second, measured in 20ms increments.
+   */
+  eventLoopDelayP75: number;
+
+  /**
+   * The 90th percentile event loop delay over the last second, measured in 20ms increments.
+   */
+  eventLoopDelayP90: number;
+
+  /**
+   * The 95th percentile event loop delay over the last second, measured in 20ms increments.
+   */
+  eventLoopDelayP95: number;
+
+  /**
+   * The 99th percentile event loop delay over the last second, measured in 20ms increments.
+   */
+  eventLoopDelayP99: number;
+
+  /**
+   * The maximum event loop delay over the last second, measured in 20ms increments.
+   */
+  eventLoopDelayMax: number;
+
+  /**
+   * The timestamp when the state metrics got recorded
+   */
+  timestamp: number;
+}
+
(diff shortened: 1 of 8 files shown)


packages/client-sdk-nodejs/src/config/middleware/experimental-event-loop-perf-middleware.ts
```diff
@@ -0,0 +1,156 @@
...
+  implements Middleware
+{
+  private readonly metricsLogInterval = 1000;
+  private readonly eventLoopDelayInterval = 20;
+  private eldMonitor: EventLoopDelayMonitor;
+  private elu: EventLoopUtilization;
+  private isLoggingStarted = false;
+  // this is typed as any because JS returns a number for intervalId but
+  // TS returns a NodeJS.Timeout.
+  private intervalId: any | null = null; // Store the interval ID
```
anitarua, line 106: nit: add es-lint comment to disable the github actions warnings

anitarua (dismissed):
lgtm! great idea!

cprice404 (commented):
discussed on zoom - I really like the idea of being able to separate these but I think it might be nicer for end users if we just kept a single Middleware, and allowed users to pass an optional constructor argument to it to enable/disable each of these 3 things independently:

* per-request metrics
* periodic event loop metrics
* periodic "active request count" metrics

lmk if that change is feasible without too much re-work

pratik151192:
@cprice404 I have made the changes but instead of squishing those 3 things in a single middleware, I instead provided a factory for the metrics middleware with toggling options for the 3 of them

anitarua (dismissed):
lgtm!

packages/client-sdk-nodejs/src/config/middleware/experimental-middleware-factory.ts
```diff
@@ -0,0 +1,48 @@
...
+import {MomentoLoggerFactory} from '@gomomento/sdk-core';
+import {ExperimentalEventLoopPerformanceMetricsMiddleware} from './experimental-event-loop-perf-middleware';
+import {ExperimentalMetricsLoggingMiddleware} from './experimental-metrics-logging-middleware';
+import {ExperimentalActiveRequestCountLoggingMiddleware} from './experimental-active-request-count-middleware';
+import {Middleware} from './middleware';
+import {ExperimentalMetricsCsvMiddleware} from './experimental-metrics-csv-middleware';
+
+interface ExperimenalMetricsMiddlewareOptions {
+  eventLoopMetrics?: boolean;
+  perRequestMetrics?: boolean;
```
cprice404, line 11: nit: it;s not clear from these names whether the two "per request" flags are related, or where the metrics for the first three options will output their metrics. maybe add the word "logging" or "log" to the 3 arg names, or somewhere in this new factory api?

packages/client-sdk-nodejs/test/integration/cache-client-close.test.ts
```diff
@@ -1,18 +1,33 @@
...
+
+function integrationTestCacheClientPropsWithExperimentalMetricsMiddleware(): CacheClientPropsWithConfig {
+  const loggerFactory = new DefaultMomentoLoggerFactory(
+    DefaultMomentoLoggerLevel.INFO
+  );
+  return {
+    configuration: Configurations.Laptop.latest(loggerFactory)
+      .withClientTimeoutMillis(90000)
+      .withMiddlewares([
+        new ExperimentalEventLoopPerformanceMetricsMiddleware(loggerFactory),
```
cprice404, line 28: might as well exercise your factory here?

cprice404 (commented):
this looks great, just a few minor comments

packages/client-sdk-nodejs/test/integration/cache-client-close.test.ts
pratik151192: yes for sure! had also thought to update `advanced` examples with factory but here makes sense too

cprice404 (approved):
yes

PR #1177, opened 2024-03-12
chore: reject promise if eager connection fails

pratik151192:
Prior to this PR, when an eager connection failed to Momento, we logged an error and moved on to data plane requests. The eager connection waits for 10 seconds to establish a connection, and if it wasn't able to do so, its highly likely that a request might fail due to a timeout or unavailable exception. Timeouts have idempotency issues with retries, so it's better if we make the eager connection failure a hard failure, so that lambdas can be retried safely and/or deployments can handle connection failures. In this PR we throw a hard failure on eager connection failures.


packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -218,8 +218,9 @@ export class CacheDataClient implements IDataClient {
       const now = new Date();
 
       if (now >= deadline) {
-        this.logger.error('Unable to connect to Momento: deadline exceeded.');
-        resolve();
+        const errorMessage = 'Unable to connect to Momento: deadline exceeded.';
+        this.logger.error(errorMessage);
+        reject(new Error(errorMessage));
```
cprice404, line 223: can we throw one of our sdk errors here instead of a generic error?

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -228,14 +229,13 @@ export class CacheDataClient implements IDataClient {
...
-              }`
-            );
-            resolve();
+            const errorMessage = `Unable to eagerly connect to Momento. Please contact Momento if this persists. currentState: ${currentState}, errorName: ${
+              error.name
+            } : errorMessage: ${error.message}, errorStack: ${
+              error.stack ? error.stack : 'Stack trace undefined'
+            }`;
+            this.logger.error(errorMessage);
+            reject(new Error(errorMessage));
```
cprice404, line 238: will this be the same error that we rejected with earlier? if so, can we just do `reject(error)`? If not, can we use one of our SDK error types here too? (or create a new one, I think I did that for the .NET SDK.)

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -253,11 +253,11 @@ export class CacheDataClient implements IDataClient {
...
           } else {
-            this.logger.error(
-              `Unable to connect to Momento: Unexpected connection state: ${newState}., oldState: ${currentState}
-              Please contact Momento if this persists.`
-            );
-            resolve();
+            const errorMessage = `Unable to connect to Momento: Unexpected connection state: ${newState}., oldState: ${currentState}
+              Please contact Momento if this persists.`;
+            this.logger.error(errorMessage);
+            reject(new Error(errorMessage));
```
cprice404, line 259: same comment here re: error type

packages/client-sdk-nodejs/test/unit/cache-client.test.ts
```diff
@@ -32,7 +32,9 @@ describe('CacheClient', () => {
     const invalidCacheNames = ['', '    '];
     const momento = await CacheClient.create({
       configuration: configuration,
-      credentialProvider: credentialProvider,
+      credentialProvider: CredentialProvider.fromEnvironmentVariable({
+        environmentVariableName: 'TEST_AUTH_TOKEN',
```
cprice404, line 36: why is this changing?

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
pratik151192: yeah added!

packages/client-sdk-nodejs/test/unit/cache-client.test.ts
pratik151192: This test used to exercise that we get a 4xx if the cache name is invalid. However, it passed an invalid auth token to the client construction. Previously that returned an invalid connection but never hard failed. The test was happy as it could still  call create cache with invalid name and get a 4xx even before the connection was used. Now that the connection had failed to establish, it won't move on from that hard failure. So i just fixed the token here as the test's objective was to assert 4xx on invalid cache names.

cprice404: but this test is a unit test, so it's not supposed to make any network requests. so you shouldn't need a real token here. if the functionality that you want to test requires a real token then we should move it to an int test.

cprice404 (approved):

PR #1188, opened 2024-03-14
chore: reduce jest timeout, auto-retry flaky vector tests in web sdk

anitarua:
Dialed down the jest timeout because increasing it a bunch didn't actually help.

Added jest directive to auto-retry tests up to 2 times in the flakiest `describe` block (vector 'getItemBatch and getItemMetadataBatch' tests) in the web sdk. These tests failed most often in the canaries and during bakes.

Wasn't able to reproduce the jest timeouts locally (only once with no helpful logs), but saw `Http response at 400 or 500 level, http status code: 0` quite a lot. That error is labeled as a [browser disconnect error](https://github.com/momentohq/client-sdk-javascript/blob/29de9d6602b00610cc697a0b0e64ad39481c6e11/packages/client-sdk-web/src/internal/pubsub-client.ts#L42) in the pubsub client, couldn't find any grpc-related info on this error. ~~Decided to upgrade the `happy-dom/jest-environment` dependency in case that helps~~ -> saw lots of that error in CI once that went in, decided to revert for now.

### DIFF
--- a/packages/client-sdk-web/jest.config.ts
+++ b/packages/client-sdk-web/jest.config.ts
@@ -14,7 +14,7 @@ const config: Config = {
     // Force module uuid to resolve with the CJS entry point, because Jest does not support package.json.exports. See https://github.com/uuidjs/uuid/issues/451
     "uuid": require.resolve('uuid'),
   },
-  // reporters: ["jest-ci-spec-reporter"]
+  reporters: ["jest-ci-spec-reporter"]
 };
 
 export default config;
--- a/packages/client-sdk-web/jest.config.ts
+++ b/packages/client-sdk-web/jest.config.ts
@@ -8,13 +8,13 @@ const config: Config = {
   transform: {
     '^.+\\.tsx?$': 'ts-jest',
   },
-  testTimeout: 480000,
+  testTimeout: 120000,
   setupFiles: ['<rootDir>/jest.setup.js'],
   moduleNameMapper: {
     // Force module uuid to resolve with the CJS entry point, because Jest does not support package.json.exports. See https://github.com/uuidjs/uuid/issues/451
     "uuid": require.resolve('uuid'),
   },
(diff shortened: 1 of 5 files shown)


packages/common-integration-tests/src/vector-data-plane.ts
```diff
@@ -1384,6 +1384,9 @@ export function runVectorDataPlaneTest(
   });
 
   describe('getItemBatch and getItemMetadataBatch', () => {
+    // Automatically retry flaky tests in this describe block twice before failing
+    jest.retryTimes(2, {logErrorsBeforeRetry: true});
```
bruuuuuuuce, line 1388: 🦾

packages/client-sdk-web/jest.config.ts
```diff
@@ -8,13 +8,13 @@ const config: Config = {
...
   },
-  testTimeout: 480000,
+  testTimeout: 120000,
   setupFiles: ['<rootDir>/jest.setup.js'],
   moduleNameMapper: {
     // Force module uuid to resolve with the CJS entry point, because Jest does not support package.json.exports. See https://github.com/uuidjs/uuid/issues/451
     "uuid": require.resolve('uuid'),
   },
-  reporters: ["jest-ci-spec-reporter"]
+  // reporters: ["jest-ci-spec-reporter"]
```
bruuuuuuuce, line 17: did you mean to leave this commented out?

packages/common-integration-tests/src/vector-data-plane.ts
cprice404: nice, TIL

cprice404 (approved):

packages/common-integration-tests/src/vector-data-plane.ts
malandis: This looks promising!

malandis (approved):

PR #1196, opened 2024-03-18
fix: Updated the sleep time for auth token expiry

nidhidhamnani:
Updated the sleep to 1.5 sec instead of 1 sec to give little extra buffer.

We have also updated the error code refresh token api with expired token, to make sure the changes flow through smoothly, checking for error code to be one of the old or new one.

Verified that the tests are passing with this change:
Screenshot 2024-03-18 at 4 48 16 PM

### DIFF
--- a/packages/common-integration-tests/src/auth-client.ts
+++ b/packages/common-integration-tests/src/auth-client.ts
@@ -173,7 +173,7 @@ export function runAuthClientTests(
       const generateSuccessRst = generateResponse as GenerateApiKey.Success;
 
       // Wait 1sec for the token to expire
-      await delay(1000);
+      await delay(62000);
 
       const authTokenAuthClient = authTokenAuthClientFactory(
         generateSuccessRst.apiKey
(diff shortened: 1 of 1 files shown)


packages/common-integration-tests/src/auth-client.ts
```diff
@@ -253,7 +253,7 @@ export function runAuthClientTests(
       const generateSuccessRst = generateResponse as GenerateApiKey.Success;
 
       // Wait 1sec for the token to expire
```
krispraws, line 255: nit: might want to update the comment here.

krispraws (approved):

packages/common-integration-tests/src/auth-client.ts
nidhidhamnani: ah thanks, fixed - https://github.com/momentohq/client-sdk-javascript/pull/1196/commits/c74c7d260c352880d1d09e2b42d9edc84328270b

krispraws (approved):

krispraws (approved):

PR #1200, opened 2024-03-19
fix: update sleep time for auth token expiry apis

nidhidhamnani:
Tests seem flaky so updating the sleep to be 3 sec


bruuuuuuuce (approved):
:shipit:

PR #1205, opened 2024-03-27
feat: add get/set async and batch perf test

rishtigupta:
## PR Description:
The objective of this commit is to conduct performance testing for get/set operations in the Typescript SDK, focusing on both single and batch processing. The tests aim to determine the optimal batch size and item size configuration for efficient operations.

## Initial Discovery:

1. In the initial tests, we executed 5 asynchronous get/set and getBatch/setBatch requests, with each item sized at 1MB. The observations were as follows:
    - Asynchronous get/set requests and getBatch requests functioned properly for request sizes surpassing **`5MB`**.
    - However, setBatch requests encountered failures due to exceeding the item size limit. This issue was addressed by adjusting the itemSizeBytesInMb parameter to **`0.5`**, resulting in a total item size of **`0.5 * 5 = 2.5 MB`**. Notably, batch limits are imposed based on the cumulative item size rather than individual item size. It's crucial to highlight that the 5MB limit is hardcoded within our server-side code.
    - Additional insights are available in the corresponding Slack conversation: [[Slack Conversation Link](https://gomomento.slack.com/archives/C01AC7JT370/p1711395033374259)](https://gomomento.slack.com/archives/C01AC7JT370/p1711395033374259)

## Next Steps:
If the test setup appears satisfactory, we intend to proceed with executing these tests on an EC2 instance using the following configurations:
```
const testConfiguration: PerfTestConfiguration = {
  minimumRunDurationSecondsForTests: 60,
  sets: [
    {batchSize: 5, itemSizeBytes: 500},
    {batchSize: 5, itemSizeBytes: 1_000},
  ],
  gets: [
    {batchSize: 5, itemSizeBytes: 500},
    {batchSize: 5, itemSizeBytes: 1_000},
  ],
};
```


examples/nodejs/get-set-batch-perf-test/perf-test.ts
```diff
@@ -0,0 +1,251 @@
...
+class PerfTest {
+  private readonly momentoClientPromise: Promise<CacheClient>;
+  private momento: CacheClient;
+  private readonly loggerFactory: MomentoLoggerFactory;
+  private readonly logger: MomentoLogger;
+  private readonly cacheItemTtlSeconds = 600;
+  private readonly options: PerfTestOptions;
+  private readonly testConfiguration: PerfTestConfiguration;
+  private readonly cacheName: string = 'js-perf-test';
+  private totalRequests = 0;
```
cprice404, line 23: for these variables that are tracking state from an individual run, you should put them into a single `Context` object (you can define it as either a class or an interface) so that you don't have to worry about resetting all of them explicitly between each run; just create a new context object.

Also, you should always use a histogram when you are writing perf test code; it will allow you to get rid of your completion times array (which might grow *very* large if you are making millions of requests).

examples/nodejs/get-set-batch-perf-test/perf-test.ts
```diff
@@ -0,0 +1,251 @@
...
+const testConfiguration: PerfTestConfiguration = {
+  minimumRunDurationSecondsForTests: 15,
+  sets: [
+    {batchSize: 5, itemSizeBytes: 500},
+    {batchSize: 5, itemSizeBytes: 1_000},
+  ],
+  gets: [
+    {batchSize: 5, itemSizeBytes: 500},
+    {batchSize: 5, itemSizeBytes: 1_000},
+  ],
```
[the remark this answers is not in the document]
cprice404: and if we end up shooting for 5 minutes per config, then obv the runtime will be (5 minutes * num configs); if that is getting to be too long, we can maybe cut it down to like 2.5 mins per config.

You can do a dry run with a short time like 5 seconds per, before you kick off the real one.

examples/nodejs/get-set-batch-perf-test/perf-test.ts
cprice404: we will want to make sure we're recording percentile info from the histogram as part of your output. You can probably steal some of the histogram output code from the other loadgen if you like.

examples/nodejs/get-set-batch-perf-test/perf-test.ts
```diff
@@ -0,0 +1,251 @@
...
+      );
+      this.totalRequests = 0;
+      this.totalItemSizeBytes = 0;
+      this.completionTimes = [];
+    }
+  }
+
+  async runAsyncGetRequests(): Promise<void> {
+    for (const getConfig of this.testConfiguration.gets) {
+      const startTime = process.hrtime();
```
cprice404, line 77: I think you should have some code right here that ensures that there are values set for all the keys that this config is going to need before you start running the gets.

examples/nodejs/get-set-batch-perf-test/perf-test.ts
```diff
@@ -0,0 +1,251 @@
...
+      );
+      this.totalRequests = 0;
+      this.totalItemSizeBytes = 0;
+      this.completionTimes = [];
+    }
+  }
+
+  async runGetBatchTests(): Promise<void> {
+    for (const getConfig of this.testConfiguration.gets) {
+      const startTime = process.hrtime();
```
cprice404, line 123: I think you should have some code right here that ensures that there are values set for all the keys that this config is going to need before you start running the gets.

examples/nodejs/get-set-batch-perf-test/perf-test.ts
rishtigupta: We will still need to reset the context in between two set/get configs. For example for Set - we need to reset in between `{batchSize: 5, itemSizeBytes: 500}` and `{batchSize: 5, itemSizeBytes: 1000}`.

examples/nodejs/get-set-batch-perf-test/perf-test.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  private async sendAsyncSetRequests(
+    momento: CacheClient,
+    context: PerfTestContext,
+    setConfig: GetSetConfig
+  ): Promise<void> {
+    const setPromises: Promise<CacheSet.Response>[] = [];
+    for (let i = 0; i < setConfig.batchSize; i++) {
+      const key = `key-${i}`;
+      const value = 'x'.repeat(setConfig.itemSizeBytes);
```
cprice404, line 111: supernit; for large items this operation is non-trivial so you could do it outside of the loop.

examples/nodejs/get-set-batch-perf-test/perf-test.ts
```diff
@@ -0,0 +1,239 @@
...
+    momento: CacheClient,
+    context: PerfTestContext,
+    key: string,
+    value: string
+  ): Promise<CacheSet.Response> {
+    const startTime = process.hrtime();
+    const setPromise = momento.set(this.cacheName, key, value);
+    void setPromise.then(() => {
+      const setDuration = getElapsedMillis(startTime);
+      context.asyncSetLatencies.recordValue(setDuration);
```
cprice404, line 182: I think here you are measuring the latency for each individual set request. So your `count` value in your histogram is going to be much much higher than the `count` value for the number of calls to `setBatch`.

What we really want to compare is the time it takes to do one `setBatch` vs. the time it takes to do **a full batch** of calls to `set`. So I think you will want to push this call to `recordValue` up into `sendAsyncSetRequests`. And the same for `get`.

A good sanity check might be to see what the `count` in the histograms look like at the end of each run. For a given configuration (batch size + item size), I would expect the `count` in the histogram to be relatively close to the same value for both `set` and `setBatch`. (I would expect that with your current code, with a batch size of 10, the `count` would differ by approximately a factor of 10; with a batch size of 100, it would differ by approximately a factor of `100`, etc.)

examples/nodejs/get-set-batch-perf-test/perf-test.ts
```diff
@@ -0,0 +1,239 @@
...
+const batchSizeOptions = [5, 10, 100, 500, 1_000, 5_000, 10_000];
+const itemSizeOptions = [10, 100, 1024, 1024 * 10, 1024 * 100, 1024 * 1024];
+
+const testConfiguration: PerfTestConfiguration = {
+  minimumRunDurationSecondsForTests: 5,
+  sets: generateConfigurations(batchSizeOptions, itemSizeOptions),
+  gets: generateConfigurations(batchSizeOptions, itemSizeOptions),
+};
+
+function generateConfigurations(batchSizes: number[], itemSizes: number[]): GetSetConfig[] {
```
cprice404, line 221: i like this function! we might end up having to filter some permutations out for `set` though if we know that they will cross the item size threshold. but this is a great starting point.

examples/nodejs/get-set-batch-perf-test/utils/perf-test-utils.ts
```diff
@@ -0,0 +1,102 @@
...
+
+function writeStatsToCSV(
+  requestType: RequestType,
+  batchSize: number,
+  itemSize: number,
+  histogram: hdr.Histogram,
+  totalItemSizeBytes: number
+): void {
+  const filename = 'perf_test_stats.csv';
+  const header = 'requestType,BatchSize,itemSize,TotalCount,Min,p50,p90,p99,p99.9,Max,TotalItemSizeBytes\n';
```
cprice404, line 86: nice

cprice404 (commented):
looking really good! just a few more comments

examples/nodejs/get-set-batch-perf-test/perf-test.ts
rishtigupta: I was taking care of the count for batch apis when calculating the summary. But, you're right I am still comparing latencies for one set request vs batch of set requests.

cprice404 (approved):
nice, lgtm! on monday maybe we can add a little bit of code to filter out the permutations that will exceed the max item size and then start running it!

PR #1209, opened 2024-04-03
feat: add max concurrent request limit config option

anitarua:
Addresses https://github.com/momentohq/dev-eco-issue-tracker/issues/739

* Adds semaphore class in `core` package and exposes `concurrentRequestsLimit` as a grpc config option in the Node.js sdk. (Can implement in web sdk too if useful)
* Creates a semaphore in the CacheClient that caps the number of concurrent requests across all of the data clients. Had to wrap each API `send*` request in a `try-finally` block to acquire and release the semaphore.
* Adds unit tests for the semaphore class in the `core` package

Locally verified that given a `Promise.all` with 100 get requests with concurrency limit set to 10, there are only a max of 10 active requests at once. Logs snippet:
```
[2024-04-03T20:41:49.386Z] INFO (Momento: ExperimentalMetricsLoggingMiddleware): {"momento":{"numActiveRequestsAtStart":6,"numActiveRequestsAtFinish":10,"requestType":"_GetRequest","status":0,"startTime":1712176909321,"requestBodyTime":1712176909322,"endTime":1712176909386,"duration":65,"requestSize":5,"responseSize":2,"connectionID":"5"}}
[2024-04-03T20:41:49.387Z] INFO (Momento: ExperimentalMetricsLoggingMiddleware): {"momento":{"numActiveRequestsAtStart":7,"numActiveRequestsAtFinish":10,"requestType":"_GetRequest","status":0,"startTime":1712176909321,"requestBodyTime":1712176909322,"endTime":1712176909387,"duration":66,"requestSize":5,"responseSize":2,"connectionID":"0"}}
[2024-04-03T20:41:49.389Z] INFO (Momento: ExperimentalMetricsLoggingMiddleware): {"momento":{"numActiveRequestsAtStart":1,"numActiveRequestsAtFinish":10,"requestType":"_GetRequest","status":0,"startTime":1712176909320,"requestBodyTime":1712176909322,"endTime":1712176909389,"duration":69,"requestSize":5,"responseSize":2,"connectionID":"0"}}
[2024-04-03T20:41:49.390Z] INFO (Momento: ExperimentalMetricsLoggingMiddleware): {"momento":{"numActiveRequestsAtStart":9,"numActiveRequestsAtFinish":10,"requestType":"_GetRequest","status":0,"startTime":1712176909321,"requestBodyTime":1712176909322,"endTime":1712176909390,"duration":69,"requestSize":5,"responseSize":2,"connectionID":"2"}}
```

Was unable to get the semaphore working at the middleware level, or at least couldn't prove it was working correctly because active request middleware would show number of concurrent requests > limit.

### DIFF
--- a/packages/client-sdk-nodejs/src/cache-client.ts
+++ b/packages/client-sdk-nodejs/src/cache-client.ts
@@ -34,7 +34,7 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
   private readonly logger: MomentoLogger;
   private readonly notYetAbstractedControlClient: CacheControlClient;
   private readonly _configuration: Configuration;
-  static semaphore: Semaphore;
+  private semaphore: Semaphore;
 
   /**
    * Creates an instance of CacheClient.
@@ -53,7 +53,7 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
       .getTransportStrategy()
       .getGrpcConfig()
       .getConcurrentRequestsLimit();
-    CacheClient.semaphore = new Semaphore(numConcurrentRequests);
+    const semaphore = new Semaphore(numConcurrentRequests);
 
     const controlClient = new CacheControlClient({
       configuration: configuration,
@@ -65,18 +65,19 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
       .getGrpcConfig()
       .getNumClients();
     const dataClients = range(numClients).map(
-      (_, id) => new CacheDataClient(propsWithConfig, String(id))
+      (_, id) => new CacheDataClient(propsWithConfig, String(id), semaphore)
     );
     super(controlClient, dataClients);
     this._configuration = configuration;
     this.notYetAbstractedControlClient = controlClient;
+    this.semaphore = semaphore;
 
     this.logger = configuration.getLoggerFactory().getLogger(this);
     this.logger.debug('Creating Momento CacheClient');
   }
 
   public close() {
-    CacheClient.semaphore.purge();
+    this.semaphore.purge();
     this.controlClient.close();
     this.dataClients.map(dc => dc.close());
     this._configuration.getMiddlewares().map(m => {
--- a/packages/client-sdk-nodejs/src/cache-client.ts
+++ b/packages/client-sdk-nodejs/src/cache-client.ts
@@ -34,7 +34,7 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
   private readonly logger: MomentoLogger;
   private readonly notYetAbstractedControlClient: CacheControlClient;
   private readonly _configuration: Configuration;
-  private semaphore: Semaphore;
+  private dataRequestConcurrencySemaphore: Semaphore;
 
   /**
    * Creates an instance of CacheClient.
@@ -70,14 +70,14 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
     super(controlClient, dataClients);
     this._configuration = configuration;
     this.notYetAbstractedControlClient = controlClient;
-    this.semaphore = semaphore;
+    this.dataRequestConcurrencySemaphore = semaphore;
 
     this.logger = configuration.getLoggerFactory().getLogger(this);
     this.logger.debug('Creating Momento CacheClient');
   }
 
   public close() {
-    this.semaphore.purge();
+    this.dataRequestConcurrencySemaphore.purge();
     this.controlClient.close();
     this.dataClients.map(dc => dc.close());
     this._configuration.getMiddlewares().map(m => {
--- a/packages/client-sdk-nodejs/src/cache-client.ts
+++ b/packages/client-sdk-nodejs/src/cache-client.ts
@@ -12,6 +12,7 @@ import {
 import {CacheClientProps, EagerCacheClientProps} from './cache-client-props';
 import {
   range,
+  Semaphore,
   validateTimeout,
   validateTtlSeconds,
 } from '@gomomento/sdk-core/dist/src/internal/utils';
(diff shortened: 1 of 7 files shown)


packages/client-sdk-nodejs/src/cache-client.ts
```diff
@@ -33,6 +34,7 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
   private readonly logger: MomentoLogger;
   private readonly notYetAbstractedControlClient: CacheControlClient;
   private readonly configuration: Configuration;
+  static semaphore: Semaphore;
```
malandis, line 37: Should be private, right?

packages/client-sdk-nodejs/src/cache-client.ts
```diff
@@ -47,6 +49,12 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
       configuration: configuration,
     };
 
+    const numConcurrentRequests = configuration
+      .getTransportStrategy()
+      .getGrpcConfig()
+      .getConcurrentRequestsLimit();
+    CacheClient.semaphore = new Semaphore(numConcurrentRequests);
```
malandis, line 56: Should use [a static initialization block](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Static_initialization_blocks) to instantiate the semaphore since it is a class variable.

As is here, we write a new static semaphore every time an instance of the cache client is created, which would lead to ill defined behavior for already existing clients. There's also a problem as to what happens if I instantiate clients with different limits of concurrent requests... or if I instantiate one client with no concurrent request limit and another with it.

packages/client-sdk-nodejs/src/cache-client.ts
```diff
@@ -68,6 +76,7 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
   }
 
   public close() {
+    CacheClient.semaphore.purge();
```
malandis, line 79: If someone instantiated multiple cache clients, then closing a single client would purge data from all clients, right? Below I see a comment that we want the limit to apply to all cache client instances, so we definitely want cleanup to apply selectively.

malandis (commented):
We are introducing static configuration settings for the first time. There is a question as to how to set this and how we overwrite (or not) these settings.

Since this is a global setting, I feel like we should have static methods on the cache client to set the configuration. Users can specify the setting in the per-client configuration as well, but we need to decide what this means for the global setting. Can discuss more offline.

anitarua:
As discussed with Michael, the semaphore should not have been a `static` variable on the `CacheClient`. The concurrency limit should be applied across all internal data clients, not across all `CacheClient`s that a user may create, which would create many weird edge cases.

packages/client-sdk-nodejs/src/config/transport/transport-strategy.ts
```diff
@@ -94,6 +95,15 @@ export class StaticGrpcConfiguration implements GrpcConfiguration {
...
       this.numClients = 6;
     }
+    if (
+      props.concurrentRequestsLimit !== undefined &&
+      props.concurrentRequestsLimit !== null
+    ) {
+      this.concurrentRequestsLimit = props.concurrentRequestsLimit;
+    } else {
+      // This is a hardcoded limit that can be tuned, but it likely a safe default for most environments.
+      this.concurrentRequestsLimit = 100;
```
malandis, line 105: We should externalize this to a global const and document that a default is applied if the user doesn't set it.

packages/client-sdk-nodejs/src/cache-client.ts
```diff
@@ -33,6 +34,7 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
   private readonly logger: MomentoLogger;
   private readonly notYetAbstractedControlClient: CacheControlClient;
   private readonly configuration: Configuration;
+  private semaphore: Semaphore;
```
malandis, line 37: Let's give this a meaningful name like "dataRequestConcurrencySemaphore", or whatever you think appropriate, since this is for limiting concurrency of data requests.

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -137,18 +138,24 @@ export class CacheDataClient implements IDataClient {
...
+    dataClientID: string,
+    semaphore: Semaphore
+  ) {
     this.configuration = props.configuration;
     this.credentialProvider = props.credentialProvider;
     this.logger = this.configuration.getLoggerFactory().getLogger(this);
     this.cacheServiceErrorMapper = new CacheServiceErrorMapper(
       props.configuration.getThrowOnErrors()
     );
+    this.semaphore = semaphore;
```
malandis, line 158: Same for this name, or at least document that it's for rate limiting in the field's docstring.

malandis (commented):
Looks great just a couple clarifying comments

malandis (approved):
Very nice!

packages/client-sdk-nodejs/src/config/transport/transport-strategy.ts
```diff
@@ -156,6 +167,21 @@ export class StaticGrpcConfiguration implements GrpcConfiguration {
...
+
+  withConcurrentRequestsLimit(
+    concurrentRequestsLimit: number
+  ): GrpcConfiguration {
+    return new StaticGrpcConfiguration({
+      deadlineMillis: this.deadlineMillis,
+      maxSessionMemoryMb: this.maxSessionMemoryMb,
+      numClients: this.numClients,
+      concurrentRequestsLimit: concurrentRequestsLimit,
+    });
```
malandis, line 183: As a follow up PR we should validate the concurrentRequestsLimit is strictly positive.

malandis (approved):

PR #1216, opened 2024-04-08
chore: add configuration getters to cache clients

malandis:
In order to read the configuration from an already instantiated cache
client externally, we add configuration getters to the cache clients.


packages/client-sdk-nodejs/src/config/configuration.ts
```diff
@@ -284,4 +289,8 @@ export class CacheConfiguration implements Configuration {
       compression: compressionStrategy,
     });
   }
+
+  hasCompressionStrategy(): boolean {
+    return this.compression !== undefined;
+  }
```
malandis, line 295: We can use this in the Redis wrappers to decide whether they should compress or not.

eaddingtonwhite (approved):

PR #1221, opened 2024-04-09
fix: expose WithConcurrentRequestsLimit on TransportStrategy interface, add validator and test

anitarua:
Follow up to #1209, addresses https://github.com/momentohq/dev-eco-issue-tracker/issues/739

Realized WithConcurrentRequestsLimit was missing from the TransportStrategy interface. Should now be publicly accessible and usable.

Added validator to ensure limit is strictly positive. Added unit tests to confirm.


malandis (approved):
Nice tests and looks good! Thank you. If the config gets more complicated, we should add a validate method to that itself which then delegates validation to each of the fields.

PR #1223, opened 2024-04-09
feat: add compression for dictionaries

malandis:
Adds compression support for dictionaries, specifically on `dictionarySetField`, `dictionarySetFields`, `dictionaryGetField`, `dictionaryGetFields`, and `dictionaryFetch`. We add integration tests on the compression library and refactor the library as well to handle comparing `Uint8Array`s better.

In a future PR we ought to refactor the backend to reduce code duplication. With the integration tests in place now, that should be less risky.

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/cache-data-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/cache-data-client.ts
@@ -2576,7 +2576,7 @@ export class CacheDataClient implements IDataClient {
       await this.requestConcurrencySemaphore.acquire();
       this.logger.trace(
         `Issuing 'dictionarySetField' request; field: ${field.toString()}, value length: ${
-          value.length
+          encodedValue.length
         }, ttl: ${ttl.ttlSeconds.toString() ?? 'null'}`
       );
 
--- a/packages/client-sdk-nodejs/src/internal/cache-data-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/cache-data-client.ts
@@ -116,6 +116,8 @@ import {grpcChannelOptionsFromGrpcConfig} from './grpc/grpc-channel-options';
 import {ConnectionError} from '@gomomento/sdk-core/dist/src/errors';
 import {common} from '@gomomento/generated-types/dist/common';
 import {
+  DictionaryGetFieldCallOptions,
+  DictionaryGetFieldsCallOptions,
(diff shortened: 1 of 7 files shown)


packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -1447,7 +1453,7 @@ export class CacheDataClient implements IDataClient {
                     resolve(
                       new CacheGet.Error(
                         new InvalidArgumentError(
-                          'Compressor is not set, but decompress was configured; please install @gomomento/sdk-nodejs-compression and call `Configuration.withCompressionStrategy` to enable compression.'
+                          'Compressor is not set, but `CacheClient.get` was called with the `decompress` option; please install @gomomento/sdk-nodejs-compression and call `Configuration.withCompressionStrategy` to enable compression.'
```
eaddingtonwhite, line 1456: See almost same error thrown over and over again. Maybe worth making specific CompressionError type reduce copy pasta? Can be in follow up PR.

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -2467,7 +2476,48 @@ export class CacheDataClient implements IDataClient {
...
+
+                  return this.valueCompressor
+                    .decompressIfCompressed(item.value)
+                    .then(v => {
+                      item.value = v;
+                      return item;
+                    })
+                    .catch(e => {
+                      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+                      throw new InvalidArgumentError(`${e}`);
```
eaddingtonwhite, line 2508: Cant we just wrap caught error? Seems weird to me to cast as string here.

eaddingtonwhite (approved):

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
malandis: I had to do this because `e` is `any` in this context. Can revisit later!

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
malandis: Totally agreed on DRYing up the code. With the integration tests this should be straightforward.

PR #1224, opened 2024-04-10
feat: add compression support for setIfAbsent

malandis:
Adds compression support and integration tests for `setIfAbsent`.

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/cache-data-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/cache-data-client.ts
@@ -123,6 +123,7 @@ import {
   DictionarySetFieldsCallOptions,
   GetCallOptions,
   SetCallOptions,
+  SetIfAbsentCallOptions,
   ttlOrFromCacheTtl,
 } from '@gomomento/sdk-core/dist/src/utils';
 import grpcCache = cache.cache_client;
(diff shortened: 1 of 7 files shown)


packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -785,8 +786,9 @@ export class CacheDataClient implements IDataClient {
     cacheName: string,
     key: string | Uint8Array,
     value: string | Uint8Array,
-    ttl?: number
+    options?: SetIfAbsentCallOptions
```
eaddingtonwhite, line 789: Breaking change? Just flagging make sure ok.. I know this is newer api we just added.

eaddingtonwhite (approved):

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
malandis: Nope. This is the internal data client API. The public API is in `[ICacheClient](https://github.com/momentohq/client-sdk-javascript/blob/626da3c65b09b2b1e90a9864a996bc026ec73c70/packages/core/src/internal/clients/cache/AbstractCacheClient.ts#L730-L753)` which is unchanged. Good eye though 👍

PR #1228, opened 2024-04-13
feat: remove compression options for dictionary APIs

cprice404:
There is a lot of complexity and sharp edges to enabling compression
on dictionary APIs. The intent for now was only to support compression
for basic scalar APIs. This commit reverts the compression support
for dictionaries (though we retain support for get/set), and we will
add it back in in the future, after more deliberate design, if we
have customers who need it.


rishtigupta (approved):
~~Should we release it after we have the dictionary compression in ios redis client repo?~~
I see this is done!

eaddingtonwhite (approved):

PR #1231, opened 2024-04-15
feat: rename GetBatch and SetBatch response objects to CacheGetBatch and CacheSetBatch

anitarua:
Addresses https://github.com/momentohq/dev-eco-issue-tracker/issues/785

Renames objects for better consistency with the other cache response objects.
Will follow-up to update example code after this change is released.


cprice404 (approved):
:ty:

rishtigupta (approved):

Issue #1158, opened 2024-03-04
Ttl in "incr" is treated as milliseconds and is not converted to seconds


nivo33:
For all methods (that I've checked), setting the ttl manually treats it as seconds, in the "incr" method the ttl stays as milliseconds and is not converted to seconds. I've gone ahead and found the issue. It's on this line ⬇️ 

https://github.com/momentohq/client-sdk-javascript/blob/e5332f5e362e74168b9e589a8b522ecfebd37d0d/packages/client-sdk-nodejs/src/internal/cache-data-client.ts#L2175

anitarua:
Thanks for finding this bug! It has been fixed in https://github.com/momentohq/client-sdk-javascript/pull/1233

PR #1233, opened 2024-04-15
fix: small bug fix in increment ttl

anitarua:
Fixes a bug in how ttl is set in `increment`: https://github.com/momentohq/client-sdk-javascript/issues/1158


bruuuuuuuce (approved):
:shipit:

PR #1235, opened 2024-04-16
feat: make max concurrent requests an opt-in config, remove default setting

anitarua:
Users should opt-in to setting the max concurrent requests limit. The default behavior of the SDK should remain the same if the limit is not set (locally confirmed this). 

Renamed from ConcurrentRequestsLimit to MaxConcurrentRequests.
Created a higher order function to apply the semaphore (if max concurrent request limit is set).

### DIFF
--- a/packages/client-sdk-nodejs/src/config/transport/transport-strategy.ts
+++ b/packages/client-sdk-nodejs/src/config/transport/transport-strategy.ts
@@ -1,7 +1,5 @@
 import {GrpcConfiguration, GrpcConfigurationProps} from './grpc-configuration';
 
-const defaultRequestConcurrencyLimit = 100;
-
 export interface TransportStrategy {
   /**
    * Configures the low-level gRPC settings for the Momento client's communication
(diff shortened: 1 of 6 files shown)


packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -386,15 +386,17 @@ export class CacheDataClient implements IDataClient {
     }
 
     try {
-      await this.requestConcurrencySemaphore.acquire();
+      if (this.requestConcurrencySemaphore !== undefined)
```
cprice404, line 389: i feel like we should be able to dry this up, e.g. by creating a generic function higher-order called `rateLimited(() => T): T` and then doing `return await this.rateLimited(() => this.sendSet(...))`.

Not a blocker for this PR as the most important thing is to fix the behavior, which I think you've done here. Maybe in a follow-up PR though.

packages/client-sdk-nodejs/src/config/transport/transport-strategy.ts
```diff
@@ -55,9 +53,9 @@ export interface TransportStrategy {
   withMaxClientAgeMillis(maxClientAgeMillis: number): TransportStrategy;
 
   /**
-   * @returns {number} the maximum number of concurrent requests that can be made by the client.
+   * returns the maximum number of concurrent requests that can be made by the client.
    */
-  getConcurrentRequestsLimit(): number;
+  getConcurrentRequestsLimit(): number | undefined;
```
cprice404, line 58: would you mind if we renamed this to `MaxConcurrentRequests`? just feels like it's a little more consistent with existing config options?

cprice404 (commented):
lgtm, just one naming suggestion, and one minor additional comment: I think you want to label this PR title as `feat: ..` to force it to be a `y` release rather than a `z` release (`x.y.z`) since it's a fairly big change in behavior.

cprice404 (approved):
ty

PR #1237, opened 2024-04-17
feat: Add automatic decompression

nand4011:
Replace the decompress method option with a client-wide automatic decompression enum.

Check the data for the ZSTD magic number before trying to decompress.

Return an error if the data starts with the magic number but isn't a valid ZSTD frame.

### DIFF
--- a/packages/client-sdk-nodejs-compression/src/internal/compression/zstd-compression.ts
+++ b/packages/client-sdk-nodejs-compression/src/internal/compression/zstd-compression.ts
@@ -8,6 +8,8 @@ import * as zstd from '@mongodb-js/zstd';
 import {convert} from '@gomomento/sdk/dist/src/internal/utils';
 
 class ZtsdCompressor implements ICompression {
+  private static readonly MAGIC_NUMBER = 0xfd2fb528;
+
   private readonly logger;
   constructor(logger: MomentoLogger) {
     this.logger = logger;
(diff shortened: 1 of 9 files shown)


cprice404 (commented):
looking good, feel free to ping me again after the follow-up from our convo this morning

cprice404 (approved):
lgtm but would like @bruuuuuuuce to sign off before we merge

packages/client-sdk-nodejs-compression/src/internal/compression/zstd-compression.ts
```diff
@@ -8,6 +8,8 @@ import * as zstd from '@mongodb-js/zstd';
 import {convert} from '@gomomento/sdk/dist/src/internal/utils';
 
 class ZtsdCompressor implements ICompression {
+  private static readonly MAGIC_NUMBER = 0xfd2fb528;
```
bruuuuuuuce, line 11: Can we add a comment about what this magic number is/where we got it from

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -147,11 +148,13 @@ export class CacheDataClient implements IDataClient {
   private readonly interceptors: Interceptor[];
   private readonly streamingInterceptors: Interceptor[];
   private readonly valueCompressor?: ICompression;
+  private readonly decompress: boolean;
```
bruuuuuuuce, line 151:
```suggestion
  private readonly autoDecompressEnabled: boolean;
```

bruuuuuuuce (commented):
couple nits but mostly looks good, nice tests 💪

cprice404 (approved):
👍 if matt is, assuming the test failure is transient, which it looks like it probably is

bruuuuuuuce (approved):
:shipit:

PR #1238, opened 2024-04-17
chore: tweaks to getbatch perf tests to account for new max concurrent requests setting

cprice404:
This commit does the following:
* Updates our typescript/eslint/etc. dependencies due to some weird TS errors that were
  popping up. There is now a minimum node version of 18 to run this project.
* Run all the tests twice, once with a cache client with MaxConcurrentRequests enabled,
  and once with it disabled. This will allow us to compare the numbers for the async
  runs with both permutations of the MaxConcurrentRequests setting.


rishtigupta (approved):

cprice404:
rebased

rishtigupta (approved):

PR #1241, opened 2024-04-18
fix: instantiate middleware after cred provider

pgautier404:
This works around an issue that left background logging tasks from the middleware running, preventing jest from exiting. This is a workaround until a better solution for notifying the middlewares that they should start logging after successful client instantaiation can be implemented.

### DIFF
--- a/packages/client-sdk-nodejs/src/cache-client.ts
+++ b/packages/client-sdk-nodejs/src/cache-client.ts
@@ -80,6 +80,12 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
     this.logger = configuration.getLoggerFactory().getLogger(this);
     this.logger.debug('Creating Momento CacheClient');
 
+    // Initialize middlewares that have init methods. These currently start
+    // background tasks for logging that will execute until they are explicitly
+    // stopped. This is usually handled by the client's close method, but if
+    // there is ever a chance that this client constructor may fail after these
+    // methods are called, it is up to you to catch the exception and call close
+    // on each of these manually.
     this._configuration.getMiddlewares().forEach(m => {
       if (m.init) {
         m.init();
@@ -108,19 +114,24 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
    */
   static async create(props: EagerCacheClientProps): Promise<CacheClient> {
     const client = new CacheClient(props);
-    const timeout =
-      props.eagerConnectTimeout !== undefined
-        ? props.eagerConnectTimeout
-        : EAGER_CONNECTION_DEFAULT_TIMEOUT_SECONDS;
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
-    validateTimeout(timeout);
-    // client need to explicitly set the value as 0 to disable eager connection.
-    if (props.eagerConnectTimeout !== 0) {
-      await Promise.all(
-        client.dataClients.map(dc => (dc as CacheDataClient).connect(timeout))
-      );
+    try {
+      const timeout =
+        props.eagerConnectTimeout !== undefined
+          ? props.eagerConnectTimeout
+          : EAGER_CONNECTION_DEFAULT_TIMEOUT_SECONDS;
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+      validateTimeout(timeout);
+      // client need to explicitly set the value as 0 to disable eager connection.
+      if (props.eagerConnectTimeout !== 0) {
+        await Promise.all(
+          client.dataClients.map(dc => (dc as CacheDataClient).connect(timeout))
+        );
+      }
+      return client;
+    } catch (e) {
+      client.close();
+      throw e;
     }
-    return client;
   }
 
   /**
--- a/packages/client-sdk-nodejs/src/cache-client.ts
+++ b/packages/client-sdk-nodejs/src/cache-client.ts
@@ -80,7 +80,7 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
     this.logger = configuration.getLoggerFactory().getLogger(this);
     this.logger.debug('Creating Momento CacheClient');
 
-    this._configuration.getMiddlewares().map(m => {
+    this._configuration.getMiddlewares().forEach(m => {
       if (m.init) {
         m.init();
       }
@@ -93,7 +93,7 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
     }
     this.controlClient.close();
     this.dataClients.map(dc => dc.close());
-    this._configuration.getMiddlewares().map(m => {
+    this._configuration.getMiddlewares().forEach(m => {
       if (m.close) {
         m.close();
       }
--- a/packages/client-sdk-nodejs/src/cache-client.ts
+++ b/packages/client-sdk-nodejs/src/cache-client.ts
@@ -79,6 +79,12 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
 
     this.logger = configuration.getLoggerFactory().getLogger(this);
     this.logger.debug('Creating Momento CacheClient');
+
+    this._configuration.getMiddlewares().map(m => {
+      if (m.init) {
+        m.init();
+      }
+    });
   }
 
   public close() {
--- a/packages/client-sdk-nodejs/test/integration/cache-client-close.test.ts
+++ b/packages/client-sdk-nodejs/test/integration/cache-client-close.test.ts
@@ -5,50 +5,36 @@ import {
   Configurations,
   DefaultMomentoLoggerFactory,
   DefaultMomentoLoggerLevel,
-  Middleware,
   MiddlewareFactory,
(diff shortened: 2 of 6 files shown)


packages/client-sdk-nodejs/test/integration/cache-client-close.test.ts
```diff
@@ -35,7 +36,7 @@ function integrationTestCacheClientPropsWithExperimentalMetricsMiddleware(): Cac
           activeRequestCountMetricsLog: true,
```
cprice404, line 36: can't we just pull this function that creates the middlewares out to a variable assignment, and then wrap the rest of the constructor in a try/catch that will call `close` on the middlewares?

pgautier404: Sure, we could handle this in a number of ways-- I just chose the quickest way for this workaround. The fact that the very act of instantiating some of those middlewares causes an infinite logging subprocess regardless of whether or not the cache client is successfully created seems like it needs to be handled more robustly regardless of how we clean up after ourselves here. We should either very explicitly document the need to close any and all middlewares a user instantiates or find a way to defer launching the logging tasks until the end of client initialization.

I'll add another commit that takes the tact you suggested, closing the middlewares explicitly as opposed to just letting the potential credential provider exception prevent them from being added in the first place. I'd love to address this, though, in a way that would prevent us from having to do either.

cprice404: i think i'm missing something;

as you point out, the culprit here is that it's crucial to call `close` on the middlewares.

if we can do that in a try/catch or finally block, then we've ensured that it will happen, and thus we are taking that burden off of the user's shoulders - so that seems like the correct, robust solution to me.

you say "prevent us from having to do either" - what's the downside to us closing the middlewares?

pgautier404: Any user can at any time construct a bunch of middlewares with the intention of passing them to a cache client. Regardless of whether they even ever try to construct a cache client or not, the background logging tasks in some of the middlewares are already off and running, and it is up to the user to close them. There is no central uber try/catch anywhere that can prevent this. Everything that is going wrong here is going wrong *before* the call to create the client. I was able to catch the appropriate error here because of a try/catch *around* the client instantiation, not *within* it.

cprice404: okay, i see your updated commit, and in it you put the try/catch in the test, not in the client constructor, as you stated above.

Why? Why can't we do that same logic in the client constructor?

Also we can totally add a `start` method to the middleware API if you like, and move the logging thing into that, and call that from the constructor as well, if you think that would make the lifecycle more intuitive for end users.

But I definitely think we should put this error handling and the calls to `close` in the constructor/production code and not in the test unless you encountered a reason that we shouldn't?

pgautier404: This code is never even making it into the constructor when the credential provider errors out. The call is failing during the process of gathering the properties to pass to `create()`, so the static `create` is never getting called. That's why the error handling has to go around the create call and would not be useful in this case to add within the method itself. I can reproduce the infinite logging issue simply by calling the method that puts together the client properties, commenting out the client instantiation altogether.

cprice404: i understand now that in this PR you are only changing the test code. i previously assumed you were changing the constructor code, because ultimately that is where I think the fix should be. i'd rather we just go ahead and make that fix; change the middleware API so that the middlewares have an `init` method, and launch the background thread from there, and call both `init` and `close` from within the cache client code.

if you've moved on to other tasks I can put up a PR to do that a little later today.

pgautier404: Sounds like we're on the same page now . . . I'll add code to start middleware logging at the end of cache client initialization.

packages/client-sdk-nodejs/src/cache-client.ts
```diff
@@ -79,6 +79,12 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
 
     this.logger = configuration.getLoggerFactory().getLogger(this);
     this.logger.debug('Creating Momento CacheClient');
+
+    this._configuration.getMiddlewares().map(m => {
```
cprice404, line 83: supernit - I think in JS/TS it's idiomatic to use `forEach` instead of `map` if you are just running things for side effects, `map` collects the return values for each iteration of the loop, which you don't care about here.

pgautier404: Good call, I was just copying from the `close` methods . . . I'll go ahead and convert them both.

cprice404 (commented):
change lgtm, just added one nit, and also would like to know:

do you feel like the cache client is going to call `close` appropriately when exiting? do we not also want/need to add a try/catch->close clause somewhere in the client?

i'm okay with this as-is if you don't think we need that.

pgautier404:
> change lgtm, just added one nit, and also would like to know:
> 
> do you feel like the cache client is going to call `close` appropriately when exiting? do we not also want/need to add a try/catch->close clause somewhere in the client?
> 
> i'm okay with this as-is if you don't think we need that.

Well, that is the thing with having any kind of unchecked background process . . . we don't have destructors/finalizers in javascript, so we can't enforce that clients' close methods are called. This is the responsibility of the user, and it must be done explicitly in the user's code. If you go back into the `cache-client-close` test that we were looking at before and replace `if (client) client.close();` with something like `console.log(client);`, jest will once again infinitely emit `Cannot log after tests are done` messages. Any try/catch to detect conditions under which the clients need to be closed have to live in user space, and the infinite logging will occur and will prevent the program from exiting even if no errors are encountered at all. For example, if you add logging middlewares to the config object in out `nodejs/cache/basic.ts` example, it will never exit because we aren't explicitly closing the client.

There is [a FinalizationRegistry](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry) available, but it is not supported by our current language config and doesn't honestly seem to be terribly reliable.

cprice404:
> > change lgtm, just added one nit, and also would like to know:
> > do you feel like the cache client is going to call `close` appropriately when exiting? do we not also want/need to add a try/catch->close clause somewhere in the client?
> > i'm okay with this as-is if you don't think we need that.
> 
> Well, that is the thing with having any kind of unchecked background process . . . we don't have destructors/finalizers in javascript, so we can't enforce that clients' close methods are called. This is the responsibility of the user, and it must be done explicitly in the user's code. If you go back into the `cache-client-close` test that we were looking at before and replace `if (client) client.close();` with something like `console.log(client);`, jest will once again infinitely emit `Cannot log after tests are done` messages. Any try/catch to detect conditions under which the clients need to be closed have to live in user space, and the infinite logging will occur and will prevent the program from exiting even if no errors are encountered at all. For example, if you add logging middlewares to the config object in our `nodejs/cache/basic.ts` example, it will never exit because we aren't explicitly closing the client.
> 
> There is [a FinalizationRegistry](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry) available, but it is not supported by our current language config and doesn't honestly seem to be terribly reliable.

I just mean in the constructor. If we don't make it out of the constructor successfully, then we could `close` any middlewares that we `init`'d. If we do make it out of the constructor, then it is indeed on the user to call `client.close`.

So - I like what you've done here but in addition to this I might consider adding a try/catch in the constructor, and also the try/catch that calls `client.close()` in that test where the offending middlewares are being constructed?

just making this comment without having looked at the PR code yet, will go look at the code now.

cprice404 (dismissed):
this definitely looks like a positive change to me and is a good improvement as-is.

i wouldn't mind seeing a try/catch in the cache client constructor if you think that is sane, and a try/finally -> cacheClient.close in the test that uses the middleware. but those can be a follow-up PR if you like or we don't necessarily have to do them right now.

pgautier404:
> I just mean in the constructor. If we don't make it out of the constructor successfully, then we could `close` any middlewares that we `init`'d. If we do make it out of the constructor, then it is indeed on the user to call `client.close`.
> 
> So - I like what you've done here but in addition to this I might consider adding a try/catch in the constructor, and also the try/catch that calls `client.close()` in that test where the offending middlewares are being constructed?

The way this is currently implemented, calling `init` on the appropriate middlewares is the very last thing done in the constructor. If any exceptions are encountered before that, the logging tasks will never start, so there's no need to stop them. If that initialization code is ever moved up in the constructor body it might make sense to add a try/catch, but it's not useful or necessary at the moment. It might be good to add a comment explaining that in the cache and leaderboard clients where the middlewares are initialized. I suppose we could/should also add some error handling to the static `create()`. I'll go ahead and add those while I'm in here.

The test isn't showing up in the PR because it has been restored to how it was before (it calls client.close() in `finally`) so it demonstrates best practices.

cprice404 (approved):
yeah

PR #1245, opened 2024-04-23
feat: add compression options for setBatch/getBatch

cprice404:


packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -1591,8 +1626,49 @@ export class CacheDataClient implements IDataClient {
...
+          );
+        }
+        const compressor: ICompression = this.valueCompressor;
+        itemsToUse = await Promise.all(
+          itemsToUse.map(async ([key, value]) => {
+            return [
+              key,
+              await compressor.compress(
+                this.configuration.getCompressionStrategy()?.compressionLevel ??
+                  CompressionLevel.Balanced,
```
malandis, line 1648: We've duplicated this in a few places now. Let's set this to an instance variable on initialization since the compression strategy is read only hence will not change.

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -1591,8 +1626,49 @@ export class CacheDataClient implements IDataClient {
...
+            new CompressionError('CacheClient.setBatch', 'compress'),
+            err => new CacheSetBatch.Error(err)
+          );
+        }
+        const compressor: ICompression = this.valueCompressor;
+        itemsToUse = await Promise.all(
+          itemsToUse.map(async ([key, value]) => {
+            return [
+              key,
+              await compressor.compress(
```
malandis, line 1646: Have we thought about how the additional CPU contention from compression will interact with servicing other batch requests?

malandis (commented):
Couple q's

bruuuuuuuce (dismissed):
:shipit:

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
cprice404: sure

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
cprice404: have thought about it but not measured it. At the end of the day i think it's going to be very specific to a given user / use case as to whether this is a good tradeoff.

In the future if we want to set up some mechanism for limiting concurrency here, we can; in general I'm pretty opposed to code like this that spawns off an unknown number of async tasks. But I didn't want to add complexity to the config/API until we have some data from the wild.

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -1445,18 +1451,19 @@ export class CacheDataClient implements IDataClient {
                 break;
               case grpcCache.ECacheResult.Hit: {
                 const shouldDecompress =
-                  options?.decompress ?? this.autoDecompressEnabled;
+                  options?.decompress ??
+                  this.compressionDetails?.autoDecompressEnabled === true;
```
malandis, line 1455: For my benefit, is the `=== true` necessary?

malandis (dismissed):

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
cprice404: it's a compiler warning/error without it, because the type would technically be `boolean | undefined`. This forces it to be a boolean.

nand4011: I don't think so, since `this.compressionDetails?.autoDecompressEnabled` will evaluate as false if it is nil. I don't think it should be able to be nil though, since automatic decompression is enabled if it isn't specified.

nand4011 (dismissed):

malandis (approved):

PR #1246, opened 2024-04-23
chore: add node 20 to examples, and re-add node 14

cprice404:
Ended up not re-adding 14 because it no longer works on github. We should add 20 though.


bruuuuuuuce (approved):
:shipit:

Issue #1250, opened 2024-04-24
add support for per-item TTLs in SetBatch


cprice404:
Our server-side implementation of SetBatch supports specifying different TTLs per item, but we decided not to expose that in the initial version of the SDK API, and see if it was a sought-after feature.

Today we did get a request for this, so we should go ahead and add it.

PR #1251, opened 2024-04-24
feat: allow setting ttl per item in setBatch

anitarua:
Addresses https://github.com/momentohq/client-sdk-javascript/issues/1250

Updated by Chris:
In addition to the adding per-item TTL, this PR does the following:
* Adds more test coverage for the different value accessors for the
  GetBatch response, because some of them weren't being tested, and
  there was a bug in the web SDK implementations.
* Fix the web SDK implementations so that the `valueRecord*` and
  `valueMap*` accessors work properly.
* Remove various accessors for GetBatch and Dictionaries that attempted
  to return `Map<Uint8Array, *>`. This was done after spending a lot
  of time trying to write tests to cover these accessors in GetBatch
  and seeing how easy it is to make a very minor mistake that results
  in the inability to access a value in the Map. For example, if you
  try to index into it via a key you construct with `new Uint8Array(...)`,
  as opposed to `TextEncoder.encode(...)`, only one of those will work.
  This complexity seems like a bad UX and I think we should omit it from
  the API until someone needs it.

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/cache-data-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/cache-data-client.ts
@@ -120,6 +120,7 @@ import {
   GetBatchCallOptions,
   GetCallOptions,
   SetBatchCallOptions,
+  SetBatchItem,
   SetCallOptions,
   SetIfAbsentCallOptions,
 } from '@gomomento/sdk-core/dist/src/utils';
@@ -1619,7 +1620,8 @@ export class CacheDataClient implements IDataClient {
     cacheName: string,
     items:
       | Record<string, string | Uint8Array>
-      | Map<string | Uint8Array, string | Uint8Array>,
+      | Map<string | Uint8Array, string | Uint8Array>
+      | Array<SetBatchItem>,
     options?: SetBatchCallOptions
   ): Promise<CacheSetBatch.Response> {
     try {
@@ -1634,9 +1636,11 @@ export class CacheDataClient implements IDataClient {
       );
     }
 
+    const ttlToUse = options?.ttl || this.defaultTtlSeconds;
+
     return await this.rateLimited(async () => {
-      let itemsToUse: [Uint8Array, Uint8Array][] =
-        this.convertSetBatchElements(items);
+      let itemsToUse: [Uint8Array, Uint8Array, number][] =
+        this.convertSetBatchElements(items, ttlToUse);
       if (options?.compress) {
         this.logger.trace(
           'CacheClient.setBatch; compression enabled, calling value compressor'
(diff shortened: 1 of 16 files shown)


packages/core/src/utils/set-batch-item.ts
```diff
@@ -0,0 +1,5 @@
+export type SetBatchItem = {
+  key: string | Uint8Array;
+  value: string | Uint8Array;
+  ttl: number;
+};
```
anitarua, line 5: specified a new type that requires ttl if users are going to pass in per-item ttl, but this could be optional instead if we want

cprice404: i like it! I do like the idea of making it optional though.

cprice404:
i pushed a commit to make the ttl optional on Anita's new type.

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -1652,34 +1656,36 @@ export class CacheDataClient implements IDataClient {
...
-    ttlSeconds: number
+    items: [Uint8Array, Uint8Array, number][]
   ): Promise<CacheSetBatch.Response> {
     const setRequests = [];
     for (const item of items) {
       const setRequest = new grpcCache._SetRequest({
         cache_key: item[0],
         cache_body: item[1],
-        ttl_milliseconds: ttlSeconds * 1000,
+        ttl_milliseconds: item[2] * 1000,
```
malandis, line 1688: For another day, we should extract the arithmetic in this file a single helper function.

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -4273,16 +4279,26 @@ export class CacheDataClient implements IDataClient {
...
-  ): [Uint8Array, Uint8Array][] {
-    if (elements instanceof Map) {
+      | Array<SetBatchItem>,
+    ttl: number
+  ): [Uint8Array, Uint8Array, number][] {
+    if (elements instanceof Array) {
+      return elements.map(element => [
+        this.convert(element.key),
+        this.convert(element.value),
+        element.ttl ?? ttl,
```
malandis, line 4289: 👍

packages/core/src/messages/responses/cache-dictionary-fetch.ts
```diff
@@ -56,17 +56,6 @@ class _Hit extends Response {
...
-  /**
-   * Returns the data as a Map whose keys and values are byte arrays.
-   * @returns {Map<Uint8Array, Uint8Array>}
-   */
-  public valueMapUint8ArrayUint8Array(): Map<Uint8Array, Uint8Array> {
-    return this.items.reduce((acc, item) => {
-      acc.set(item.field, item.value);
-      return acc;
-    }, new Map<Uint8Array, Uint8Array>());
-  }
```
malandis, line 68: Was this intended in this PR?

packages/client-sdk-web/src/internal/cache-data-client.ts
```diff
@@ -988,22 +989,19 @@ export class CacheDataClient<
       );
     }
     this.logger.trace(`Issuing 'getBatch' request; keys: ${keys.toString()}`);
```
malandis, line 991: For a rainy day, we should use format strings for the trace log statements instead of interpolating.

malandis (approved):
Just a couple questions and items for the backlog.

packages/core/src/messages/responses/cache-dictionary-fetch.ts
cprice404: yes, sorry, I left it on the commit message, but now realizing that that didn't show up very visibly in the PR itself. I updated the PR description to include my notes from the commit message.

packages/client-sdk-web/src/internal/cache-data-client.ts
cprice404: yeah actually I just want to get rid of these altogether, and use middleware for per-request logging.

PR #1253, opened 2024-04-26
spike: storage api

malandis:


examples/nodejs/storage/user-program.ts
```diff
@@ -0,0 +1,98 @@
...
+      //    } else if (storeValue instanceof StoreFloatValue) {
+      //      // ...
+      //    } else {
+      //      console.error('Unknown type');
+      //    }
+      // }
+      //
+      // I feel this pattern makes sense in a language with first-class pattern matching, such as
+      // Rust, Scala, Kotlin, etc.
+      // But in JavaScript, I feel the switch is at the right level of complexity.
```
cprice404, line 73: I agree that the switch seems like it would be nicer in typescript than the instanceof. It'd be really awesome though to find a way to be able to use a switch and still get the strong typing inside of the case branches. This SO post looks really interesting:

https://stackoverflow.com/a/56597745

I wonder if we could do that and sort of get the best of both worlds?

Follow up question will be: if we decide to go down one of these paths that differs from our `instanceof` pattern in the CacheClient, is it going to feel weird that the persistent store client has a very different API? If so, is there any forward-compatible way for us to make the CacheClient API feel more similar to whatever we decide to do here? Or, maybe the persistent store API lands in a 2.0 version of the SDK where the CacheClient one changes to match it?

examples/nodejs/storage/user-program.ts
```diff
@@ -0,0 +1,98 @@
...
+      // This introduces an extra level of indirection where the dev has to access the `Success`
+      // value and then the `value` field of the `StoreValue` object.
+    } else if (getResponse instanceof StoreGet.Error) {
```
cprice404, line 76: Since we are most likely down to a maximum of 2 different types available for any given response (Success/Error), I'm less certain now that it makes sense to model the errors as a return type. *Personally* I would still prefer that, but it's harder to make a case against the language idioms when it's only two types instead of 3.

In go/Rust, there are first-class patterns for this but we don't have that here. I assume if you play with the DDB, redis, or mongo node.js libs they will just throw and not introduce this indirection.

It'd be compelling to be able to get rid of the entire `if` block here for happy path examples, though I realize that would basically imply that `try/catch` was mandatory for production-quality code. It'd also have big implications for consistency of UX if it is shipped in the same package as the CacheClient :)

If nothing else you could move the `.tryGetValue*` methods up to the parent class so that they could be accessed directly on the response without having to do the instance check first?

examples/nodejs/storage/user-program.ts
```diff
@@ -0,0 +1,98 @@
...
+      console.log(response.tryGetValueInteger());
+    } catch (error: unknown) {
+      console.error(error);
+    }
+  } else if (response instanceof StoreGet.Error) {
+    console.error('Error:', response.message);
+  }
+}
+
+main().then().catch(console.error);
```
cprice404, line 98: thanks for putting this together. my other question is, as you were doing this, did you have any sense of whether this is going to feel good/bad/weird/different/better/worse vs. if we did not model the types directly? e.g. if the proto was just `bytes`? honestly on that front I feel like this exercise was a success, because we would still need to support `string | uint8Array` everywhere, and from my first read through this PR it seems like the challenges that we faced in surfacing that through the current CacheClient API are about the same as the challenges here (from an end-user type-checking perspective), and here we have stronger correctness guarantees. WDYT?

examples/nodejs/storage/user-program.ts
malandis: @bruuuuuuuce has a compelling PR open [here](https://github.com/momentohq/client-sdk-javascript/pull/1254) where explored there. There are TS options that can enforce exhaustive enum coverage on a switch.

examples/nodejs/storage/user-program.ts
malandis: > Since we are most likely down to a maximum of 2 different types available for any given response (Success/Error), I'm less certain now that it makes sense to model the errors as a return type. Personally I would still prefer that, but it's harder to make a case against the language idioms when it's only two types instead of 3.

One thing I punted on exploring here was how to surface the `not found` case. In the cache case we have:
- hit
- miss
- error, including cache not found

Whereas in this PR we have:
- success (hit)
- error, which includes both store not found, key not found, and the other gamut of errors.

The latter here is consistent with HTTP and gRPC response codes. We have a 200 for OK/resource found and a 404 for not found. So I would expect to be writing code like:
```typescript
if (getResponse instanceof StoreGet.Success) {
// ...
} else if (getResponse instanceof StoreGet.Error) {
   switch (getResponse.errorCode) {
       // auto gen'd based off of an enum with the most common at the top
      case StoreErrorCode.KEY_NOT_FOUND:
         //...
      case StoreErrorCode.STORE_NOT_FOUND:
        // ...
// ...
}
```

examples/nodejs/storage/user-program.ts
malandis: > If nothing else you could move the .tryGetValue* methods up to the parent class so that they could be accessed directly on the response without having to do the instance check first?

I thought about this and didn't feel good about the user having to distinguish accessing the wrong type vs the panoply of other errors that could have happened. This may be the lowest friction UX though:

```typescript
const getResponse = await client.get('my-store', 'my-key');
try {
    return getResponse.tryGetValueString();
} catch (GetResponseException e) {
   switch (e.errorCode) {
        case GetResponse.ErrorCode.WrongType:
            // ...
        case GetResponse.ErrorCode.KEY_NOT_FOUND:
          // ...
        case GetResponse.ErrorCode.STORE_NOT_FOUND:
         // ...
   }
}
```

malandis: > thanks for putting this together. my other question is, as you were doing this, did you have any sense of whether this is going to feel good/bad/weird/different/better/worse vs. if we did not model the types directly? e.g. if the proto was just bytes? honestly on that front I feel like this exercise was a success, because we would still need to support string | uint8Array everywhere, and from my first read through this PR it seems like the challenges that we faced in surfacing that through the current CacheClient API are about the same as the challenges here (from an end-user type-checking perspective), and here we have stronger correctness guarantees. WDYT?

I agree. Even if on the wire we only had bytes, users time and again have shown string is the default (possibly legacy from redis but that's another story). So we already have the problem of supporting more than one type on the SDK dev-facing API.

I really like the idea of having more types supported on the wire for portability. A developer could write an integer from a mobile frontend and read it from a web backend. If we support the types out of the box then they can write an integer directly without worrying about all the integer encoding nuances or marshalling to a neutral type and parsing themselves.

examples/nodejs/storage/user-program.ts
cprice404: re: switching on error codes, yes, agreed. But I think my bigger question is whether it even makes sense to have to do the instanceof check when we know there are only 2 possible cases.

To say it another way... I think before, when there were 3 cases, we argued that users were going to need an `if` block no matter what, so it made more sense to push the error into a 3rd branch of the if statement rather than forcing them to wrap the `if` with a `try/catch`.

But in this case, users don't necessarily need the `if` block at all if they only care about Success. and adding a try/catch around the API call is no longer (necessarily) any more unwieldy than adding the `if` you're using here.

So I'm asking whether it should just be `const value = await client.get(storeName, key);`

examples/nodejs/storage/user-program.ts
cprice404: > I'd say the store API has much less surface area than the cache

The Cache API started with only get/set. :)

The Store API is going to grow over time and very well may end up supporting the same types of collections and such. So we should make sure that the current minimal size of the API is not a precondition to us preferring the explicitly modeled data types :)

examples/nodejs/storage/user-program.ts
malandis: I feel like `const value = await client.get(storeName, key);` is broadly more idiomatic and less cluttered. I think users would expect a missing value to be `undefined` though (ie the return value is optional), in which case they can branch on `undefined` or assume it's there.

That said, if we do that, then we lose forward compatibility for this API.

cprice404: In the CacheClient we support `.value()` on the response for those who want to short-circuit the pattern matching / type checking stuff, maybe that's sufficient here too

PR #1254, opened 2024-04-26
feat: explicit success types

bruuuuuuuce:


bruuuuuuuce:
If you switch on the type it can infer the value
image (6)


if you don't switch on the type its a union of all possible values
image (7)

examples/nodejs/storage/user-program.ts
```diff
@@ -9,82 +9,25 @@ async function main() {
...
+            break;
+          case StoreValueType.BOOLEAN:
+            console.log('The value is a boolean:', getResponse.value);
+            break;
+        }
+        console.log(`The value is: ${getResponse.value} and the type is ${getResponse.type}`);
+        break;
+      case StoreGet.ResponseType.Error:
+        console.error('Error:', getResponse.message);
+        break;
```
cprice404, line 31: can you add some comments in here? I want to know things like:

* is there a flag that you find that makes the compiler check for exhaustive branches?
* if you have that flag enabled and you omit a branch what happens?
* it seems that the trick here (and from that SO post that I linked on MIchael's) is to have a nested field that is basically an enum type that differentiates the different response types, so you'd always be switching on that field but then accessing other (sibling) fields of the response object in the case branches, does that match up with your understanding?
* If we decided we loved this but that it feels too inconsistent with the existing SDK, can you imagine a way we could augment the existing CacheResponse types to support this pattern? (I think "yes" but would like to see/hear your thoughts)

bruuuuuuuce: 1. Yes there is a rule here https://typescript-eslint.io/rules/switch-exhaustiveness-check/
2. This is what it looks like
image
3. Yes this is also my understanding, but the way that they do it in the stack overflow post, I am not sure if there is any way to guarantee exhaustive check like there is with the enum
4. I need to play around with this a bit before I can say for sure, I can do that and get back to you on this thread with an example if I can get it working

PR #1257, opened 2024-04-29
feat: show automatic decompression in example

nand4011:
Update the compression example to show automatic decompression, as well as overwriting the automatic setting with the method option.


rishtigupta (approved):
lgtm! :shipit:

rishtigupta (approved):

PR #1258, opened 2024-04-30
feat: spike on converting existing responses to new pattern

bruuuuuuuce:
This pr stems from a conversation on another [pr](https://github.com/momentohq/client-sdk-javascript/pull/1254#discussion_r1583835211), trying to see if we can move forward with this new matching pattern with existing responses, without breaking backwards compatibility


examples/nodejs/cache/advanced.ts
```diff
@@ -91,8 +91,6 @@ async function setGetDeleteExample() {
     logger.info('cache miss');
   } else if (getResponse instanceof CacheGet.Error) {
     logger.info(`Error: ${getResponse.message()}`);
-  } else {
```
bruuuuuuuce, line 94: with this new pattern, this now errors out since we have exhausted all type possibilities, and the `getResponse` type is not of type `never`

image (8)

packages/core/src/messages/responses/cache-dictionary-fetch.ts
```diff
@@ -187,4 +187,16 @@ class _Error extends Response {
  * - `message()` - a human-readable description of the error
  * - `innerException()` - the original error that caused the failure; can be re-thrown.
  */
-export class Error extends ResponseError(_Error) {}
+export class Error extends BaseResponseError implements IResponse {
+  constructor(_innerException: SdkError) {
+    super(_innerException);
+  }
+
+  responseType = ResponseType.Error;
```
bruuuuuuuce, line 195: again don't fully love the name `responseType` happy to change it to something else. Maybe `type` or just `response`

PR #1264, opened 2024-05-06
feat: add (explicit) zstd compression extensions

cprice404:
After discussion we've decided that we want to have a compression
extension that is really easy to get up and running with, and
doesn't require special packaging considerations. The current
implementation, based on `zstd`, requires a binary dependency
and special packaging considerations because `zstd` is not
supported in the node.js stdlib. Therefore, we are going to
migrate the "default" compression extensions to use gzip, which
is supported in the stdlib and thus won't require any special
considerations for packaging.

This commit is the first step toward that; in this commit we
simply make a copy of the existing compression extensions and
change the package name to include a `-zstd` suffix. Once this
package is deployed to npmjs, we can migrate existing users
to it and then re-purpose the original one.


packages/client-sdk-nodejs-compression-zstd/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@gomomento/sdk-nodejs-compression-zstd",
+  "version": "0.0.1",
+  "description": "Compression extensions for Momento node.js SDK",
```
malandis, line 4: Include details on that this is for zstd?

packages/client-sdk-nodejs-compression-zstd/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "unit-test": "jest unit",
+    "integration-test": "jest integration",
+    "build-deps": "cd ../core && npm run build && cd - && cd ../client-sdk-nodejs && npm run build && cd -",
+    "build-and-run-tests": "npm run build-deps && jest",
+    "lint": "eslint . --ext .ts",
+    "format": "eslint . --ext .ts --fix",
+    "watch": "tsc -w",
+    "build": "rm -rf dist && tsc"
+  },
+  "author": "",
```
malandis, line 27: Momento authors

bruuuuuuuce (dismissed):
:shipit:

packages/client-sdk-nodejs-compression-zstd/README.md
```diff
@@ -0,0 +1,55 @@
+<img src="https://docs.momentohq.com/img/momento-logo-forest.svg" alt="logo" width="400"/>
+
+[![project status](https://momentohq.github.io/standards-and-practices/badges/project-status-official.svg)](https://github.com/momentohq/standards-and-practices/blob/main/docs/momento-on-github.md)
+[![project stability](https://momentohq.github.io/standards-and-practices/badges/project-stability-stable.svg)](https://github.com/momentohq/standards-and-practices/blob/main/docs/momento-on-github.md)
+
+
+# Compression Extensions for Momento Node.js SDK
```
malandis, line 7: with zstd

packages/client-sdk-nodejs-compression-zstd/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+This package provides extensions for the Momento Node.js SDK to support compression and decompression of data that
+is stored using the `CacheClient.set` function and accessed through the `CacheClient.get` function.
+
+To use the library, you will need to install it from npm:
+
+```bash
+npm install @gomomento/sdk-compression
```
malandis, line 15: Is this still the zstd package id?

nand4011 (approved):

malandis (commented):
Few questions about naming

packages/client-sdk-nodejs-compression-zstd/README.md
cprice404: the README.md file gets generated from the README.template.md file. check that one, I think I got everything there.

packages/client-sdk-nodejs-compression-zstd/package.json
cprice404: this is the same setup we use for all the other packages for now. we can change it at some point if we want to but seems like a separate thing

packages/client-sdk-nodejs-compression-zstd/package.json
cprice404: done in https://github.com/momentohq/client-sdk-javascript/pull/1264/commits/93d0c1a97da7fde7e4a9a225ea7b2a7caab890d4

malandis (approved):

PR #1266, opened 2024-05-07
fix: add javascript patterns examples files

bruuuuuuuce:
adds `patterns.ts` example file to be consumed downstream in dev docs

### DIFF
--- a/examples/nodejs/cache/doc-example-files/patterns.ts
+++ b/examples/nodejs/cache/doc-example-files/patterns.ts
@@ -8,7 +8,7 @@
  * a second file in the future if desired.
  *
  */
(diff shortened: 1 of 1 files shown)


malandis:
There are various warnings reported by GH actions, eg unused imports and non-null assertions. Were these intended?

bruuuuuuuce:
> There are various warnings reported by GH actions, eg unused imports and non-null assertions. Were these intended?

hmmm no, the non-null assertion is expected, i can probably ignore that with a lint rule. The unused imports should be removed good catch

malandis:
> hmmm no, the non-null assertion is expected, i can probably ignore that with a lint rule. The unused imports should be removed good catch

Thanks. Those appear in GH actions in every subsequent PR involving those files if left unaddressed.

malandis (approved):
🏆

PR #1267, opened 2024-05-07
feat: Switch default compression to gzip

cprice404:
Now that we have the @gomomento/sdk-nodejs-compression-zstd package,
this commit updates the default compression extensions package
to use gzip as the compression algorithm rather than zstd. This
allows us to get rid of the binary dependency that makes packaging
complicated. We made the decision to do this after benchmarking
showed very little difference in compression ratio and only a little
difference in performance between the two compression libraries;
therefore, shedding the packaging complexity for a simpler user
experience seemed like the best UX choice.


packages/client-sdk-nodejs-compression/package.json
```diff
@@ -48,8 +48,7 @@
   },
   "dependencies": {
     "@gomomento/generated-types": "0.107.4",
-    "@gomomento/sdk": "file:../client-sdk-nodejs",
-    "@mongodb-js/zstd": "1.2.0"
+    "@gomomento/sdk": "file:../client-sdk-nodejs"
```
malandis, line 51: Since zlib is part of the node runtime, does this whole package need to be an external dependency then? Couldn't it be included with the nodejs client?

malandis (approved):
No blockers

packages/client-sdk-nodejs-compression/package.json
cprice404: we talked about this in the team meeting the other day. for now i'd prefer that users need to stop and consider options, so i like it being separate.

PR #1268, opened 2024-05-08
feat: make RequestLoggingMiddleware non-experimental and call toString on MiddlewareMessages

anitarua:
Addresses https://github.com/momentohq/dev-eco-issue-tracker/issues/811

Adds a `toString` method to `MiddlewareMessage` that prints extra details for `CacheClient` and `LeaderboardClient` requests (as those are the only two with middlewares enabled for now).

Affects only the NodeJS SDK, similar changes can be made in web sdk as desired.

Snippet of what the logs look like now:
```
[2024-05-08T00:24:36.541Z] DEBUG (Momento: RequestLoggingMiddleware): Logging middleware: request 43 onRequestMetadata: {"cache":["js-sdk-doc-examples-cache-77092f86e4002a9e"]}
[2024-05-08T00:24:36.541Z] DEBUG (Momento: RequestLoggingMiddleware): Logging middleware: request 43 onRequestBody: request type: _DictionarySetRequest, request size: 55, details: DictionarySetRequest with dictionary name "test-dictionary" and ttl 60000 ms and refresh ttl true and fields (field "key1" | value "value1"), (field "key2" | value "value2"), 
[2024-05-08T00:24:36.592Z] DEBUG (Momento: RequestLoggingMiddleware): Logging middleware: request 43 onResponseMetadata: {"content-type":["application/grpc"],"access-control-allow-credentials":["true"],"vary":["origin, access-control-request-method, access-control-request-headers"],"access-control-expose-headers":["grpc-status,grpc-message,grpc-encoding,grpc-accept-encoding"],"date":["Wed, 08 May 2024 00:24:35 GMT"]}
[2024-05-08T00:24:36.592Z] DEBUG (Momento: RequestLoggingMiddleware): Logging middleware: request 43 onResponseBody: response type: _DictionarySetResponse, response size: 0
[2024-05-08T00:24:36.592Z] DEBUG (Momento: RequestLoggingMiddleware): Logging middleware: request 43 onResponseStatus: status code: 0
```

I've also got a full output file of log statements from running the doc example snippets file with this middleware enabled, just let me know if you want it!


packages/client-sdk-nodejs/src/config/middleware/request-logging-middleware.ts
```diff
@@ -27,11 +25,12 @@ class ExperimentalLoggingMiddlewareRequestHandler
   }
   onRequestBody(request: MiddlewareMessage): Promise<MiddlewareMessage> {
     this.logger.debug(
-      'Logging middleware: request %s onRequestBody: request type: %s, request size: %s',
+      'Logging middleware: request %s onRequestBody: request type: %s, request size: %s, details: %s',
       this.requestId,
       // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-      request.constructor.name,
```
anitarua, line 33: this used to just print `MiddlewareMessage` rather than the request type

packages/client-sdk-nodejs/src/config/middleware/middleware.ts
```diff
@@ -34,6 +37,548 @@ export class MiddlewareMessage {
     }
     return 0;
   }
+
+  constructorName(): string {
+    return this._grpcMessage.constructor.name;
+  }
+
+  // Note: APIs that use streaming interceptors (e.g. GetBatch and SetBatch)
```
anitarua, line 45: there were no middleware messages shown for GetBatch or SetBatch when I tested locally

packages/client-sdk-nodejs/src/config/middleware/middleware.ts
```diff
@@ -34,6 +37,548 @@ export class MiddlewareMessage {
...
+        const request = this._grpcMessage as cache.cache_client._GetRequest;
+        return `GetRequest with key "${TEXT_DECODER.decode(
+          request.cache_key
+        )}"`;
+      }
+      case cache.cache_client._SetRequest: {
+        const request = this._grpcMessage as cache.cache_client._SetRequest;
+        return `SetRequest with key "${TEXT_DECODER.decode(
+          request.cache_key
+        )}" and value "${TEXT_DECODER.decode(request.cache_body)}" and ttl ${
```
pgautier404, line 59: `request.cache_body` could contain megabytes of data. We used a `truncateString()` utility method in the response classes to limit the amount of output from `toString()`. I'm guessing in a situation where someone is using a middleware logger they're going to want to see absolutely everything, but I thought it might be worth discussion.

cprice404: that's a good callout. I think for now let's just leave it the way Anita has it  (logging the whole item) and see if we get feedback.

packages/client-sdk-nodejs/src/config/middleware/middleware.ts
```diff
@@ -34,6 +37,548 @@ export class MiddlewareMessage {
...
+          )
+        );
+      }
+      case cache.cache_client._SetIfRequest: {
+        const request = this._grpcMessage as cache.cache_client._SetIfRequest;
+        return `SetIfRequest with key "${TEXT_DECODER.decode(
+          request.cache_key
+        )}" and value "${TEXT_DECODER.decode(
+          request.cache_body
+        )}" and condition "${request.condition}"`;
```
pgautier404, line 103: This is missing the `ttl_milliseconds`. I think it might also be worthwhile to log the condition value here for requests that require one like `present_and_not_equal`.

packages/client-sdk-nodejs/src/config/middleware/middleware.ts
```diff
@@ -34,6 +37,548 @@ export class MiddlewareMessage {
...
     return 0;
   }
+
+  constructorName(): string {
+    return this._grpcMessage.constructor.name;
+  }
+
+  // Note: APIs that use streaming interceptors (e.g. GetBatch and SetBatch)
+  // will not see these debug messages
+  toString(): string {
```
cprice404, line 47: thank you so much for doing all of this! i have a few suggestions, i apologize for not discussing this with you in more depth before you started on the ticket. If any of these seem too onerous, they are not urgent, and I can add a commit and some point to help out with the tedious parts. :)

1. For detailed logs like this it's really valuable to structure the log output as JSON or some other format that is easy to parse programmatically. So, I know it's going to be repetitive, but I'd probably define an `interface` for each one, like `GetRequestLogFormat`, and then the implementation should create an instance of the interface and then call `JSON.stringify` on it.
2. JSON.stringify might fail on binary data so we should test that briefly before doing all of the rest of the work, and if it does, decide what we want to do about it. I have a few ideas we can discuss if it happens.
3. We will need a big switch statement no matter what, but I think it would be best if each case in the switch just called a single function to do the conversion for the given request/response type, so that we have a separate function that does the actual work for each conversion.
4. It would be really awesome to have a unit test that would assert that we have defined the necessary conversion functions for all of the possible types. kinda like what we are doing here: https://github.com/momentohq/client-sdk-javascript/blob/0831824a2c0efd4f4730258b4eeae1d1a914c4a6/packages/core/test/unit/clients/imomentocache-interface.test.ts#L15-L19 . That way if a new release of the protos adds a new request/response type, we will get a compile-time reminder to add support for it.

I know the list above is a lot of tedious changes so feel free to approach it incrementally in between other tasks, or ask for volunteers to help with some subset of the types, etc. (I am very willing to contribute!)

packages/client-sdk-nodejs/src/config/middleware/middleware.ts
```diff
@@ -34,6 +37,548 @@ export class MiddlewareMessage {
...
+            request.dictionary_name
+          )}" and ttl ${request.ttl_milliseconds} ms and refresh ttl ${
+            request.refresh_ttl ? 'true' : 'false'
+          } and fields ` +
+          request.items.reduce(
+            (acc, item) =>
+              acc +
+              `(field "${TEXT_DECODER.decode(
+                item.field
+              )}" | value "${TEXT_DECODER.decode(item.value)}"), `,
```
pgautier404, line 184: Same comment as above re: potentially truncating these values, as they might be very large and there may be a lot of them.

packages/client-sdk-nodejs/src/config/middleware/middleware.ts
anitarua: Thanks for the feedback! And my bad for not checking in before diving in first haha 😅

Sounds like a good plan for revisions. It looks like `JSON.stringify` would treat binary data like a list:
```
const binaryData = Uint8Array.of(108, 111, 108);
console.log(JSON.stringify({ data: binaryData }));
// prints '{"data":{"0":108,"1":111,"2":108}}'
```

But I think we'll be able to get around this using the TextDecoder when creating an instance of each `*RequestLogFormat` interface:
```
console.log(JSON.stringify({ data: TEXT_DECODER.decode(binaryData) }));
// prints '{"data":"lol"}'
```

cprice404: sounds good. TEXT_DECODER.decode might throw an error if the bytes aren't utf-8 though, so we might have to figure out what we want to do for those cases.

cprice404: and no worries re: checking in, I should have put a note in the ticket description.

packages/client-sdk-nodejs/src/config/middleware/middleware.ts
```diff
@@ -34,6 +37,548 @@ export class MiddlewareMessage {
...
+        );
+      }
+      case cache.cache_client._SetDifferenceRequest: {
+        const request = this
+          ._grpcMessage as cache.cache_client._SetDifferenceRequest;
+        return (
+          `SetDifferenceRequest with set name "${TEXT_DECODER.decode(
+            request.set_name
+          )}" and elements ` +
+          request.subtrahend.set.elements.reduce(
```
pgautier404, line 256: I think you also need to add a section for `minuend` here.

packages/client-sdk-nodejs/src/config/middleware/middleware.ts
```diff
@@ -34,6 +37,548 @@ export class MiddlewareMessage {
...
+      }
+      case cache.cache_client._UpdateTtlRequest: {
+        const request = this
+          ._grpcMessage as cache.cache_client._UpdateTtlRequest;
+        return `UpdateTtlRequest with key "${TEXT_DECODER.decode(
+          request.cache_key
+        )}" and amount "${
+          String(request.increase_to_milliseconds) + ' (increase)' ||
+          String(request.decrease_to_milliseconds) + ' (decrease)' ||
+          String(request.overwrite_to_milliseconds) + ' (overwrite)'
```
pgautier404, line 133: It looks to me like this will always output the first and only the first of these strings. If I run:
```
console.log(`${String(0) || 'hello' || 'world'}`);
```
it outputs `0`.

packages/client-sdk-nodejs/src/config/middleware/middleware.ts
```diff
@@ -34,6 +37,548 @@ export class MiddlewareMessage {
...
+        return `SortedSetLengthByScoreRequest with set name "${TEXT_DECODER.decode(
+          request.set_name
+        )}" and min score ${
+          String(request.inclusive_min) + ' (inclusive)' ||
+          String(request.exclusive_min) + ' (exclusive)' ||
+          'unbounded'
+        } and max score ${
+          String(request.inclusive_max) + ' (inclusive)' ||
+          String(request.exclusive_max) + ' (exclusive)' ||
+          'unbounded'
```
pgautier404, line 506: I believe these will also only ever output the first of the strings.

packages/client-sdk-nodejs/src/config/middleware/middleware.ts
anitarua: confirmed conditions are logged as strings

```
[2024-05-13T19:26:52.702Z] DEBUG (Momento: RequestLoggingMiddleware): Logging middleware: request 14 onRequestBody: request type: _SetIfRequest, request size: 44, details: {"key":"test-key","value":"test-field","condition":"present_and_not_equal","ttl_milliseconds":60000}
```

anitarua:
Current state of the PR: 

* I've made a first pass at creating logging format interfaces and conversion functions for each request type.
    * Note: there were a few overlapping ones (e.g. Get, Delete, etc) so I made a common one for those, but if completely separate is preferred, that's fine too.
* The switch statement calls `JSON.stringify(<conversion function>)` for each request type.
* Does not yet handle the case of: "TEXT_DECODER.decode might throw an error if the bytes aren't utf-8 though"
* Does not yet include a unit test to assert that all desired conversion functions are present

anitarua:
Example output file from running docs examples snippets locally:
[output.txt](https://github.com/momentohq/client-sdk-javascript/files/15299231/output.txt)

packages/client-sdk-nodejs/src/config/middleware/request-logging-middleware.ts
anitarua: I think there are probably a few other places in the middleware where calling `constructorName()` will be more useful but haven't tracked them all down yet

cprice404:
rebased, now will pick up from where anita left off.

cprice404:
I think I'm done with this for now. It's not perfect and probably warrants some discussion on Monday but I think it's probably a reasonable direction.

packages/client-sdk-nodejs/src/config/middleware/middleware.ts
```diff
@@ -34,6 +42,26 @@ export class MiddlewareMessage {
...
   }
+
+  constructorName(): string {
+    return this._grpcMessage.constructor.name;
+  }
+
+  // Note: APIs that use streaming interceptors (e.g. GetBatch and SetBatch)
+  // will not see these debug messages
+  toLogFormat(): RequestLog {
+    const requestToLogConverter = CacheRequestToLogInterfaceConverter.get(
```
anitarua, line 53: ooh much nicer than the giant switch statement

packages/client-sdk-nodejs/test/unit/config/middleware/cache-request-logging.test.ts
```diff
@@ -0,0 +1,847 @@
...
+        minScore: 41,
+        minScoreExclusive: false,
+        maxScore: 'unbounded',
+        maxScoreExclusive: undefined,
+      });
+    });
+  });
+
+  describe('supported cache request types', () => {
+    it('should have a converter for all known cache request types', () => {
```
anitarua, line 815: 🙌

bruuuuuuuce:
Spent some time playing around with this pr this morning, I think it looks good. I really like that last test that was added to ensure we have logging for all request types. I wonder if it would be good in our canaries/integration tests to enable this middleware so we have it being tested across all of our apis in an environment on a regular cadence.

One nit that I saw was the jsdoc for the class

`RequestLoggingMiddleware`

Still has comments about how it is experimental.

```
This middleware implements per-request logging which can be used for
debugging.  The log format is currently considered experimental; in a
future release, once the log format is considered stable, this class will
be renamed to remove the Experimental prefix.
```

Once we remove this text blurb I am :shipit: on this pr

anitarua:
@bruuuuuuuce I removed that part in https://github.com/momentohq/client-sdk-javascript/pull/1268/commits/cdf6b28ee62d35cb40f41e8ccbff70f4f6276798

bruuuuuuuce (approved):
:shipit:

PR #1269, opened 2024-05-08
fix: convert first few responses to new exhaustive enum pattern

bruuuuuuuce:
Converts

- cache get
- cache set
- cache dictionary fetch
- cache dictionary get field
- cache dictionary get fields

to new exhaustive enum response pattern


packages/core/src/messages/responses/enums/index.ts
```diff
@@ -0,0 +1,28 @@
+export enum MomentoResponse {
+  Success = 'Success',
+  Error = 'Error',
+}
+
+export enum CacheGetResponse {
```
bruuuuuuuce, line 6: this is a bit weird, maybe we can just have like a `CacheFetchResponse` for generic hit, miss, error responses

packages/core/src/messages/responses/enums/index.ts
```diff
@@ -0,0 +1,28 @@
+export enum MomentoResponse {
```
bruuuuuuuce, line 1: we should name this something else that is not `Response` since node has a default response object, and it confuses the "compiler"

packages/core/src/messages/responses/cache-set.ts
```diff
@@ -45,4 +23,12 @@ class _Error extends Response {
  * - `message()` - a human-readable description of the error
  * - `innerException()` - the original error that caused the failure; can be re-thrown.
  */
-export class Error extends ResponseError(_Error) {}
+export class Error extends BaseResponseError implements IResponse {
+  constructor(_innerException: SdkError) {
+    super(_innerException);
+  }
+
+  readonly type: MomentoResponse.Error;
```
bruuuuuuuce, line 31:
```suggestion
  readonly type: MomentoResponse.Error = MomentoResponse.Error;
```

anitarua (approved):
got it working and it lgtm! really like the exhaustive switch IDE hint 👍

PR #1280, opened 2024-05-14
feat: Add support for numConnections in TopicClient configuration

cprice404:
To ensure even load distribution on the servers, we need to make sure that it's possible to create multiple gRPC channels in TopicClient. Prior to this commit, it was not. This commit adds support.

This commit also duplicates the cache TransportStrategy and GrpcStrategy, and makes Topic-specific versions of them. I did not think it would be a good idea to re-use the existing ones because there were so many settings on them that don't necessarily seem relevant for topics, and would require some thought and testing before supporting. The topic-specific versions of these configs only support numGrpcChannels for now.

This commit also adds a more user-friendly copy constructor, `withNumConnections`, on the top-level configuration object. Users have been needing access to this config setting with enough frequency that I think it's time to promote its visibility and make it simpler to configure it.


eaddingtonwhite (approved):

cprice404:
rebased

pgautier404 (approved):

PR #1283, opened 2024-05-15
fix: require explicit override keyword for overriding abstract method properties/functions

bruuuuuuuce:
closes https://github.com/momentohq/dev-eco-issue-tracker/issues/831 This pr addresses/fixes the following inconsistencies in this repo

- every tsconfig.json file should have the `noImplicitOverride` and the `noFallthroughCasesInSwitch` flags set to `true`. 
  - The `noImplicitOverride` tsconfig setting requires that a user specifies the `override` keyword when overriding a function or property attribute. 
  - The `noFallthroughCasesInSwitch` prevents the user from writing a code block that looks like
```
switch(value) {
  default: { console.log('default'); }
}
```

- adds the `override` keyword in all places where we are overriding a method/property of a class to satisfy the updated `tsconfig.json` rule
- changes the `test nodejs` step to use the `macos-latest` runner so that it doesn't time out after 4 min

image

### DIFF
--- a/examples/web/vite-chat-app/tsconfig.node.json
+++ b/examples/web/vite-chat-app/tsconfig.node.json
@@ -4,7 +4,9 @@
     "skipLibCheck": true,
     "module": "ESNext",
     "moduleResolution": "bundler",
-    "allowSyntheticDefaultImports": true
+    "allowSyntheticDefaultImports": true,
+    // requires the `override` keyword to override a method/property from a parent class
+    "noImplicitOverride": true
   },
   "include": ["vite.config.ts"]
 }
--- a/examples/web/vite-chat-app/tsconfig.node.json
+++ b/examples/web/vite-chat-app/tsconfig.node.json
@@ -6,7 +6,8 @@
     "moduleResolution": "bundler",
     "allowSyntheticDefaultImports": true,
     // requires the `override` keyword to override a method/property from a parent class
-    "noImplicitOverride": true
+    "noImplicitOverride": true,
+    "noFallthroughCasesInSwitch": true
   },
   "include": ["vite.config.ts"]
 }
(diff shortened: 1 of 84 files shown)


examples/web/vite-chat-app/tsconfig.node.json
```diff
@@ -4,7 +4,10 @@
     "skipLibCheck": true,
     "module": "ESNext",
     "moduleResolution": "bundler",
-    "allowSyntheticDefaultImports": true
+    "allowSyntheticDefaultImports": true,
+    // requires the `override` keyword to override a method/property from a parent class
```
pgautier404, line 8: Github highlighted this in red, presumably because comments aren't allowed in JSON. I'm not sure why it only flagged this one, but this comment should probably be removed wherever it appears in JSON.

bruuuuuuuce: I can remove it here, I know that sometimes json does allow comments, but if you look at the default tsconfig file that gets generated for you by `tsc --init`, it actually contains lots of comments like [this](https://github.com/momentohq/client-sdk-javascript/pull/1283/files#diff-1dbbd8a9cf2a108111f1ad4a08c47f8af4934f44491121d86262e9b90f91a2b5R95-R98). I suspect that github doesn't recognize the `tsconfig.node.json` filename, and thinks its just a regular json file. I think removing it here is the correct thing to do since clearly github doesn't like it. But I do think the comment is useful, and the rest of the `tsconfig.json` files should be ok with comments in them

bruuuuuuuce, line 8:
```suggestion
```

pgautier404: Ah, ok, good to know. Thanks.

pgautier404 (approved):
Thank you, this was obviously a ton of tedious work.

PR #1284, opened 2024-05-15
chore: add example for using zstd compression in nodejs lambda

cprice404:
This commit adds an advanced compression example, which illustrates
how to package your code for use in AWS lambda when you are using
the `@gomomento/sdk-nodejs-compression-zstd` library. It walks
through the steps required to make sure that you package the
correct binary for the target platform, even if you are doing
the build from a different platform.


bruuuuuuuce (approved):
:shipit:

PR #1290, opened 2024-05-20
Fix: Ensure endpoints starting with http:// are correctly prefixed wi…

lucas-pelton:
…th https://

Previously, endpoints starting with http:// were prefixed with https://, which could lead to incorrect URL formation. This commit adjusts the prefix process to use regex to handle these cases properly.

### DIFF
--- a/examples/cloudflare-workers/http-api/src/worker.ts
+++ b/examples/cloudflare-workers/http-api/src/worker.ts
@@ -13,7 +13,11 @@ class MomentoFetcher {
 	private readonly baseurl: string;
 	constructor(key: string, endpoint: string) {
 		this.apiKey = key;
-		const correctedEndpoint = endpoint.replace(/^(http:\/\/)?/, "https://");
+		if ( !( endpoint.startsWith('https://') || endpoint.startsWith('http://') ) ) {
+			this.baseurl = `https://${endpoint}/cache`;
+		} else {
+			this.baseurl = `${endpoint}/cache`;
+		}
 		this.baseurl = `${correctedEndpoint}/cache`;
 	}
 
--- a/examples/cloudflare-workers/http-api/src/worker.ts
+++ b/examples/cloudflare-workers/http-api/src/worker.ts
@@ -13,11 +13,8 @@ class MomentoFetcher {
 	private readonly baseurl: string;
 	constructor(key: string, endpoint: string) {
 		this.apiKey = key;
-		if (!endpoint.startsWith('https://')) {
-			this.baseurl = `https://${endpoint}/cache`;
-		} else {
-			this.baseurl = `${endpoint}/cache`;
-		}
+		const correctedEndpoint = endpoint.replace(/^(http:\/\/)?/, "https://");
+		this.baseurl = `${correctedEndpoint}/cache`;
 	}
 
 	async get(cacheName: string, key: string) {
--- a/examples/cloudflare-workers/http-api/src/worker.ts
+++ b/examples/cloudflare-workers/http-api/src/worker.ts
@@ -18,7 +18,6 @@ class MomentoFetcher {
 		} else {
 			this.baseurl = `${endpoint}/cache`;
 		}
-		this.baseurl = `${correctedEndpoint}/cache`;
 	}
 
 	async get(cacheName: string, key: string) {


examples/cloudflare-workers/http-api/src/worker.ts
```diff
@@ -13,11 +13,8 @@ class MomentoFetcher {
 	private readonly baseurl: string;
 	constructor(key: string, endpoint: string) {
 		this.apiKey = key;
-		if (!endpoint.startsWith('https://')) {
-			this.baseurl = `https://${endpoint}/cache`;
-		} else {
-			this.baseurl = `${endpoint}/cache`;
-		}
+		const correctedEndpoint = endpoint.replace(/^(http:\/\/)?/, "https://");
```
cprice404, line 16: thanks for the submission!

I don't think this patch correctly handles the case where the URL already begins with `https://`, and I'm also not 100% sure whether we want to forcefully change the URL to `https://` if the caller explicitly gave us `http://`?

```
 npx ts-node
> "foo.bar.com".replace(/^(http:\/\/)?/, "https://");
'https://foo.bar.com'
> "https://foo.bar.com".replace(/^(http:\/\/)?/, "https://");
'https://https://foo.bar.com'
>
```

I think maybe another idea would be to do:

```typescript
if (!(endpoint.startsWith('https://') || endpoint.startsWith('http://')) {
...
```

That would leave either of the two protocols intact if a user passed them, while still adding `https://` prefix if they did not. WDYT?

cprice404: I agree that parsing the URL is probably a more correct way to do it. In another environment, I would definitely write the code the way you have here.

I suspect the full URL parsing is quite a bit more expensive than just a `.startsWith`, though, and in a CloudFlare environment I wonder if the perf difference could matter. That's probably why we did it this way initially.

The latency of the HTTP request itself probably dwarfs the URL parsing though, so what I said above is probably a micro-optimization :)

Momento doesn't currently support plaintext endpoints but it's theoretically possible we could someday. So overall my thoughts are:

* probably best not to change the protocol that the user expressed if they expressed one
* parsing the URL is definitely the most correct way to handle it
* perf implications of parsing the URL vs startsWith are probably not really a big deal.

So we can change it either to what you suggested above, or what I suggested in my prev comment, I would be fine with either!

lucas-pelton: Excellent observations. Let's go with your suggestion of `if ( !( endpoint.startsWith('https://') || endpoint.startsWith('http://') ) ) {`

- I think it's clearer what's being tested in the conditional
- Rebuilding the original endpoint from the parsed URL components is quite verbose
- The opportunity to throw an error for an improperly formatted URL that was probably copy/pasted from an admin panel seems like an opportunity that will be seldom taken.

Do note that your inline code suggestion was lacking a third closing parenthesis, which is fixed in my inline code above.

Thanks for your time!

cprice404: sure thing, thanks for pointing out the issue!

cprice404, line 16:
```suggestion
		if ( !( endpoint.startsWith('https://') || endpoint.startsWith('http://') ) ) {
			this.baseurl = `https://${endpoint}/cache`;
		} else {
			this.baseurl = `${endpoint}/cache`;
		}
```

examples/cloudflare-workers/http-api/src/worker.ts
```diff
@@ -13,11 +13,8 @@ class MomentoFetcher {
...
+		this.baseurl = `${correctedEndpoint}/cache`;
```
cprice404, line 17:
```suggestion
```

lucas-pelton:
Rad! This is my first "successful" PR on github. Thanks!

cprice404:
if you can accept those suggestions above i will kick off our CI and try to get this merged!

Our CI might fail because it relies on some secrets that are only available in our org - we are hoping to fix this soon. But let's see what happens and we'll get it merged one way or the other.

lucas-pelton:
Sorry... This being my first PR, I'm actually not sure of how it actually goes. I clicked all the "Resolve Conversation" buttons, but I'm not confident that is what you're asking me to do... 😉

cprice404:
Ah! So basically I had done some in-line edits to suggest the changes that we were discussing. You would want to click the "commit suggestion" button in order to accept those suggestions.

Alternately you can add a commit that makes those changes.

Alternately I can pull this over to my own branch and make the changes, but I didn't want you to miss out on getting your first PR merged :)

lucas-pelton:
Ok. I appreciate your help and supporting me get me a PR! :-)
I think I did it right. If not, just do your own merge. You have enough to do maintaining OSS to drag me through GitHub bootcamp. 

Thanks again!

cprice404:
moved this to #1306 . Thanks for the contribution @lucas-pelton !

PR #1293, opened 2024-05-28
feat: add withMomentoLocal and option to connect without TLS to CredentialProvider

anitarua:
Allows a user to specify whether each overridden endpoint should use TLS or not. 

~~Also changes `StringMomentoTokenProvider` to detect whether `:443` is contained in the endpoint to determine secure vs insecure when given no overrides.~~
Removed the auto-detect port `443` because it turns out endpoints like `cache.etc.etc.momentohq.com` don't contain the port number.

### DIFF
--- a/packages/client-sdk-nodejs/package-lock.json
+++ b/packages/client-sdk-nodejs/package-lock.json
@@ -42,6 +42,7 @@
       }
     },
     "../common-integration-tests": {
+      "name": "@gomomento/common-integration-tests",
       "version": "0.0.1",
       "dev": true,
       "license": "Apache-2.0",
@@ -73,6 +74,7 @@
       }
     },
     "../core": {
+      "name": "@gomomento/sdk-core",
       "version": "0.0.1",
       "license": "Apache-2.0",
       "dependencies": {
--- a/packages/client-sdk-nodejs/package-lock.json
+++ b/packages/client-sdk-nodejs/package-lock.json
@@ -42,7 +42,6 @@
       }
     },
     "../common-integration-tests": {
-      "name": "@gomomento/common-integration-tests",
       "version": "0.0.1",
       "dev": true,
       "license": "Apache-2.0",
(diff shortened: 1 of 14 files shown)


packages/core/src/auth/credential-provider.ts
```diff
@@ -197,20 +233,56 @@ export class StringMomentoTokenProvider extends CredentialProviderBase {
...
+    return this.allEndpoints.tokenEndpoint.secureConnection;
   }
 
   areEndpointsOverridden(): boolean {
     return this.endpointsOverridden;
   }
+
+  withMomentoLocal(): CredentialProvider {
+    const momentoLocalOverride = {
+      endpoint: '127.0.0.1:8080',
```
anitarua, line 274: this is hardcoded for now but can be made configurable later when it's configurable on momento-local

packages/client-sdk-nodejs/package-lock.json
```diff
@@ -42,6 +42,7 @@
       }
     },
     "../common-integration-tests": {
+      "name": "@gomomento/common-integration-tests",
```
anitarua, line 45: added the lines in this file back in and the integration tests pass now, but should these be here?

malandis (commented):
Looks good just one small change 👍

packages/client-sdk-nodejs/package-lock.json
anitarua: int tests failed again actually, and these don't seem to be necessary so going to remove them

anitarua:
Also re: tests getting cancelled after 4 minutes, it might be a bug with ubuntu 24 ([reference](https://github.com/actions/runner-images/issues))

malandis (approved):

packages/client-sdk-nodejs/package-lock.json
cprice404: you should never edit this file by hand. if it ever gets messed up, just delete it, and run `npm install`.

PR #1295, opened 2024-05-30
feat: clean up response type enum folders

bruuuuuuuce:
closes https://github.com/momentohq/dev-eco-issue-tracker/issues/842
closes https://github.com/momentohq/dev-eco-issue-tracker/issues/843

this pr moves the response enums into specific folders according to the
- service they are for (ie `cache`)
- data structure (ie `dictionary`)

Also completes the migration for all dictionary apis to the new pattern

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ concurrency:
 
 jobs:
   verify:
-    runs-on: macos-latest
+    runs-on: ubuntu-24.04
     steps:
       - name: Setup repo
         uses: actions/checkout@v3
@@ -327,7 +327,7 @@ jobs:
       matrix:
         node: [18]
     name: Test Cache on Node ${{ matrix.node }}
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
     env:
       MOMENTO_API_KEY: ${{ secrets.ALPHA_TEST_AUTH_TOKEN }}
       TEST_SESSION_TOKEN: ${{ secrets.MOMENTO_PREPROD_SESSION_TOKEN }}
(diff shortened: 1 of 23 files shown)


packages/core/src/messages/responses/cache-set.ts
```diff
@@ -1,16 +1,16 @@
 import {SdkError} from '../../errors';
 import {BaseResponseError, BaseResponseSuccess} from './response-base';
-import {MomentoResponse} from './enums';
+import {CacheSetResponse} from './enums';
 
 interface IResponse {
-  type: MomentoResponse;
+  type: CacheSetResponse;
```
bruuuuuuuce, line 6: We had a slack discussion about this. Having a generic `MomentoResponse` for apis that return `Success` or `Error`, but having api specific enums for apis that return `Hit` `Miss` or `Error` was not consistent. We are giving each api their own unique enum, as that gives us the most flexibility moving forward. If we need to add a third return type to the `Set` api in the future, we just need to change this one enum, and not worry about affecting other apis

cprice404: what you described above is the correct approach, ty

.github/workflows/build.yml
```diff
@@ -10,7 +10,7 @@ concurrency:
 
 jobs:
   verify:
-    runs-on: ubuntu-24.04
+    runs-on: macos-latest
```
anitarua, line 13: no longer need this

cprice404:
rebased and fixed macos lines

anitarua (approved):
lgtm!

PR #1296, opened 2024-05-30
chore: run cache tests on macos-latest instead of ubuntu 24

anitarua:
Only the `Build` job gets cancelled after 4 minutes when using ubuntu 24, so trying to revert that one to macos-latest for now


bruuuuuuuce (approved):
:shipit:

PR #1304, opened 2024-06-04
feat: migrate scalar responses to new enum pattern

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/844

### DIFF
--- a/examples/nodejs/cache/advanced.ts
+++ b/examples/nodejs/cache/advanced.ts
@@ -96,8 +96,6 @@ async function setGetDeleteExample() {
     logger.info(`Error deleting cache key: ${deleteResponse.message()}`);
   } else if (deleteResponse instanceof CacheDelete.Success) {
     logger.info('Deleted key from cache');
-  } else {
-    throw new Error(`Unrecognized response: ${deleteResponse.toString()}`);
   }
 }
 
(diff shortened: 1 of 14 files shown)


examples/nodejs/cache/advanced.ts
```diff
@@ -96,8 +96,6 @@ async function setGetDeleteExample() {
     logger.info(`Error deleting cache key: ${deleteResponse.message()}`);
   } else if (deleteResponse instanceof CacheDelete.Success) {
     logger.info('Deleted key from cache');
-  } else {
```
bruuuuuuuce, line 99: :w00t:

packages/client-sdk-web/src/internal/cache-data-client.ts
```diff
@@ -431,7 +431,7 @@ export class CacheDataClient<
                 break;
               default:
                 resolve(
-                  new CacheGet.Error(
+                  new CacheSetIfNotExists.Error(
```
bruuuuuuuce, line 434: was this a bug in the old code?

anitarua: yup

bruuuuuuuce: :w00t:

PR #1306, opened 2024-06-04
chore: handle http endpoint url

cprice404:
chore: In cloudflare example, ensure endpoints starting with http:// are correctly prefixed with https://

Previously, endpoints starting with http:// were prefixed with https://, which could lead to incorrect URL formation. This commit adjusts the prefix process to handle these cases properly.


cprice404:
This is a dupe of #1290 , just based on a branch against the Momento org because our CI doesn't currently support PRs from forks.

PR #1308, opened 2024-06-04
feat: migrate list responses to new enum pattern

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/846

### DIFF
--- a/packages/common-integration-tests/src/list.ts
+++ b/packages/common-integration-tests/src/list.ts
@@ -597,7 +597,9 @@ export function runListTests(
         expectWithMessage(() => {
           expect(response).toBeInstanceOf(CacheListPopFront.Hit);
         }, `expected a HIT but got ${response.toString()}`);
-        expect((response as CacheListPopBack.Hit).valueString()).toEqual('foo');
+        expect((response as CacheListPopFront.Hit).valueString()).toEqual(
+          'foo'
+        );
       });
 
       it('should support accessing value for CacheListPopFront.Hit without instanceof check', async () => {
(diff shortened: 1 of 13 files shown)


packages/common-integration-tests/src/list.ts
```diff
@@ -619,7 +621,7 @@ export function runListTests(
 
         expect(popResponse.value()).toEqual('foo');
 
-        const hitResponse = popResponse as CacheListPopBack.Hit;
+        const hitResponse = popResponse as CacheListPopFront.Hit;
```
bruuuuuuuce, line 624: another one  💪

packages/core/src/messages/responses/cache-list-pop-front.ts
```diff
@@ -1,62 +1,45 @@
...
-export abstract class Response extends ResponseBase {
-  public value(): string | undefined {
-    if (this instanceof Hit) {
-      return (this as Hit).value();
-    }
-    return undefined;
-  }
+interface IResponse {
+  value(): string | undefined;
+  type: CacheListPopFrontResponse;
```
bruuuuuuuce, line 14: I just learned that you can make this value on the interface `readonly`.

```
interface IResponse {
  value(): string | undefined;
  readonly type: CacheListPopFrontResponse;
}
```

 I think we should do that, so we don't accidentally make it mutable inside of the class. I can work on changing the existing ones to be readonly.

anitarua: oh nice! yeah I can update the ones here to be readonly

bruuuuuuuce (approved):
:shipit:

PR #1313, opened 2024-06-05
feat: migrate item responses to new enum pattern

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/867

Also includes a bug fix in a KeysExist test (was expecting the wrong response type).
Also updates the setIf Stored/NotStored responses so they extend ResponseBase instead of BaseResponseSuccess for consistency with the update ttl responses which use Set/NotSet/Miss.

### DIFF
--- a/packages/common-integration-tests/src/keys-exist.ts
+++ b/packages/common-integration-tests/src/keys-exist.ts
@@ -103,7 +103,7 @@ export function runKeysExistTest(
         expect(response).toBeInstanceOf(CacheKeysExist.Success);
       }, `expected SUCCESS but got ${response.toString()}`);
 
-      const success = response as CacheKeyExists.Success;
+      const success = response as CacheKeysExist.Success;
       expect(success.exists()).toEqual([]);
     });
 
--- a/packages/core/src/messages/responses/cache-item-get-ttl.ts
+++ b/packages/core/src/messages/responses/cache-item-get-ttl.ts
@@ -7,6 +7,7 @@ import {
 import {CacheItemGetTtlResponse} from './enums';
 
 interface IResponse {
+  value(): number | undefined;
   readonly type: CacheItemGetTtlResponse;
 }
 
@@ -24,7 +25,15 @@ export class Hit extends ResponseBase implements IResponse {
 
   /**
    * Returns the remaining ttl in milliseconds for object stored at passed key.
-   * @returns string
+   * @returns number
+   */
+  public value(): number {
+    return this.ttlMillis;
+  }
+
+  /**
+   * Returns the remaining ttl in milliseconds for object stored at passed key.
+   * @returns number
    */
   public remainingTtlMillis(): number {
     return this.ttlMillis;
@@ -44,6 +53,10 @@ export class Miss extends BaseResponseMiss implements IResponse {
   constructor() {
     super();
   }
+
+  value(): undefined {
+    return undefined;
+  }
 }
 
 /**
@@ -61,6 +74,10 @@ export class Error extends BaseResponseError implements IResponse {
 
   constructor(_innerException: SdkError) {
     super(_innerException);
+  }
+
+  value(): undefined {
+    return undefined;
   }
 }
 
--- a/packages/core/src/messages/responses/cache-item-get-ttl.ts
+++ b/packages/core/src/messages/responses/cache-item-get-ttl.ts
@@ -1,35 +1,20 @@
 import {SdkError} from '../../errors';
 import {
+  BaseResponseError,
+  BaseResponseMiss,
   ResponseBase,
-  ResponseError,
-  ResponseHit,
-  ResponseMiss,
 } from './response-base';
+import {CacheItemGetTtlResponse} from './enums';
+
+interface IResponse {
+  readonly type: CacheItemGetTtlResponse;
+}
 
 /**
- * Parent response type for a item ttl request.  The
- * response object is resolved to a type-safe object of one of
- * the following subtypes:
- *
- * - {Hit}
- * - {Miss}
- * - {Error}
- *
- * `instanceof` type guards can be used to operate on the appropriate subtype.
- * @example
- * For example:
- * ```
- * if (response instanceof ItemGetTtl.Error) {
- *   // Handle error as appropriate.  The compiler will smart-cast `response` to type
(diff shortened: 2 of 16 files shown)


packages/core/src/messages/responses/cache-ttl-decrease.ts
```diff
@@ -58,4 +42,13 @@ class _Error extends Response {
...
+export class Error extends BaseResponseError implements IResponse {
+  constructor(_innerException: SdkError) {
+    super(_innerException);
+  }
+
+  readonly type: CacheDecreaseTtlResponse.Error =
+    CacheDecreaseTtlResponse.Error;
+}
+
+export type Response = Set | NotSet | Miss | Error;
```
anitarua, line 54: I saw that JS and Go implemented only Set/Miss/Error for the update/increase/decrease ttl responses, but in Rust we implemented NotSet as well because it was possible a precondition for increase/decrease ttl would fail ([Rust int test for reference](https://github.com/momentohq/client-sdk-rust/blob/main/sdk/tests/cache/ttl.rs#L198)), but not for update ttl.

Should we audit the SDKs for consistency on this front? Could be:

- Set/NotSet/Miss/Error for all 3 commands (update, increase, decrease ttl)
- Set/NotSet/Miss/Error for increase and decrease, but Set/Miss/Error for update
- Set/Miss/Error for all 3 commands

packages/core/src/messages/responses/cache-item-get-ttl.ts
```diff
@@ -1,35 +1,20 @@
...
+  BaseResponseMiss,
   ResponseBase,
-  ResponseError,
-  ResponseHit,
-  ResponseMiss,
 } from './response-base';
+import {CacheItemGetTtlResponse} from './enums';
+
+interface IResponse {
+  readonly type: CacheItemGetTtlResponse;
```
anitarua, line 10: ItemGetType and ItemGetTtl did not previously have a `value()` method so I omitted them here too, but should all responses that follow the Hit/Miss/Error pattern provide `value()`?

packages/common-integration-tests/src/keys-exist.ts
```diff
@@ -103,7 +103,7 @@ export function runKeysExistTest(
         expect(response).toBeInstanceOf(CacheKeysExist.Success);
       }, `expected SUCCESS but got ${response.toString()}`);
 
-      const success = response as CacheKeyExists.Success;
+      const success = response as CacheKeysExist.Success;
```
bruuuuuuuce, line 106: 💪

packages/core/src/messages/responses/cache-item-get-ttl.ts
bruuuuuuuce: I am actually not sure if there were only specific apis that had the `value()` method, or it was all `Hit/Miss/Error` apis. The way I had been handling this was only treating the apis that had these methods as ones that needed these methods, but I didn't look closely enough and see if there was in fact a pattern of which apis had these

cc @cprice404

packages/core/src/messages/responses/cache-set-if-absent-or-equal.ts
```diff
@@ -9,15 +9,15 @@ interface IResponse {
 /**
  * Indicates the new value was set because the key did not exist or the existing item was equal to the supplied `equal` value.
  */
-export class Stored extends BaseResponseSuccess implements IResponse {
+export class Stored extends ResponseBase implements IResponse {
```
bruuuuuuuce, line 12: just wondering what was the logic here, I could go both ways on this, cause techincally `Stored` and `NotStored` are successful api calls

packages/core/src/messages/responses/cache-ttl-decrease.ts
bruuuuuuuce: good call I would probably try and align as much as possible with the protos

packages/core/src/messages/responses/cache-set-if-absent-or-equal.ts
anitarua: The Hit classes also extend ResponseBase instead of BaseResponseSuccess, so I thought only the classes named Success should extend that one. 
Should they all go the other way instead and mostly extend BaseResponseSuccess?

bruuuuuuuce: I don't think it entirely matters honestly I think this `SuccessBase` class is mostly used for us to have `DRY` integration tests. I think that makes sense to me that if `Hit` doesnt extend `Success`, then `Stored` should also not extend `Success`

packages/core/src/messages/responses/cache-ttl-decrease.ts
cprice404: thinking about this, will reply with more thoughts soon

cprice404: I definitely think we should include `NotSet` in the enums where it is a possible response type; let's eliminate the `FailedPrecondition` option.

I'm on the fence about whether to include `NotSet` for the cases where it is not actually a possible return type. I think I would vote for *not* including it on those. We can always add it in the future, but technically we can't remove it.

cprice404: and yes we should create tickets for adding `NotSet` in any of the SDKs where it got skipped.

cprice404: i created a generic ticket for the missing `NotSet` in other SDKs and put it on the board

packages/core/src/messages/responses/cache-item-get-ttl.ts
cprice404: My intent was this:

Any APIs where the successful response type has some data on it that a user may be trying to access after the request, the base response type should have an `| undefined` accessor. This is the only way we can provide an API where users never need to do any pattern matching on the response type if that's how they choose to structure their code.

cprice404: so it'd be great to add any that we notice are missing.

cprice404 (commented):
left my thoughts; i'm good when matt is

bruuuuuuuce (approved):
:shipit:

PR #1314, opened 2024-06-05
feat: migrate control responses to new enum pattern

anitarua:


malandis (commented):
Looks great to me!

packages/core/src/messages/responses/create-cache.ts
```diff
@@ -1,39 +1,23 @@
...
+ * Indicates that the cache already exists.
  */
-export class Success extends ResponseSuccess(_Success) {}
-
-class _Error extends Response {
-  constructor(protected _innerException: SdkError) {
-    super();
-  }
+export class AlreadyExists extends BaseResponseSuccess implements IResponse {
+  readonly type: CreateCacheResponse.Success = CreateCacheResponse.Success;
```
bruuuuuuuce, line 20: should this be a separate type from `success`? This doesn't seem like a successful request?

anitarua: good call, I can change that to extend ResponseBase instead

bruuuuuuuce (approved):
:shipit:

PR #1302, opened 2024-06-03
feat: add eventbridge example project

rishtigupta:

PR #1316, opened 2024-06-06
chore: improve webapp + add script to deploy and run webapp

rishtigupta:
## PR Description:
- Improve webapp based on this [pr](https://github.com/momentohq/client-sdk-javascript/pull/1302) comments
    - JSON field names in the topic messages match the DDB table schema column names exactly
    - configure the TTL
    - simplify deploy / launch steps. Add a single config file at the root, and a shell script that will make the configs available to both infra and webapp, do the CDK deploy, and launch the webapp. 
    
## Relevant PRs:
- Addition of demo to docs - https://github.com/momentohq/public-dev-docs/pull/1022


cprice404:
- [ ] add a CLI script that they can run instead of the web app if they prefer; it can just do a topic subscribe, write to ddb, cache get, ddb delete, cache get. it's probably fine for it to live in the same dir as the web app so you can re-use the typescript project files
- [ ] Update the README to include information about the CLI script
- [ ] Add some of the information from the beginning of your app modal to the README so the users will know, after reading the README, what kinds of thing the app is going to do, even if they don't run it.
- [ ] templatize the README file
- [ ] add links to this new example in the relevant parent READMEs
- [ ] check in the dev docs and see if there is any place there where we should mention this and add links
- [ ] if the `.env` file doesn't exist, give a better error message. Direct them to the README so they will know what they need to put in the file.
- [ ] add a note in the README that AWS_SESSION_TOKEN is not required if you are using a long-lived token
- [ ] you need to add `set -e` and `set -x` to your script so people can see what's going on, and so it will fail if a command fails. maybe do it after you source the env file so that the creds don't get printed to the screen.
- [ ] i think you committed a package-lock.json with codeartifact urls again
- [ ] for some reason, even with set -e and set -x, with the bad package-lock.json file in infrastructure dir, the bash script wasn't exiting when the npm install command failed. this is bad because when it moves on to running the vite app, the vite UI erases all of the console output so i can't see the error message anymore. we should try to figure out how to make the script fail if the npm install fails.
- [ ] after much trial and error and googling i discovered that [bash set -e doesn't work with && commands](https://stackoverflow.com/a/53121496). We should remove all of the &&'s from the script.

rishtigupta:
The "Info Modal" is loaded on initial page load. I added it to the local storage so that it doesn't pop up on every page load. 
And hence, the Help button. 
Also, the "cache not exists" dialog box only appears after the intro modal has been closed by the user. I thought since it explains that the cache is necessary for this project, we should be displaying that first instead.

rishtigupta:
Relevant PR for adding the demo to the docs: https://github.com/momentohq/public-dev-docs/pull/1022

examples/nodejs/aws/eventbridge/README.md
```diff
@@ -10,43 +10,43 @@ The app can be used to create, update and delete items in a DynamoDB table and t
 - HTTP API endpoint the same region as Momento API Key. You can copy the endpoint from the console after creating the API Key or refer to the [Regions Section here in the documentation](https://docs.momentohq.com/topics/develop/api-reference/http-api#regions)
```
cprice404, line 10: you can delete this file now (assuming your README.template.md has basically the same content) - it will be generated in CI.

examples/nodejs/aws/eventbridge/README.template.md
```diff
@@ -0,0 +1,52 @@
+# Welcome to Momento <-> Eventbridge Integration Example Project
```
cprice404, line 1: you need to update the github workflow files for PR and push, to tell them about this template file that needs to be generated.

examples/nodejs/aws/eventbridge/deploy-and-run-cli.sh
```diff
@@ -0,0 +1,115 @@
...
+fi
+
+# Load environment variables from .env.development file
+set -a
+source .env
+set +a
+
+# Check if environment variables are loaded
+if [ -z "$MOMENTO_API_KEY" ] || [ -z "$AWS_ACCESS_KEY_ID" ] || [ -z "$AWS_SECRET_ACCESS_KEY" ] || [ -z "$AWS_REGION" ]; then
+    echo "Error: Environment variables not loaded from .env file."
```
cprice404, line 18: these error messages do not tell the user which env vars are missing or how to fix it.

examples/nodejs/aws/eventbridge/deploy-and-run-cli.sh
```diff
@@ -0,0 +1,115 @@
...
+    exit 1
+fi
+
+if [ -z "$AWS_ACCESS_KEY_ID" ] || [ -z "$AWS_SECRET_ACCESS_KEY" ] || [ -z "$AWS_REGION" ]; then
+    echo "Error: Environment variables not loaded from .env file."
```
cprice404, line 23: you are requiring AWS_REGION here (which is a good thing i think) but your README doesn't tell the user to put it in the env file.

cprice404: also this error message doesn't tell me how to fix my problem.

examples/nodejs/aws/eventbridge/deploy-and-run.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
```
cprice404, line 1: if you're going to name the other file `and-run-cli.sh` then this one should probably be `deploy-and-run-web-app.sh`.

examples/nodejs/aws/eventbridge/webapp/src/svgs/corner-left-down.tsx
```diff
@@ -0,0 +1,9 @@
+export const CornerLeftDown = () => {
+  return (
+    <svg className="h-8 w-8 mt-2 text-gray-500" width="24" height="24" viewBox="0 0 24 24" strokeWidth="2"
+         stroke="currentColor" fill="none" strokeLinecap="round" strokeLinejoin="round">
+      <path stroke="none" d="M0 0h24v24H0z"/>
+      <path d="M18 6h-6a3 3 0 0 0 -3 3v10l-5 -5m10 0l-5 5"/>
+    </svg>
+  );
+};
```
cprice404, line 9: do you think we could add two EventBridge boxes to our "arch diagram"? so like, the arrows that come out of the ddb box would go to EventBridge boxes, and then those would have arrows that go to the momento boxes?

i think this would all be horizontal real estate, no vertical, so hopefully it should work?

examples/nodejs/aws/eventbridge/README.template.md
```diff
@@ -0,0 +1,52 @@
...
+
```
cprice404, line 2: in this PR i still do not see any updates to the READMEs in the parent directories, to provide links to this new example.

examples/nodejs/aws/eventbridge/deploy-and-run-cli.sh
```diff
@@ -0,0 +1,115 @@
...
+TTL="300" # 5 minutes
+
+# Set Momento Cache and Topic Name
+CACHE_NAME="momento-eventbridge-cache"
+TOPIC_NAME="momento-eventbridge-topic"
+
+## Install Momento CLI
+brew tap momentohq/tap
+brew install momento-cli
+brew upgrade momento-cli
```
cprice404, line 53: FYI - I would probably advise to never put brew commands in a script that you are asking users to run :) it's pretty easy to completely break your brew install so you should let them deal with that on their own. not relevant here because of another comment i'm going to make about doing this script in TS though :)

examples/nodejs/aws/eventbridge/deploy-and-run-cli.sh
```diff
@@ -0,0 +1,115 @@
...
+dynamodb_output=$(aws dynamodb get-item \
+    --table-name "$TABLE_NAME" \
+    --region "$AWS_REGION" \
+    --key '{
+        "Location": {"S": "'"$LOCATION"'"}
+    }')
+
+# Get Item From Momento Cache
+echo "Getting item from Momento Cache: $CACHE_NAME"
+momento cache get --cache $CACHE_NAME $LOCATION
```
cprice404, line 88: This is cool that you were able to get all of this working in bash. we can keep the bash around as a third example if you like - but I think the most important thing we were advocating for for the CLI run was just a typescript script that did all the things.

So you would still have a bash launcher script that does the CDK deploy (which you can refactor out to a separate script if you want to keep things DRY, up to you), and then it would just call `npm run cli-demo` or something. And then that would call some typescript code that would do the topic subscription, ddb writes, momento cache gets, etc.

That will be more portable for users, if they want to copy that code and use it in their own app they can do it easily. Whereas the bash ddb/momento actions are cool and interesting but not something that a user would be likely to copy/paste into an app.

examples/nodejs/aws/eventbridge/webapp/src/App.tsx
```diff
@@ -82,93 +89,181 @@ const App = () => {
...
+                  <Tippy content={"Unsubscribed"} placement={"top"}>
+                    <div><BanCircle/></div>
+                  </Tippy>
+                )}
+              </div>
+            </div>
+            <div className="bg-gray-100 p-2 rounded-lg text-sm">
+                <pre>
+                  <code className={"whitespace-pre-wrap"}>
+                    {`topicClient.subscribe(${cacheName}, ${topicName}, {onItem: onItemCallback})`}
```
cprice404, line 232: you need to put quotes around the cache/topic name, here and in the GetItem one.

examples/nodejs/aws/eventbridge/webapp/src/vite-env.d.ts
```diff
@@ -5,6 +5,7 @@ interface ImportMetaEnv {
   readonly VITE_AWS_SECRET_ACCESS_KEY: string;
   readonly VITE_AWS_SESSION_TOKEN: string;
   readonly VITE_MOMENTO_API_KEY: string;
+  readonly VITE_AWS_REGION: string;
 }
 
 interface ImportMeta {
```
cprice404, line 11: The changes you made to give us more vertical space are GREAT. nice work.
The modals are working great for me now too. Even when I deleted my cache etc.

new checklist just to try to make it easier to keep track of this round of feedback:
- [ ] delete readme (it will get generated from template)
- [ ] update github workflows for new readme template
- [ ] add links to this example in the relevant READMEs in parent dirs in examples tree
- [ ] add AWS_REGION to example .env section in README
- [ ] improve error messages for missing env vars
- [ ] port CLI demo to typescript
- [ ] rename -deploy-and-run to -deploy-and-run-web-app?
- [ ] add EventBridge boxes to arch diagram?

I think that's probably it! After that we can merge and then I will make one more quick pass to suggest text tweaks in a few places and we should be done done!

.github/workflows/on-push-to-main-branch.yml
```diff
@@ -204,6 +204,33 @@ jobs:
           template_file: ./examples/nodejs/aws/secrets-manager/README.template.md
           output_file: ./examples/nodejs/aws/secrets-manager/README.md
 
+      - name: Generate README for examples/nodejs/aws/eventbridge
```
cprice404, line 207: you need to do this in the PR workflow file too

rishtigupta: I forgot, will do. Thanks!

examples/nodejs/aws/eventbridge/run-cli.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
```
cprice404, line 1: can we rename this file to `run-bash-cli-app.sh`, and the other one to `run-typescript-cli-app.sh`?

cprice404 (commented):
This looks great! just two small suggestions and then once CI is green I think we can merge!

cprice404 (approved):
dance

PR #1317, opened 2024-06-06
fix: correct convertToB64String function to correctly handle internat…

rishtigupta:
…ional characters

## Relevant Ticket:
https://github.com/momentohq/dev-eco-issue-tracker/issues/829

## Explanation:
**For Strings:**
- TextEncoder().encode(v) converts the string to a Uint8Array of UTF-8 bytes.
- String.fromCharCode(...utf8Bytes) converts the Uint8Array to a binary string.
- btoa(binaryString) converts the binary string to a base64-encoded string.

**For Uint8Array:**
- String.fromCharCode(...v) converts the Uint8Array directly to a binary string.
- btoa(binaryString) converts the binary string to a base64-encoded string.

## Testing:
Tested using the basic example in web directory. result:
```
npm run example                                                                                 

> momento-nodejs-example@1.0.0 example
> tsc && node dist/basic.js

cache already exists
Storing key=foo, value=çCMduo
Key stored successfully!
cache hit: çCMduo
success!!
```

### DIFF
--- a/packages/client-sdk-web/src/utils/web-client-utils.ts
+++ b/packages/client-sdk-web/src/utils/web-client-utils.ts
@@ -11,6 +11,23 @@ export function convertToB64String(v: string | Uint8Array): string {
   }
 }
 
+export function convertToBytesFromB64String(b64: string): Uint8Array {
+  const binaryString = atob(b64);
+  const binaryLength = binaryString.length;
+  const bytes = new Uint8Array(binaryLength);
+
+  for (let i = 0; i < binaryLength; i++) {
+    bytes[i] = binaryString.charCodeAt(i);
+  }
+
+  return bytes;
+}
+
+export function convertToStringFromB64String(b64: string): string {
+  const bytes = convertToBytesFromB64String(b64);
+  return new TextDecoder().decode(bytes);
+}
+
 export function createCallMetadata(
   cacheName: string,
   timeoutMillis: number
--- a/packages/client-sdk-web/src/utils/web-client-utils.ts
+++ b/packages/client-sdk-web/src/utils/web-client-utils.ts
@@ -2,11 +2,26 @@ import {CredentialProvider} from '@gomomento/sdk-core';
 
 export function convertToB64String(v: string | Uint8Array): string {
   if (typeof v === 'string') {
-    return btoa(v);
+    const encodedString = encodeURIComponent(v);
+    let decodedString = '';
+    for (let i = 0; i < encodedString.length; i++) {
+      if (encodedString[i] === '%') {
+        const hex = encodedString.substr(i + 1, 2);
+        decodedString += String.fromCharCode(parseInt(hex, 16));
+        i += 2;
+      } else {
+        decodedString += encodedString[i];
+      }
+    }
+    return btoa(decodedString);
+  } else {
+    const chars = new Uint8Array(v);
+    let binary = '';
+    for (let i = 0; i < chars.length; i++) {
+      binary += String.fromCharCode(chars[i]);
+    }
+    return btoa(binary);
   }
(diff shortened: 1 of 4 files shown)


bruuuuuuuce (dismissed):
:shipit:

packages/client-sdk-web/src/utils/web-client-utils.ts
```diff
@@ -2,11 +2,26 @@ import {CredentialProvider} from '@gomomento/sdk-core';
 
 export function convertToB64String(v: string | Uint8Array): string {
   if (typeof v === 'string') {
```
cprice404, line 4: I think it would be a lot easier to reason about this and to make sure we have confidence in it if we added some unit tests.

Also, are there places where this is called where we genuinely don't know whether `v` is a `string` or `Uint8Array`? it seems to me like the logic for those two cases is so wildly different that it'd be better to have two separate functions, and I'd kinda hope that at the call sites for this that we would know which one we were dealing with but maybe not if this is getting called for a lot of cache values when we are about to send the data up to the server.

packages/client-sdk-web/test/unit/utils/web-client-utils.test.ts
```diff
@@ -136,4 +137,32 @@ describe('getWeb*Endpoint', () => {
...
+    it('should convert a Uint8Array to base64', () => {
+      const input = new Uint8Array([104, 101, 108, 108, 111]);
+      const expectedOutput = 'aGVsbG8=';
+      expect(convertToB64String(input)).toBe(expectedOutput);
+    });
+    it('should convert a Uint8Array with special characters to base64', () => {
+      const input = new Uint8Array([104, 195, 169, 108, 108, 195, 182]);
+      const expectedOutput = 'aMOpbGzDtg==';
+      expect(convertToB64String(input)).toBe(expectedOutput);
+    });
```
cprice404, line 166: yay thanks for adding these! do you think it would be valuable to add tests to round-trip them back to the original values?

rishtigupta: Sure, will add the tests

cprice404: thanks! as a sanity check you could also comment out your new implementation and change it back to the old one and hopefully the round-trip tests would fail. that would be a nice confidence booster for this change :)

packages/client-sdk-web/src/utils/web-client-utils.ts
```diff
@@ -2,11 +2,13 @@ import {CredentialProvider} from '@gomomento/sdk-core';
...
-    return btoa(v);
+    const utf8Bytes = new TextEncoder().encode(v);
+    const binaryString = String.fromCharCode(...utf8Bytes);
+    return btoa(binaryString);
```
cprice404, line 7: this looks a lot cleaner/safer than the previous versions. i'm okay with this unless you find something better.

packages/client-sdk-web/test/unit/utils/web-client-utils.test.ts
cprice404: @rishtigupta thanks for adding the integration tests! I think those are valuable. When I said roundtrip here though what I meant was, is there a way in these unit tests, to convert the base64 back to the original input format and verify that we got back the same value we started with? Like, do we have another function that is the opposite of `convertToBase64String` that we call when we are going the other direction? If so can we just update each of the tests above to call that on the base64 data and add an assertion that the result equals the original input?

that's my last comment on this PR, i'm +1 once that's done. and if we don't have a function like that that it makes sense to add to the tests, we can punt and merge this

cprice404 (commented):
just left one more minor question/comment

cprice404 (approved):
dance

PR #1334, opened 2024-06-17
feat: make sure Agent and Runtime-Version headers are sent once

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/888

All the clients that send headers:

- cache
- topic
- leaderboard
- webhook
- auth
- ping

### DIFF
--- a/packages/client-sdk-web/src/internal/client-metadata-provider.ts
+++ b/packages/client-sdk-web/src/internal/client-metadata-provider.ts
@@ -4,6 +4,7 @@ import {version} from '../../package.json';
 export interface ClientMetadataProps {
   authToken?: string;
   readConcern?: ReadConcern;
+  clientType?: string;
 }
 
 /**
@@ -12,11 +13,13 @@ export interface ClientMetadataProps {
 export class ClientMetadataProvider {
   private static readonly agentName: string = 'js-web';
   private readonly authToken?: string;
-  private isAgentSent = false;
+  private readonly clientType: string;
+  private areFirstTimeHeadersSent = false;
   private readConcern?: ReadConcern;
   constructor(props: ClientMetadataProps) {
     this.authToken = props.authToken;
     this.readConcern = props.readConcern;
+    this.clientType = props.clientType ?? 'unknown';
   }
 
   /**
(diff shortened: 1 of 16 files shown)


packages/client-sdk-web/src/internal/client-metadata-provider.ts
```diff
@@ -12,11 +13,13 @@ export interface ClientMetadataProps {
 export class ClientMetadataProvider {
```
anitarua, line 13: As of #528, the Web SDK uses `ClientMetadataProvider` to send headers because the interceptor didn't work.

However, this means the Web SDK headers are sent once per new client (e.g. cache, topic, auth, etc) rather than once at the start of any client usage, like with the NodeJS SDK.

packages/client-sdk-web/src/internal/client-metadata-provider.ts
```diff
@@ -28,13 +31,50 @@ export class ClientMetadataProvider {
...
     }
     if (this.readConcern) {
       metadata['read-concern'] = this.readConcern;
     }
     return metadata;
   }
 }
+
+// Each browser uses its own JS engine, the best we can do is detect the browser name.
+// Source: https://developer.mozilla.org/en-US/docs/Web/API/Window/navigator
```
anitarua, line 52: Web SDK's "runtime version" -> I interpreted this as browser JS rendering engine version.

Mozilla docs had this function for returning the browser name as a string based on the information found in the `navigator.userAgent` field. If there's anything else we want to add to the header, we can add it

packages/client-sdk-web/src/internal/client-metadata-provider.ts
```diff
@@ -4,6 +4,7 @@ import {version} from '../../package.json';
 export interface ClientMetadataProps {
   authToken?: string;
   readConcern?: ReadConcern;
+  clientType?: string;
```
malandis, line 7: Should be this required? I think this is known in all the call sites.

malandis (commented):
Looks good. Just one question about the web client interface.

packages/client-sdk-web/src/internal/client-metadata-provider.ts
malandis: I agree this makes sense. As for the below `getBrowserName`, I see we attempt to classify the `navigator.userAgent` content and normalize it. In the event we cannot, we return `unknown-browser`.

I feel like as long as `navigator.userAgent` is not an empty string, we should return it as-is, since this would help with metrics tracking on our end. If we normalize all other browsers to `unknown-browser`, then we're in the dark as to what it was or if it was a naming variant we were unaware of.

malandis (commented):
Regarding browser classification

packages/client-sdk-web/src/internal/client-metadata-provider.ts
anitarua: I'm worried that it would bloat the header too much because `userAgent` can return a really long string. Examples:
```
"Mozilla/5.0 (iPhone; CPU iPhone OS 15_6_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/15.6 Mobile/15E148 Safari/604.1"

"Mozilla/5.0 (Linux; Android 9; SAMSUNG SM-G955F Build/PPR1.180610.011) AppleWebKit/537.36 (KHTML, like Gecko) SamsungBrowser/9.4 Chrome/67.0.3396.87 Mobile Safari/537.36"

"Mozilla/5.0 (Macintosh; Intel Mac OS X 12_5_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.0.0 Safari/537.36 OPR/90.0.4480.54"
```

We could parse out the more useful parts to us

malandis: I see. There are third-party solutions we can use to structure this such as [this](https://github.com/faisalman/ua-parser-js). We can leave that for future work.

malandis (approved):

packages/client-sdk-web/src/internal/client-metadata-provider.ts
anitarua: ooh nice, yeah I can make a ticket for improving this

PR #1337, opened 2024-06-18
chore: remove express readconcern

cprice404:


rishtigupta:
The PR looks good to me. Could you please share the motivation behind removing the "Express" read-concern?

rishtigupta (approved):

cprice404:
> The PR looks good to me. Could you please share the motivation behind removing the "Express" read-concern?

yep, will start a slack thread

PR #1339, opened 2024-06-20
WIP: sketch changes to handle not found and introduce value type

cprice404:


examples/nodejs/cache/basic.ts
```diff
@@ -1,36 +1,133 @@
-import {CacheGet, CreateCache, CacheSet, CacheClient, Configurations, CredentialProvider} from '@gomomento/sdk';
```
cprice404, line 1: i just included this file so you can see how i was testing things and playing with the API. these changes won't go into the final PR.

examples/nodejs/cache/basic.ts
```diff
@@ -1,36 +1,133 @@
...
+  console.log(`PutResponse: ${putResponse}`);
+  const getResponse = await storageClient.get(storeName, 'numberKey');
+  // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+  console.log(`GetResponse: ${getResponse}`);
+  const numberVal = getResponse.value();
+  // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+  console.log(`Number value: ${numberVal}`);
+  console.log(`Type of number value: ${typeof numberVal}`);
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  console.log(`YOLO happy path number value: ${getResponse.value()!.int()!}`);
```
cprice404, line 99: This is what I'm really trying to get us to.

examples/nodejs/cache/basic.ts
```diff
@@ -1,36 +1,133 @@
...
-  if (getResponse instanceof CacheGet.Hit) {
-    console.log(`cache hit: ${getResponse.valueString()}`);
-  } else if (getResponse instanceof CacheGet.Miss) {
-    console.log('cache miss');
-  } else if (getResponse instanceof CacheGet.Error) {
-    console.log(`Error: ${getResponse.message()}`);
+  const notFoundGetResponse = await storageClient.get(storeName, 'notFoundKey');
+  // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+  console.log(`NotFoundGetResponse: ${notFoundGetResponse}`);
+  const notFoundNumber = notFoundGetResponse.value();
```
cprice404, line 119: and this is what i was trying to get us to in terms of being able to clearly identify a "not found"

examples/nodejs/cache/package.json
```diff
@@ -33,7 +33,8 @@
     "typescript": "4.9.5"
   },
   "dependencies": {
-    "@gomomento/sdk": "^1.89.1"
+    "@aws-sdk/client-dynamodb": "^3.600.0",
+    "@gomomento/sdk": "file:../../../packages/client-sdk-nodejs"
```
cprice404, line 37: don't include this in final pr either

packages/client-sdk-nodejs/src/errors/cache-service-error-mapper.ts
```diff
@@ -76,8 +80,19 @@ export class CacheServiceErrorMapper
         return new UnknownServiceError(...errParams);
       case Status.UNAVAILABLE:
         return new ServerUnavailableError(...errParams);
-      case Status.NOT_FOUND:
-        return new NotFoundError(...errParams);
+      case Status.NOT_FOUND: {
```
cprice404, line 83: this is a key thing in my changes - after playing with the API and discussion with some other folks we are feeling like we don't want users to have to deal with the full error handling patterns in order to tell when an item just did not exist. But the server is returning it as an error. So this code is what we need in order to disambiguate between different types of NOT FOUND.

packages/client-sdk-nodejs/src/internal/storage-data-client.ts
```diff
@@ -189,10 +198,10 @@ export class StorageDataClient implements IStorageDataClient {
     });
   }
 
-  public async put(
```
cprice404, line 192: `put` is not sufficient in javascript because when a user passes us a `number` we can't tell for sure whether they intended to store it as a double or as an integer.

packages/client-sdk-nodejs/test/unit/cache-service-error-mapper.test.ts
```diff
@@ -92,15 +92,18 @@ describe('CacheServiceErrorMapper', () => {
       expect(resolved).toBeInstanceOf(BadRequestError);
     });
   });
-  it('should return cache not found error when grpc status is NOT_FOUND', () => {
+  // TODO: NEED TO FIX THIS AND COVER MORE CASES
+  it.skip('should return cache not found error when grpc status is NOT_FOUND', () => {
```
cprice404, line 96: i added a skip on this test; we don't need to fix this before the initial PR gets merged (to unblock bug bash) but this test needs to be extended and re-enabled in a fast follow.

packages/core/src/messages/responses/storage/scalar/storage-get.ts
```diff
@@ -1,142 +1,159 @@
...
-
-  valueInt(): undefined {
-    return undefined;
-  }
-
-  valueString(): undefined {
-    return undefined;
-  }
-}
+// export class StringResponse extends Success {
```
cprice404, line 42: pulling this stuff out into a separate StorageValue class makes the logic simpler, and makes it possible / much easier to distinguish between NOT FOUND and "you chose the wrong type"

PR #1342, opened 2024-06-22
chore: port initial batch of example code to use `switch` style

cprice404:
This commit ports the first handful of API examples over to use
the new switch-based pattern matching style. It also updates
the API reference docs for `get` and `getBatch` to show both
the "simplified" and "pattern matching" styles of accessing
the retrieved data.

This is intended to serve as a starting point for discussion
about whether we like this approach for the docs. I built the docs
locally and looked at these examples, and I like them. If other
folks on the team are on board with this approach then we can
move forward quickly on porting the rest of the docs/examples
over.

### DIFF
--- a/examples/nodejs/cache/.eslintrc.json
+++ b/examples/nodejs/cache/.eslintrc.json
@@ -51,7 +51,8 @@
     // the intent of the developer. Functions are async when they want to await.
     "require-await": "error",
     "import/no-duplicates": "error",
-    "@typescript-eslint/switch-exhaustiveness-check": "error"
+    "@typescript-eslint/switch-exhaustiveness-check": "error",
+    "@typescript-eslint/no-non-null-assertion": "off"
   },
   "settings": {
     "import/resolver": {
--- a/packages/core/src/messages/responses/response-base.ts
+++ b/packages/core/src/messages/responses/response-base.ts
@@ -36,7 +36,7 @@ export interface IResponseSubscription {
 }
 
 export abstract class BaseResponseError extends ResponseBase {
-  public _innerException: SdkError;
+  private readonly _innerException: SdkError;
 
   protected constructor(innerException: SdkError) {
     super();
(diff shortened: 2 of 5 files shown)


rishtigupta (approved):
:shipit:

packages/core/src/messages/responses/response-base.ts
```diff
@@ -36,7 +36,7 @@ export interface IResponseSubscription {
 }
 
 export abstract class BaseResponseError extends ResponseBase {
-  public _innerException: SdkError;
+  private readonly _innerException: SdkError;
```
rishtigupta, line 39: The CI fails on making `inner_exception` field private. It might be getting accessed somewhere in the sdk/examples?

packages/core/src/messages/responses/response-base.ts
```diff
@@ -36,7 +36,7 @@ export interface IResponseSubscription {
 }
 
 export abstract class BaseResponseError extends ResponseBase {
-  public _innerException: SdkError;
+  protected readonly _innerException: SdkError;
```
malandis, line 39: Did you mean for this change to be in the PR?

examples/nodejs/cache/.eslintrc.json
```diff
@@ -51,7 +51,8 @@
     // the intent of the developer. Functions are async when they want to await.
     "require-await": "error",
     "import/no-duplicates": "error",
-    "@typescript-eslint/switch-exhaustiveness-check": "error"
+    "@typescript-eslint/switch-exhaustiveness-check": "error",
+    "@typescript-eslint/no-non-null-assertion": "off"
```
malandis, line 55: Will we include these details in the example README?

malandis (approved):
LGTM, no blockers, couple questions

examples/nodejs/cache/.eslintrc.json
cprice404: we definitely should, i will add some notes in the tickets

packages/core/src/messages/responses/response-base.ts
cprice404: yeah

PR #1343, opened 2024-06-24
fix: get element not found errors from grpc metadata

bruuuuuuuce:


packages/client-sdk-web/src/errors/cache-service-error-mapper.ts
```diff
@@ -118,16 +106,7 @@ export class CacheServiceErrorMapper
...
-        // TODO: Remove this once the error message is standardized on the server side
-        const errorMessage = errParams[0]?.toString();
-        const isStoreAlreadyExists =
-          errorMessage?.includes('Store with name:') &&
-          errorMessage?.includes('already exists');
-        // If errCause is not already set to 'store_already_exists', check for store_already_exists error
-        if (!errCause && isStoreAlreadyExists) {
-          errCause = 'store_already_exists';
-        }
+        const errCause = errParams[2]?.['err']?.[0];
```
rishtigupta, line 109: Can we confirm from Joseph if he added the change for "store_already_exists" error key too?

PR #1352, opened 2024-06-26
fix: remove `TEST_CACHE_NAME` env var reference in tests

malandis:
Previoulsy we read the test cache name from the environment in CI. We
no longer do this. However a developer may have `TEST_CACHE_NAME`
set.

In that case the integration tests will always use that for the cache
name. Because of this, control tests fail since some invoke
`testCacheName()` expecting a unique cache name as opposed to a
consistent one.

Speaking with other developers, they have been hit by this gotcha. One
solution would be to change tests to specify if they expect a unique
cache name vs some cache name. That will make the tests more difficult
to maintain, so we will remove the env var instead.

### DIFF
--- a/packages/client-sdk-nodejs-compression-zstd/test/integration/integration-setup.ts
+++ b/packages/client-sdk-nodejs-compression-zstd/test/integration/integration-setup.ts
@@ -26,9 +26,7 @@ export interface CacheClientPropsWithConfig extends CacheClientProps {
 }
 
 function testCacheName(): string {
-  return (
-    process.env.TEST_CACHE_NAME || `js-integration-test-compression-${v4()}`
-  );
+  return `js-integration-test-compression-${v4()}`;
 }
 
 let _credsProvider: CredentialProvider | undefined = undefined;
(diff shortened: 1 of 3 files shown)


bruuuuuuuce:
can we also remove it 

[here](https://github.com/momentohq/client-sdk-javascript/blob/b60b2da8e66d6ad1371c8d992eee74d3c1a42c74/packages/client-sdk-nodejs-compression-zstd/test/integration/integration-setup.ts#L30) and [here](https://github.com/momentohq/client-sdk-javascript/blob/b60b2da8e66d6ad1371c8d992eee74d3c1a42c74/packages/client-sdk-nodejs-compression/test/integration/integration-setup.ts#L30)

malandis:
> can we also remove it
> 
> [here](https://github.com/momentohq/client-sdk-javascript/blob/b60b2da8e66d6ad1371c8d992eee74d3c1a42c74/packages/client-sdk-nodejs-compression-zstd/test/integration/integration-setup.ts#L30) and [here](https://github.com/momentohq/client-sdk-javascript/blob/b60b2da8e66d6ad1371c8d992eee74d3c1a42c74/packages/client-sdk-nodejs-compression/test/integration/integration-setup.ts#L30)

Removed in this commit: [43e14fc](https://github.com/momentohq/client-sdk-javascript/pull/1352/commits/43e14fc39af0244c9292eb8d34c7e5fee05f9785)

bruuuuuuuce (approved):
:shipit:

Issue #1341, opened 2024-06-22
Neither setContrainsElement nor setContrainsElements exist in the sdk


dobeerman:
Just double-checked the SDK. The following methods are mentioned in the documentation but are not actually implemented.

1. Property `setContainsElement` does not exist on type`CacheClient`
2. Property `setContainsElements` does not exist on type `CacheClient`

Am I missed something? :thinking:

dobeerman:
omg, I've miossed the matrix https://docs.momentohq.com/cache/develop/api-reference/language-support

Any plans to implement it? 
Anyway, we can do it by ourselves :wink:

malandis:
@dobeerman we have added `setContainsElement` and `setContainsElements` per your request. Please try it out in v1.92.0. Let me know if you have any other concerns or questions.

PR #1357, opened 2024-06-26
feat: add set contains element and set contains elements

malandis:
Adds the `setContainsElement` command and `setContainsElements` command.

Exposes the result of a `setContainsElements` as a record from the element name to a boolean flag indicating presence/absence.

Addresses #1341

### DIFF
--- a/packages/core/src/messages/responses/cache-set-contains-elements.ts
+++ b/packages/core/src/messages/responses/cache-set-contains-elements.ts
@@ -1,0 +1,97 @@
+import {SdkError} from '../../errors';
+import {
+  BaseResponseError,
+  BaseResponseMiss,
+  ResponseBase,
+} from './response-base';
+import {truncateString} from '../../internal/utils';
+import {CacheSetContainsElementsResponse} from './enums';
+
+const TEXT_DECODER = new TextDecoder();
+
+interface IResponse {
+  contains(): Record<string, boolean> | undefined;
+  readonly type: CacheSetContainsElementsResponse;
+}
+
+/**
+ * Indicates that the requested data was successfully retrieved from the cache.  Provides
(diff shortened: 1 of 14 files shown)


packages/core/src/messages/responses/cache-set-contains-elements.ts
```diff
@@ -0,0 +1,103 @@
...
+
+const TEXT_DECODER = new TextDecoder();
+
+interface IResponse {
+  /**
+   * Returns a mapping of the elements to their presence in the cache.
+   * @returns {Record<string, boolean> | undefined} A mapping of the elements to their presence in the cache.
+   * If the set itself was not found, (ie the response was a `Miss` or `Error`), this method returns `undefined`.
+   */
+  containsElements(): Record<string, boolean> | undefined;
```
bruuuuuuuce, line 18: does record not support `Record<string | Uint8Array, boolean>`? If the user wants to find if a specific binary element was a hit, they will have to know to use a textdecoder to decode the binary element first and then do a lookup

malandis: I modelled this off of `dictionaryGetFields` (see [here](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/core/src/messages/responses/cache-dictionary-get-fields.ts)).

There are similar concerns: a user can request various fields from a dictionary. We give the response back as a record for convenience. The fields can be string vs Uint8Array and so can the values. We have accessors there for each combination, but we assume the most common request is for string-string.

Here I am assuming the most common case is the set contains strings. We can add other accessors if users request. Aiming to keep the initial API slim.

bruuuuuuuce (approved):
:shipit:

PR #1364, opened 2024-06-27
fix: lowercase runtime-version and agent headers

bruuuuuuuce:


malandis:
I would search globally for `runtime-version` insensitive, as they are also in the "headers only sent once" variables etc.

rishtigupta (approved):
📦

PR #1378, opened 2024-06-28
test: support testing aganst momento local by env var presence

malandis:
To simplify testing against momento local, we use the momento local
credential provider when the `MOMENTO_LOCAL` environment variable
exists.

### DIFF
--- a/packages/client-sdk-nodejs/test/integration/integration-setup.ts
+++ b/packages/client-sdk-nodejs/test/integration/integration-setup.ts
@@ -86,11 +86,20 @@ function sessionCredsProvider(): CredentialProvider {
   return _sessionCredsProvider;
 }
 
+function testAgainstMomentoLocal(): boolean {
+  return process.env.MOMENTO_LOCAL !== undefined;
+}
+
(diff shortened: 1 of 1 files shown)


packages/client-sdk-nodejs/test/integration/integration-setup.ts
```diff
@@ -86,11 +86,20 @@ function sessionCredsProvider(): CredentialProvider {
   return _sessionCredsProvider;
 }
 
+function testAgainstMomentoLocal(): boolean {
+  return process.env.MOMENTO_LOCAL !== undefined;
+}
+
 export function integrationTestCacheClientProps(): CacheClientPropsWithConfig {
+  let credentialProvider = credsProvider();
```
bruuuuuuuce, line 94: do we also want this in the web client?

malandis: Good question! At the present time we're only using the node client for testing against momento local. In the future we may broaden this configuration feature to all SDKs. For now I'd like to keep it scoped to only what we're using.

bruuuuuuuce (approved):
:shipit:

PR #1381, opened 2024-07-03
fix: storage get response should use Found and NotFound cases

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/935

Based on written notes, I don't believe there's anything else to address in the JS storage clients, but please let me know if there are other issues.

This PR replaces `StorageGetResponse.Success` with `Found` (always returns a `StorageValue`) and `NotFound` (always returns `undefined`).

```
const getResponse = await client.get(storeName, 'key');

// quick check
console.log('what is the value?', getResponse.value()?.string());

// full switch check
switch (getResponse.type) {
    case StorageGetResponse.Found:
      console.log(`Got string value: ${getResponse.value().string() ?? 'not a string'}`);
      console.log(`Got string value: ${getResponse.value().int() ?? 'not an int'}`);
      break;
    case StorageGetResponse.NotFound: {
      console.log('Key not found: hello');
      break;
    }
    case StorageGetResponse.Error:
      console.error('Failed to get string value: world');
      break;
  }
```

When locally testing, VSCode pointed out incomplete switch statements when `"@typescript-eslint/switch-exhaustiveness-check": "error"` was specified in the `eslintrc.json` file.

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/storage-data-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/storage-data-client.ts
@@ -162,22 +162,22 @@ export class StorageDataClient implements IStorageDataClient {
             switch (value) {
               case 'double_value': {
                 return resolve(
-                  StorageGet.Success.ofDouble(resp.value.double_value)
+                  StorageGet.Found.ofDouble(resp.value.double_value)
                 );
               }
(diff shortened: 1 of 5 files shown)


cprice404 (approved):
high five

PR #1384, opened 2024-07-09
chore: add example docs snippets for PreviewStorageClient

cprice404:
This commit adds a skeleton example project for storage examples,
and the initial docs example snippets that will be used in the
public dev docs website.

Also cleans up some example code using the older more verbose
style of specifying a credential provider via env var.


anitarua (approved):
lgtm!

PR #1386, opened 2024-07-09
feat: add cacheclient.create function in web sdk

rishtigupta:
https://github.com/momentohq/dev-eco-issue-tracker/issues/930

### DIFF
--- a/packages/client-sdk-web/src/cache-client.ts
+++ b/packages/client-sdk-web/src/cache-client.ts
@@ -51,6 +51,10 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
   public get configuration(): Configuration {
     return this._configuration;
   }
+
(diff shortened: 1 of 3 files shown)


packages/client-sdk-web/src/cache-client.ts
```diff
@@ -51,6 +51,10 @@ export class CacheClient extends AbstractCacheClient implements ICacheClient {
   public get configuration(): Configuration {
     return this._configuration;
   }
+
+  public static create(cacheClientProps: CacheClientProps): CacheClient {
```
cprice404, line 55: the example code that we are creating this for always calls `await` on this, so I think the function will need to be `async`.

I know the linter will complain about this, so you will probably need to add one of those `eslint-disable` pragmas, and a comment explaining why.

cprice404 (commented):
one minor change, thanks for doing this

cprice404 (approved):

PR #1387, opened 2024-07-09
fix: nodejs first-time headers should be sent per client

anitarua:
Each client creates an instance of `class HeaderInterceptorProvider` which is responsible for determining whether to send the first-request headers.

To ensure each client sends them on their first request, I switched the `areOnlyOnceHeadersSent` from a `static` variable to a non-static one.


bruuuuuuuce (approved):
:shipit: nice find!

PR #1388, opened 2024-07-09
fix: env var cred provider syntax in docs examples

cprice404:


examples/nodejs/cache/doc-example-files/doc-examples-js-apis.ts
```diff
@@ -111,7 +111,7 @@ function retrieveApiKeyFromYourSecretsManager(): string {
 }
 
 function example_API_CredentialProviderFromEnvVar() {
-  CredentialProvider.fromEnvironmentVariable({environmentVariableName: 'MOMENTO_API_KEY'});
+  CredentialProvider.fromEnvironmentVariable('MOMENTO_API_KEY');
 }
 
 function example_API_CredentialProviderFromString() {
```
anitarua, line 117: does this new syntax apply to FromString too?

anitarua (approved):

examples/nodejs/cache/doc-example-files/doc-examples-js-apis.ts
cprice404: it's available there, yeah, but I don't feel strongly about going through and fixing all of those proactively since it's a little less verbose.

PR #1396, opened 2024-07-25
feat: add retry strategy to storage client

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/878

Prototyping implementation of storage client retry strategy.
Uses a new interceptor to create timeouts and retry every 1 second that data is not received, up to a 5s overall timeout. 

Added some unit tests against the new storage retry strategy class and the changes to the storage client config. Also manually tested against [stubbed storage service](https://github.com/momentohq/momento-local/pull/89/files) and verified that it will end after 5 seconds with a CANCELLED error if the server doesn't return anything.


cprice404:
> Also manually tested against stubbed storage service and verified that it will end after 5 seconds with a CANCELLED error if the server doesn't return anything.

Out of curiosity, how did you do this? It might be a really valuable to document / share with the team.

packages/client-sdk-nodejs/src/config/transport/storage/transport-strategy.ts
```diff
@@ -28,13 +34,26 @@ export interface StorageTransportStrategy {
...
 
 export interface StorageTransportStrategyProps {
   /**
    * low-level gRPC settings for communication with the Momento server
    */
   grpcConfiguration: StorageGrpcConfiguration;
+  /**
+   * time to wait for response data to be received before retrying (defaults to 1s)
+   */
+  responseDataReceivedTimeout?: number;
```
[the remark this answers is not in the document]
anitarua: good point, I can move it to the grpc config level instead of transport strategy

packages/client-sdk-nodejs/src/internal/grpc/storage-client-timeout-interceptor.ts
```diff
@@ -0,0 +1,89 @@
...
+          newDate.getMilliseconds() + this.overallRequestTimeoutMs
+        );
+        this.overallDeadline = newDate;
+
+        const deadline = new Date(Date.now());
+        deadline.setMilliseconds(
+          deadline.getMilliseconds() + this.responseDataReceivedTimeoutMs
+        );
+        options.deadline = deadline;
+        this.logger.debug(`new deadline set to ${options.deadline.valueOf()}`);
```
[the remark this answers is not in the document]
anitarua: `new deadline set to _` is logged 4 times, which seems to be in line with expectations.
However, the retry interceptor logs are showing up every 100ms, ending with: `Request eligible for retry: path: /store.Store/Get; response status code: 13; number of attempts (31); will retry in 100ms`

packages/client-sdk-nodejs/src/config/retry/storage-default-eligibility-strategy.ts
```diff
@@ -0,0 +1,104 @@
...
+            props.grpcStatus.metadata.get('retry_disposition');
+          if (retryMetadata.length === 1) {
+            const retryDisposition = retryMetadata[0].toString();
+            switch (retryMetadata[0].toString()) {
+              case 'retryable':
+                // Retryable = could retry soon (default 100ms)
+                return true;
+              case 'possibly_retryable':
+                // Possibly retryable = could retry in a second or two
+                return true; // return a value here instead?
```
anitarua, line 70: asked Ben what sdks should default to doing when "possibly_retryable" is received and he said it means the server is likely in maintenance, could retry the request in 1-2 seconds, but currently we return only true/false from this method and default to retrying in 100ms if true

packages/client-sdk-nodejs/src/internal/grpc/storage-client-timeout-interceptor.ts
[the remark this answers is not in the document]
anitarua: For the cache client which has a 0ms wait before retrying, the retry interceptor will retry as soon as possible up to the fixed count maximum.

Snippet of logs from a small test:
- 1st retry: 22:14:51.253
- 2nd retry: 22:14:51.258
- 3rd retry: 22:14:51.261

anitarua:
alternately, we could try tuning the retry options in the grpc service config channel option?

- https://grpc.io/docs/guides/retry/
- https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/README.md#supported-channel-options

packages/client-sdk-nodejs/src/config/retry/storage-default-eligibility-strategy.ts
cprice404: this doesn't match the docstrings in his client-protos PR, nor my expectations about these values based on our previous discussions:

https://github.com/momentohq/client-protos/blob/2038613adca37ef57cfd81dd7c1539a2d570bbc3/proto/store.proto#L38

my belief is that the server would always be sending `retryable` when in maintenance mode, and if we get `possibly_retryable` it means it's not safe to retry unless the operation is idempotent or the user has configured us to retry on idempotent operations.

cprice404:
> alternately, we could try tuning the retry options in the grpc service config channel option?
> 
>     * https://grpc.io/docs/guides/retry/
> 
>     * https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/README.md#supported-channel-options
> 
> 
> except they don't recommend using this with a retry interceptor, and without the interceptor setup we currently have, i don't think we'd have as fine-grained control over when to retry and for what methods (e.g. inspect metadata to determine retries)

Correct, we've ruled those out in the past because they don't give us enough control.

packages/client-sdk-nodejs/src/config/retry/storage-default-eligibility-strategy.ts
anitarua: revised the metadata parsing and removed the "err" metadata checking (it appears irrelevant to determining retries since the error code checking occurs first)

packages/client-sdk-nodejs/src/internal/grpc/headers-interceptor.ts
```diff
@@ -36,9 +36,9 @@ export class HeaderInterceptorProvider {
     return (options, nextCall) => {
       return new InterceptingCall(nextCall(options), {
         start: (metadata, listener, next) => {
-          this.headersToAddEveryTime.forEach(h =>
-            metadata.add(h.name, h.value)
-          );
+          this.headersToAddEveryTime.forEach(h => {
+            metadata.set(h.name, h.value);
```
anitarua, line 40: Bug fix: `set` will overwrite the header value, `add` will blindly append

The auth header was continually getting appended with each retried request, thus short-circuiting the actual retried request and causing that loop of retries that didn't go anywhere

anitarua: After fixing this bug, the retries via resetting grpc deadline appear to work as expected!

Logs snippet below with bold font to highlight timing: 1 second between retries, 100ms between determining eligibility and actually retrying

[2024-08-14T21:21:21.112Z] TRACE (Momento: StorageDataClient): Issuing 'put' request; store: store, key: key
...
[2024-08-14T**21:21:22.116**Z] DEBUG (Momento: RetryInterceptor): Request eligible for retry: path: /store.Store/Put; response status code: 4; number of attempts (1); will retry in 100ms
[2024-08-14T**21:21:22.217**Z] DEBUG (Momento: StorageClientTimeoutInterceptor): intercepting call with options.deadline 1723670482113
...
[2024-08-14T**21:21:23.218**Z] DEBUG (Momento: RetryInterceptor): Request eligible for retry: path: /store.Store/Put; response status code: 4; number of attempts (2); will retry in 100ms
[2024-08-14T**21:21:23.319**Z] DEBUG (Momento: StorageClientTimeoutInterceptor): intercepting call with options.deadline 1723670483217
...
[2024-08-14T**21:21:24.320**Z] DEBUG (Momento: RetryInterceptor): Request eligible for retry: path: /store.Store/Put; response status code: 4; number of attempts (3); will retry in 100ms
[2024-08-14T**21:21:24.422**Z] DEBUG (Momento: StorageClientTimeoutInterceptor): intercepting call with options.deadline 1723670484319
...
[2024-08-14T**21:21:25.423**Z] DEBUG (Momento: RetryInterceptor): Request eligible for retry: path: /store.Store/Put; response status code: 4; number of attempts (4); will retry in 100ms
[2024-08-14T**21:21:25.523**Z] DEBUG (Momento: StorageClientTimeoutInterceptor): intercepting call with options.deadline 1723670485422
[2024-08-14T21:21:25.523Z] DEBUG (Momento: StorageClientTimeoutInterceptor): received deadline < current time, resetting
[2024-08-14T21:21:25.523Z] DEBUG (Momento: StorageClientTimeoutInterceptor): Unable to successfully retry request within client timeout, canceling request
[2024-08-14T21:21:25.523Z] DEBUG (Momento: DefaultStorageRetryStrategy): Determining whether request is eligible for retry; status code: 1, request type: /store.Store/Put, attemptNumber: 5
[2024-08-14T21:21:25.523Z] DEBUG (Momento: DefaultStorageEligibilityStrategy): Response with status code 1 is not retryable.
[2024-08-14T21:21:25.523Z] DEBUG (Momento: RetryInterceptor): Request not eligible for retry: path: /store.Store/Put; retryable status code: 1; number of attempts (5).
putInt: The request was cancelled by the server; please contact us at support@momentohq.com: 1 CANCELLED: Unable to successfully retry request within client timeout

cprice404: @anitarua awesome, thanks - do you have any more changes planned on this atm or should I pull it down and play with it some more at this point?

anitarua: no further changes planned currently -- feel free to pull it down and let me know what you think!

cprice404:
tested this locally and the behavior looks about right now - thanks!

one functional thing I think we are still missing is jitter: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

you can just do something like:

```
const jitter = (Math.random() * (1.2 * delay)) - (1.1 * delay)
```

I think that will give you somewhere between -10% and + 10% jitter.

will review the code more closely in a moment but I think there are still some API things we should address here before we consider merging. Here are some rough thoughts, happy to discuss more concrete ideas/proposals:

1. I think the responseDataReceivedTimeoutMillis should be part of the retry strategy, rather than part of the static grpc config. it's only ever relevant / used by the retry strategy, right? I realize that you might need to add an accessor function to the RetryStrategy interface in order to make this work but that feels right to me?
2. we should be thinking about how a user would override this configuration if they wanted to. Probably because we want to keep the `RetryStrategy` interface generic, we can't really give them a clean path to do `config.getRetryStrategy().withBlah()`, they will have to build a retry strategy object up on their own. Which is fine but that means we want to make that as easy as possible and maybe even provide a factory function for it.
3. I don't think this retry strategy (which you've currently called `DefaultStorageRetryStrategy`) is really specific to storage use cases. I think we can give it a generic name and theoretically allow it to be used with other clients in the future. So we just need to come up with a name... for the other one we used `FixedCountRetryStrategy`, for this one I think it could maybe be called `RetryWithinDeadlineStrategy` or something like that, but I'm open to other ideas.

**Note**: it's going to be important that the computation for whether or not we can still retry lives inside of the retry strategy class itself; I think right now you might have it plumbed directly into the client. If you need to add more metadata to `DetermineWhenToRetryRequestProps` in order to allow the retry strategy to compute whether we can still retry or not, that would be fine.

let me know if you have any questions or want to zoom to brainstorm on any of this.

cc: @malandis if you have any thoughts on the names I floated, or in general the UX of a user explicitly configuring their strategy rather than using the default, would love to hear ideas

cprice404 (commented):
looking good, left a comment with a few UX/API thoughts

anitarua:
> 3. I don't think this retry strategy (which you've currently called `DefaultStorageRetryStrategy`) is really specific to storage use cases. I think we can give it a generic name and theoretically allow it to be used with other clients in the future. So we just need to come up with a name... for the other one we used `FixedCountRetryStrategy`, for this one I think it could maybe be called `RetryWithinDeadlineStrategy` or something like that, but I'm open to other ideas.

I kinda think `RetryStrategy` should remain together as a phrase though. Some additional ideas:

- FixedDeadlineRetryStrategy
- TimeIntervalRetryStrategy
- IntervalRetryStrategy

anitarua:
`FixedDeadlineRetryStrategy` or `FixedTimeoutRetryStrategy` sounds good to me.

I think we may also want to rename the interceptor that goes with it from `StorageClientTimeoutInterceptor` to something similar. `FixedDeadlineInterceptor` or `FixedTimeoutInterceptor`?

packages/client-sdk-nodejs/src/internal/grpc/retry-interceptor.ts
```diff
@@ -25,6 +25,10 @@ export function createRetryInterceptorIfEnabled(
   ];
 }
 
+function addJitter(whenToRetry: number): number {
+  return (1 + Math.random()) * (1.2 * whenToRetry) - 1.1 * whenToRetry;
+}
+
```
anitarua, line 31: slightly different formula as the previous one was returning negative milliseconds

packages/client-sdk-nodejs/src/config/retry/storage-default-retry-strategy.ts
```diff
@@ -0,0 +1,75 @@
...
+  ): DefaultStorageRetryStrategy {
+    return new DefaultStorageRetryStrategy({
+      loggerFactory: props?.loggerFactory ?? new DefaultMomentoLoggerFactory(),
+      eligibilityStrategy: props?.eligibilityStrategy,
+      retryDelayIntervalMillis: props?.retryDelayIntervalMillis,
+      responseDataReceivedTimeoutMillis:
+        props?.responseDataReceivedTimeoutMillis,
+    });
+  }
+}
```
anitarua, line 75: added a factory for creating this retry strategy

cprice404:
> `FixedDeadlineRetryStrategy` or `FixedTimeoutRetryStrategy` sounds good to me.
> 
> I think we may also want to rename the interceptor that goes with it from `StorageClientTimeoutInterceptor` to something similar. `FixedDeadlineInterceptor` or `FixedTimeoutInterceptor`?

oh, I didn't realize you had to create a custom interceptor but that makes sense in retrospect.

The ideal scenario would be that we only have one RetryInterceptor, and that we have augmented the `RetryStrategy` interface enough for that single interceptor to work with all the different retry strategies.

If we can't figure out how to do that, then we're in a situation where users can't swap in a different `RetryStrategy` without also changing the interceptor, and that will make our API/UX a lot more burdensome/confusing. So if we can't figure out how to pull off a single shared interceptor then we should push pause and talk through options.

happy to pair on what the refactoring of the shared interceptor would need to look like in order for it to successfully accommodate multiple strategies.

packages/client-sdk-nodejs/src/internal/storage-data-client.ts
```diff
@@ -111,9 +114,31 @@ export class StorageDataClient implements IStorageDataClient {
...
+      const responseDataReceivedTimeoutMs =
+        retryStrategy.getResponseDataReceivedTimeoutMillis();
+      timeoutInterceptor = createStorageClientTimeoutInterceptor(
+        _loggerFactory,
+        this.requestTimeoutMs,
+        responseDataReceivedTimeoutMs
+      )[0];
+    } else {
+      timeoutInterceptor = ClientTimeoutInterceptor(this.requestTimeoutMs);
+    }
```
anitarua, line 132: oh whoops, saw your comment just after I'd pushed a fix for that. I made the internal `initializeInterceptors` function choose which interceptor to add based on the type of the retry strategy, but I can look into creating a single interceptor instead

packages/client-sdk-nodejs/src/config/storage-configuration.ts
```diff
@@ -1,6 +1,9 @@
-import {MomentoLoggerFactory} from '../';
+import {MomentoLoggerFactory, RetryStrategy} from '../';
 import {StorageTransportStrategy} from './transport/storage';
 
+// TODO: include ResponseDataReceivedTimeout as a top-level config?
+// Or keep at the StorageTransportStrategy level?
```
malandis, line 5: This seems like lower level. We can keep it in transport strategy and add an accessor up here if it becomes an issue.

packages/client-sdk-nodejs/src/internal/grpc/client-timeout-interceptor.ts
```diff
@@ -1,14 +1,123 @@
...
+
+        const deadline = new Date(Date.now());
+        deadline.setMilliseconds(
+          deadline.getMilliseconds() + this.responseDataReceivedTimeoutMs
+        );
+
+        if (deadline.valueOf() > this.overallDeadline.valueOf()) {
+          this.logger.debug(
+            'Unable to successfully retry request within client timeout, canceling request'
+          );
```
malandis, line 105: An important consideration here is how close to the deadline do we need to be to not bother retrying. In the current implementation, that value is `responseDataReceivedTimeoutMs`:

Consider an overall deadline at time=1000ms, a `responseDataReceivedTimeoutMs` of 100ms, and that we're now at time=925ms. Then the request deadline is `925ms + 100ms = 1025ms`, which exceeds our overall deadline, so we will not retry. But maybe the request could succeed in less than 75ms?

We could consider some percentage of the data received timeout instead, like 50%. Up for discussion.

So we need to decide how close to the `overallDeadline` do we need to be to not even bother retrying. Right now if are within `responseDataReceivedTimeoutMs` of the `overallDeadline`, we will not retry.

malandis (commented):
I left a few comments for discussion. Looks great.

packages/client-sdk-nodejs/src/internal/grpc/retry-interceptor.ts
anitarua: The original was suggested by Chris:

```typescript
const jitter = Math.random() * (1.2 * delay) - (1.1 * delay)
```

I saw it was producing negative outputs so I added a  `+ 1`:

```typescript
const jitter = (1 + Math.random()) * (1.2 * delay) - (1.1 * delay)
```

But the goal was to add between -10% to +10% of jitter to the delay value. 
I think there's probably some more tweaking to do to the formula on second glance. Perhaps modifying your simplified version like so?

```typescript
const jitter = (0.2 * Math.random() + 0.9) * whenToRetry
```

packages/client-sdk-nodejs/src/internal/grpc/retry-interceptor.ts
```diff
@@ -91,10 +96,14 @@ export class RetryInterceptor {
                       savedMessageNext(savedReceiveMessage);
                       next(status);
                     } else {
+                      const whenToRetryWithJitter = addJitter(whenToRetry);
```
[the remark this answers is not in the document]
anitarua: Possibly. I just added jitter here because there was a [TODO note in this file](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/internal/grpc/retry-interceptor.ts#L43) about it

packages/client-sdk-nodejs/src/config/storage-configuration.ts
anitarua: oh whoops, I moved ResponseDataReceivedTimeout to be part of the specific RetryStrategy class it's used for. Will remove this comment

packages/client-sdk-nodejs/src/config/transport/storage/transport-strategy.ts
```diff
@@ -22,11 +22,11 @@ export interface StorageTransportStrategy {
 
   /**
    * Copy constructor to update the client-side timeout
-   * @param {number} clientTimeoutMillis
+   * @param {number} requestTimeoutMillis
```
anitarua, line 25: should I also undo this `clientTimeoutMillis` to `requestTimeoutMillis` renaming now that the new settings are confined to retry strategy and not transport/grpc settings?

packages/client-sdk-nodejs/src/internal/grpc/client-timeout-interceptor.ts
anitarua: For sure! If we do a percent-based margin around that, would that have to become another configuration option? Or will we just hardcode that in for now?

packages/client-sdk-nodejs/src/internal/grpc/retry-interceptor.ts
cprice404: that last one looks promising. -10% to + 10% is definitely the goal.

packages/client-sdk-nodejs/src/internal/grpc/client-timeout-interceptor.ts
```diff
@@ -1,14 +1,123 @@
...
-    if (!options.deadline) {
-      const deadline = new Date(Date.now());
-      deadline.setMilliseconds(deadline.getMilliseconds() + requestTimeoutMs);
-      options.deadline = deadline;
-    }
-    return new InterceptingCall(nextCall(options));
-  };
+  if (
+    retryStrategy !== undefined &&
+    retryStrategy instanceof DefaultStorageRetryStrategy
```
cprice404, line 16: I think I like how you broke the timeout interceptors up into two cases here, but we don't want to do this instanceof check to choose between them. One design goal here is that a user could write their own implementation of `RetryStrategy` and leverage the building blocks we are putting together, and this would preclude that.

I suspect the best way to handle this is going to be to extend the `interface` for `RetryStrategy` to include the (optional) `responseDataReceivedTimeoutMillis` property. Then you can just check that here to decide which timeout strategy to use, and if users write their own retry strategy that adheres to the interface, they will get the behavior too.

packages/client-sdk-nodejs/src/internal/grpc/client-timeout-interceptor.ts
```diff
@@ -1,14 +1,123 @@
...
+        return new InterceptingCall(nextCall(options));
+      }
+
+      const receivedDeadline = options.deadline?.valueOf() || 0;
+      this.logger.debug(
+        `intercepting call with options.deadline ${receivedDeadline}`
+      );
+
+      // Reset incremental deadline only if it has been reached
+      if (receivedDeadline < Date.now()) {
```
cprice404, line 94: this is probably okay, but a little different than what I'd been imagining. I'd been imagining that if we had hit a timeout then we would have gotten a response status of DEADLINE_EXCEEDED, and that we could check that to decide whether to reset the deadline.

I haven't thought that through completely, it's possible there are cases where your approach is better and handles something that the one I am suggesting would not, but it's worth thinking about.

packages/client-sdk-nodejs/src/internal/grpc/client-timeout-interceptor.ts
cprice404: would like to avoid more config options if possible. could we not just check `Date.now() > this.overallDeadline`, and if it's not, then we set the next deadline to `min(this.overallDeadline, deadline + this.responseDataReceivedTimeoutMs)`?

packages/client-sdk-nodejs/src/internal/grpc/retry-interceptor.ts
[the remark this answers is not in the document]
cprice404: agreed with michael, the jitter should go inside the retry strategy, and we can just remove that old comment. Thanks!

cprice404 (commented):
Did another partial review on this and called out a few things. Feels like we are getting close, thanks!

cprice404:
> `FixedDeadlineRetryStrategy` or `FixedTimeoutRetryStrategy` sounds good to me.
> 
> I think we may also want to rename the interceptor that goes with it from `StorageClientTimeoutInterceptor` to something similar. `FixedDeadlineInterceptor` or `FixedTimeoutInterceptor`?

checking back in on this: I think I'm leaning `FixedTimeoutRetryStrategy`, so we will have that one and `FixedCountRetryStrategy` as the two pre-existing ones. @malandis if you have any other ideas here lmk.

packages/client-sdk-nodejs/src/internal/grpc/client-timeout-interceptor.ts
```diff
@@ -1,14 +1,123 @@
...
+      overallRequestTimeoutMs,
+      responseDataReceivedTimeoutMs
+    ).createTimeoutInterceptor();
+  }
+  return new BasicTimeoutInterceptor(
+    overallRequestTimeoutMs
+  ).createTimeoutInterceptor();
 };
+
+export class BasicTimeoutInterceptor {
```
cprice404, line 31: I don't think these need to be exported, hopefully they are never used/visible outside of this file, right?

packages/client-sdk-nodejs/src/internal/grpc/client-timeout-interceptor.ts
```diff
@@ -1,14 +1,123 @@
...
+          deadline.getMilliseconds() + this.overallRequestTimeoutMs
+        );
+        options.deadline = deadline;
+      }
+      return new InterceptingCall(nextCall(options));
+    };
+  }
+}
+
+export class RetryUntilTimeoutInterceptor {
```
cprice404, line 52: there was another thread somewhere about the names of these; i think these names are probably fine for now, especially if you are able to un-export them so they are private implementation details inside this file.

anitarua:
I think I've addressed all the feedback so far since the last major revision:

- simplified the `RetryUntilTimeoutInterceptor` implementation (was messy from trying to debug the retry interceptor bug)
- added the optional property `responseDataReceivedTimeoutMillis` to the `RetryStrategy` interface and `ClientTimeoutInterceptor` checks for existence of that instead of the `instanceof` check
- updated jitter formula to return between 0.9 to 1.1 times the original `whenToRetry` value
- moved jitter addition into the renamed `FixedTimeoutRetryStrategy` class
- no longer export the timeout interceptor classes

My remaining [question](https://github.com/momentohq/client-sdk-javascript/pull/1396#discussion_r1725524079) is whether to revert some renamings in transport/grpc configs.

Any other questions/concerns on your end, @cprice404 and @malandis ?

packages/client-sdk-nodejs/src/internal/grpc/client-timeout-interceptor.ts
```diff
@@ -1,14 +1,123 @@
...
+      );
+      if (newDeadline > this.overallDeadline) {
+        this.logger.debug(
+          `new deadline would exceed overall deadline, setting to overall deadline ${this.overallDeadline.valueOf()}`
+        );
+        options.deadline = this.overallDeadline;
+      } else {
+        this.logger.debug(`new deadline set to ${newDeadline.valueOf()}`);
+        options.deadline = newDeadline;
+      }
```
malandis, line 118: This is almost there. We need to make a couple small changes:

- get the current time
- if the `overallDeadline` has already elapsed (current time exceeds `overallDeadline`), then cancel;
- otherwise we have *some* time left to run the operation.
    - In that case we set the deadline to current ms + `responseDataReceivedTimeoutMs`
    - We clip the deadline to `overallDeadline` if needed

I recommend a helper function to compute this since the calculation is repeated on line 76. We should also document we have intentionally decided to run the operation even when we have less time than `responseDataReceivedTimeoutMs`

malandis (commented):
Almost there!

packages/client-sdk-nodejs/src/config/retry/fixed-timeout-retry-strategy.ts
```diff
@@ -0,0 +1,75 @@
...
+    );
+    if (!this.eligibilityStrategy.isEligibleForRetry(props)) {
+      // null means do not retry
+      return null;
+    }
+
+    this.logger.debug(
+      `Request is eligible for retry (attempt ${props.attemptNumber}), retrying soon.`
+    );
+    // retry after a fixed time interval has passed (=/- some jitter)
```
malandis, line 54:
```suggestion
    // retry after a fixed time interval has passed (+/- some jitter)
```

packages/client-sdk-nodejs/src/config/transport/storage/transport-strategy.ts
cprice404: i think i have a slight pref for keeping this consistent with the existing configs for CacheClient etc, so yeah I'd probably change it back. thanks for catching that.

anitarua:
Test is failing -- discovered that the interceptor is not resetting the overall deadline between different successful requests.

packages/client-sdk-nodejs/src/config/retry/fixed-timeout-retry-strategy.ts
```diff
@@ -0,0 +1,75 @@
...
+    return addJitter(this.retryDelayIntervalMillis);
+  }
+}
+
+export class FixedTimeoutRetryStrategyFactory {
+  static getRetryStrategy(
+    props?: FixedTimeoutRetryStrategyProps
+  ): FixedTimeoutRetryStrategy {
+    return new FixedTimeoutRetryStrategy({
+      loggerFactory: props?.loggerFactory ?? new DefaultMomentoLoggerFactory(),
```
cprice404, line 64: I don't think we can do this (default value for logger factory) - it can result in them ending up with two different logging destinations. Hopefully in the pre-built configurations we are constructing the strategy using the single logger factory that gets passed in there. And I think we just have to lean into the fact that if someone wants to customize the strategy they will have to pass in a logger factory.

and then, if we remove that, this factory function isn't really any more succinct than the main constructor so I think we can ditch it.

I think with the state you've gotten this stuff to now, the constructor is pretty approachable/reasonable if someone wants to customize the arguments, so we are probably good there. But I will pull down the code and get my hands dirty with the API/UX for a few minutes to confirm.

packages/client-sdk-nodejs/src/config/retry/fixed-timeout-retry-strategy.ts
```diff
@@ -0,0 +1,75 @@
...
+      eligibilityStrategy: props?.eligibilityStrategy,
+      retryDelayIntervalMillis: props?.retryDelayIntervalMillis,
+      responseDataReceivedTimeoutMillis:
+        props?.responseDataReceivedTimeoutMillis,
+    });
+  }
+}
+
+export function addJitter(whenToRetry: number): number {
```
cprice404, line 73: this probably doesn't need to be exported?

cprice404 (commented):
just a few more minor nit comments from me, and i wanna play around with instantiating the strategy directly, but i'm very close to 🚢

packages/client-sdk-nodejs/src/config/retry/fixed-timeout-retry-strategy.ts
```diff
@@ -0,0 +1,75 @@
...
+  eligibilityStrategy?: EligibilityStrategy;
+
+  // Retry request after a fixed time interval (defaults to 100ms)
+  retryDelayIntervalMillis?: number;
+
+  // Number of milliseconds the client is willing to wait for response data to be received before retrying (defaults to 1000ms). After the overarching GRPC config deadlineMillis has been reached, the client will terminate the RPC with a Cancelled error.
+  responseDataReceivedTimeoutMillis?: number;
+}
+
+export class FixedTimeoutRetryStrategy implements RetryStrategy {
```
cprice404, line 24: we need to add this to the exports in the main `index.ts` so that users can import it.

cprice404:
I pulled this down and tested it with this code:

```
const loggerFactory = new DefaultMomentoLoggerFactory(DefaultMomentoLoggerLevel.TRACE);

  const retryStrategy = new FixedTimeoutRetryStrategy({
    loggerFactory: loggerFactory,
    responseDataReceivedTimeoutMillis: 100,
    retryDelayIntervalMillis: 1000,
  });

  const storageClient = new PreviewStorageClient({
    configuration: StorageConfigurations.Laptop.latest(loggerFactory).withRetryStrategy(retryStrategy),
    credentialProvider: CredentialProvider.fromEnvironmentVariable('MOMENTO_API_KEY').withMomentoLocal(),
  });

```

My thoughts:
1. I am good with this UX for overriding the retry strategy. Feels reasonable to me. Nice job!
2. Anecdotally it seemed like it was probably doing the right thing; I saw 5 requests get through to the server which is what I was expecting.
3. It'd be nice if the trace-level logs in the retrystrategy/timeout interceptor gave a bit more timing info; i'd like to see my actual configured values (100, 1000) appear in those logs so it'd be easy to mentally connect them back to my config.
4. Also in the logs - when we fail and give up, it shows up in the logs as `attemptNumber: 6` and `7`. That threw me off for a second because I was only expecting 5 requests. Then I watched the server-side logs and confirmed it was actually only 5 requests, so I think it's just a slightly unexpected path through the retry code that causes those last two sets of logs to get printed, even though the request isn't actually being issued again. This is not really a big deal, not worth spending a ton of time on, but if you are seeing this in your logs too and you see any easy way to clear up those last two chunks of logs that'd be bonus points.

packages/client-sdk-nodejs/src/internal/grpc/client-timeout-interceptor.ts
```diff
@@ -64,48 +65,79 @@ class RetryUntilTimeoutInterceptor {
...
       }
 
+      // Reset overall and incremental deadlines in these cases:
+      // Case 1: overallDeadline is undefined (first request or previous was canceled)
+      // Case 2: different RPC path requested through same client
+      // Case 3: new request through same client after original retry deadline has passed
+      // (Note: Case 3 check must occur after the receivedDeadline == options.deadline check
+      // or else the request is always retried).
+      if (
+        this.overallDeadline === undefined ||
```
cprice404, line 102: oh gosh, I hadn't thought about the scope of this variable. If this interceptor is basically global to the client object, and we're trying to store this overallDeadline in a member variable, then we only have 1 of those variables for the whole client. That won't work, the client could be handling 100s of concurrent requests.

We'll need a way to have a separate copy of that variable for each request. I think you might be able to do that by defining that var using a `let` to create a local variable inside of the callback function, so that the deadline logic is basically a closure over the local variable.

Another option might be to store it in the metadata of the request, but we don't really want it to get sent up to the server. maybe there's some other place on the request object that we can attach metadata?

Failing that I think we would need to maintain a map where each request had its own key/value, but that will be tricky to make sure we clean it up properly so that we don't leak memory.

:(

very glad you caught this though. We should think about how to test this with two concurrent requests.

I can pull the code down again in a second and see if I can help figure out a way to store the variable as a local inside of the closure rather than a class instance variable.

cprice404:
I patched @anitarua 's latest version to combine the `ClientTimeoutInterceptor` into the `RetryInterceptor`. They were starting to have too much overlap, and the RetryInterceptor was already set up in a way that made it easy to create local variables per-request to keep track of overall deadlines, and just close over those variables with the higher-order functions. This solves the problem where we were trying to maintain that state as a member variable on the ClientTimeoutInterceptor class.

No public API changes from Anita's version.

Tested locally using the program below; I believe this is ready for review/merge now:

```typescript
import {
  CreateStoreResponse,
  CredentialProvider,
  DefaultMomentoLoggerFactory,
  DefaultMomentoLoggerLevel,
  FixedTimeoutRetryStrategy,
  PreviewStorageClient,
  StorageConfigurations,
  StorageGetResponse,
  StoragePutResponse,
} from '@gomomento/sdk';

async function startWorker(storageClient: PreviewStorageClient, id: number) {
  await delay(Math.random() * 1000);
  const storeName = 'my-store';
  const putResponse = await storageClient.putString(storeName, `test-key-${id}`, `test-value-${id}`);
  switch (putResponse.type) {
    case StoragePutResponse.Success:
      console.log(`Key 'test-key-${id}' stored successfully`);
      break;
    case StoragePutResponse.Error:
      console.log(`Error occurred when trying to store key 'test-key-${id}': ${putResponse.toString()}`);
      break;
    // throw new Error(
    //   `An error occurred while attempting to store key 'test-key-${id}' in store '${storeName}': ${putResponse.errorCode()}: ${putResponse.toString()}`
    // );
  }

  const getResponse = await storageClient.get(storeName, `test-key-${id}`);
  // pattern-matching style; safer for production code
  switch (getResponse.type) {
    case StorageGetResponse.Found:
      // if you know the value is a string:
      console.log(`Retrieved value for key 'test-key-${id}': ${getResponse.value().string()!}`);
      break;
    case StorageGetResponse.NotFound:
      console.log(`Key 'test-key-${id}' was not found in store '${storeName}'`);
      break;
    case StorageGetResponse.Error:
      console.log(`Error occurred when trying to get key 'test-key-${id}': ${getResponse.toString()}`);
    // throw new Error(
    //   `An error occurred while attempting to get key 'test-key' from store '${storeName}': ${getResponse.errorCode()}: ${getResponse.toString()}`
    // );
  }
}

async function main() {
  const loggerFactory = new DefaultMomentoLoggerFactory(DefaultMomentoLoggerLevel.TRACE);

  const retryStrategy = new FixedTimeoutRetryStrategy({
    loggerFactory: loggerFactory,
    responseDataReceivedTimeoutMillis: 1000,
    retryDelayIntervalMillis: 10,
  });

  const storageClient = new PreviewStorageClient({
    configuration: StorageConfigurations.Laptop.latest(loggerFactory).withRetryStrategy(retryStrategy),
    credentialProvider: CredentialProvider.fromEnvironmentVariable('MOMENTO_API_KEY').withMomentoLocal(),
  });

  const storeName = 'my-store';
  const createStoreResponse = await storageClient.createStore(storeName);
  switch (createStoreResponse.type) {
    case CreateStoreResponse.AlreadyExists:
      console.log(`Store '${storeName}' already exists`);
      break;
    case CreateStoreResponse.Success:
      console.log(`Store '${storeName}' created`);
      break;
    case CreateStoreResponse.Error:
      throw new Error(
        `An error occurred while attempting to create store '${storeName}': ${createStoreResponse.errorCode()}: ${createStoreResponse.toString()}`
      );
  }

  const workerPromises = range(5).map(i => startWorker(storageClient, i));
  await Promise.all(workerPromises);
}

function range(n: number): Array<number> {
  return [...Array(n).keys()];
}

function delay(ms: number) {
  return new Promise(resolve => setTimeout(resolve, ms));
}

main().catch(e => {
  console.error('An error occurred! ', e);
  throw e;
});

```

cprice404:
cc: @malandis may want to get your eyes on this when you have some time.

malandis (commented):
lgtm; one question about topics

malandis (approved):

cprice404:
Ran the topic subscribe example with this code; turned off wifi, observed the client detect the connection issue and go into the retry loop; when i turned the wifi back on it reconnected successfully.

PR #1397, opened 2024-08-01
feat: add SetPop and SetLength

anitarua:
Adds missing set methods to node and web SDKs

### DIFF
--- a/packages/common-integration-tests/src/set.ts
+++ b/packages/common-integration-tests/src/set.ts
@@ -1043,9 +1043,9 @@ export function runSetTests(
         setName
       );
       expectWithMessage(() => {
-        expect(lengthResponse).toBeInstanceOf(CacheSetLength.Hit);
+        expect(lengthResponse).toBeInstanceOf(CacheSetLength.Miss);
       }, `expected HIT but got ${lengthResponse.toString()}`);
-      expect((lengthResponse as CacheSetLength.Hit).length()).toEqual(0);
+      // expect((lengthResponse as CacheSetLength.Hit).length()).toEqual(0);
     });
 
     it('should return HIT when set exists - count > set size', async () => {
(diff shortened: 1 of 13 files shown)


packages/common-integration-tests/src/set.ts
```diff
@@ -921,4 +923,285 @@ export function runSetTests(
...
+        expect(responseMore).toBeInstanceOf(CacheSetPop.Hit);
+      }, `expected HIT but got ${responseMore.toString()}`);
+      const valuesArrayMore = (responseMore as CacheSetPop.Hit).valueArray();
+      expect(valuesArrayMore).toBeArrayOfSize(5);
+      for (const value of valuesArrayMore) {
+        expect(elements).toContain(value);
+      }
+    });
+
+    it('should support happy path for fetch via curried cache via ICache interface', async () => {
```
malandis, line 1067:
```suggestion
    it('should support happy path for setPop via curried cache via ICache interface', async () => {
```

packages/common-integration-tests/src/set.ts
```diff
@@ -921,4 +923,285 @@ export function runSetTests(
...
+      );
+      expectWithMessage(() => {
+        expect(responseZero).toBeInstanceOf(CacheSetLength.Hit);
+      }, `expected HIT but got ${responseZero.toString()}`);
+      expect((responseZero as CacheSetLength.Hit).length()).toEqual(
+        elements.length
+      );
+    });
+
+    it('should support happy path for fetch via curried cache via ICache interface', async () => {
```
malandis, line 1159:
```suggestion
    it('should support happy path for setLength via curried cache via ICache interface', async () => {
```

packages/core/src/internal/clients/cache/AbstractCacheClient.ts
```diff
@@ -751,6 +753,46 @@ export abstract class AbstractCacheClient implements ICacheClient {
     return await client.setSample(cacheName, setName, limit);
   }
 
+  /**
+   * Fetch and remove a random sample of elements from the set.
+   * Returns a different random sample for each call.
```
malandis, line 758: I'm not clear what we mean by this comment? Is it that when we call set pop, we remove elements, hence can never get the same random sample twice? Or do we mean something else?

malandis (commented):
Looks good. I have a few clarifying comments

malandis (approved):

PR #1399, opened 2024-08-02
chore: refactor AbstractPubSubClient to ensure safe use of required vars

cprice404:
This commit contains no functional changes; simply refactors the
AbstractPubSubClient to stop leaking out member variables to its
subclasses. The way it was previously defined, there were several
member variables that were required to be initialized by the child
classes, which, if you did not initialize them would result in
a runtime error instead of a compile-time error. This commit
changes the variables to be private, provides protected accessor
functions as needed, and creates a constructor function that requires
them to be passed in at initialization time. This moves any related
bugs from run time to compile time and makes it simpler to reason
about the ownership of the variables.

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/pubsub-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/pubsub-client.ts
@@ -10,7 +10,6 @@ import {version} from '../../package.json';
 import {middlewaresInterceptor} from './grpc/middlewares-interceptor';
 import {
   CredentialProvider,
-  MomentoLogger,
   StaticGrpcConfiguration,
   TopicItem,
   TopicPublish,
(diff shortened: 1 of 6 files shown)


pgautier404 (dismissed):
LGTM!

packages/client-sdk-nodejs/src/internal/pubsub-client.ts
```diff
@@ -63,10 +58,21 @@ export class PubsubClient extends AbstractPubsubClient<ServiceError> {
     const grpcConfig = new StaticGrpcConfiguration({
       deadlineMillis: this.unaryRequestTimeoutMs,
       maxSessionMemoryMb: PubsubClient.DEFAULT_MAX_SESSION_MEMORY_MB,
+      keepAlivePermitWithoutCalls: 1,
+      keepAliveTimeMs: 5000,
+      keepAliveTimeoutMs: 1000,
```
cprice404, line 63: this should not be in here. this is what i was working on when i got distracted by the confusing variables... will remove this

packages/client-sdk-nodejs/src/internal/pubsub-client.ts
```diff
@@ -67,6 +62,14 @@ export class PubsubClient extends AbstractPubsubClient<ServiceError> {
 
     const channelOptions = grpcChannelOptionsFromGrpcConfig(grpcConfig);
 
+    console.log(
```
anitarua, line 65: stray console log?

cprice404: thank you

anitarua (approved):

PR #1401, opened 2024-08-06
fix: broken auth tests

cprice404:
This commit fixes the auth tests; we no longer use the auth0 token
for instantiating any clients, we only use the MGA account session
token.

This commit also changes the run scripts in the `package.json` such
that the generic `integration-test` target runs *all* integration
tests; previously it was only running cache/control/topics tests,
which felt counter-intuitive to me. I added a new target that runs
only the cache/control/topics tests but it seemed to me like the
default behavior should be to run all of the tests, to avoid
surprising anyone and missing the execution of certain tests.

There may be changes necessary in the canaries code to account for this.


packages/client-sdk-nodejs/package.json
```diff
@@ -19,7 +19,8 @@
     "integration-test-store": "jest storage.test.ts --maxWorkers 1",
     "unit-test": "jest unit",
     "integration-test-leaderboard": "jest leaderboard --maxWorkers 1",
-    "integration-test": "jest integration --testPathIgnorePatterns \"auth-client.test.ts|leaderboard.test.ts|storage.test.ts\" --maxWorkers 1",
+    "integration-test-control-cache-topics": "jest integration --testPathIgnorePatterns \"auth-client.test.ts|leaderboard.test.ts|storage.test.ts\" --maxWorkers 1",
+    "integration-test": "jest integration --maxWorkers 1",
```
cprice404, line 23: @momentohq/dev-eco lmk if anyone has any concerns about this change. It just seemed confusing / suprising to me that the default target didn't run all the tests.

rishtigupta: no concerns from my end 🚀

rishtigupta (approved):

rishtigupta:
:shipit:

PR #1404, opened 2024-08-12
feat: port webhooks api to new switch pattern

rishtigupta:
## Description:
I had to comment out the example code for webhooks since the CI kept failing for it. 
Will publish this sdk version and update the tests accordingly in a follow-up pr.

## Issue
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/956

### DIFF
--- a/examples/nodejs/cache/doc-example-files/doc-examples-js-apis.ts
+++ b/examples/nodejs/cache/doc-example-files/doc-examples-js-apis.ts
@@ -1481,64 +1481,64 @@ async function example_API_TopicSubscribe(topicClient: TopicClient, cacheName: s
   }
 }
 
-// async function example_API_ListWebhooks(topicClient: TopicClient, cacheName: string) {
-//   const result = await topicClient.listWebhooks(cacheName);
-//   if (result instanceof ListWebhooks.Success) {
-//     // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-//     console.log(`listed webhooks: ${result.getWebhooks()}`);
-//   } else if (result instanceof ListWebhooks.Error) {
-//     throw new Error(
-//       `An error occurred while attempting to list webhooks for cache '${cacheName}': ${result.errorCode()}: ${result.toString()}`
-//     );
-//   }
-// }
+async function example_API_ListWebhooks(topicClient: TopicClient, cacheName: string) {
+  const result = await topicClient.listWebhooks(cacheName);
+  if (result instanceof ListWebhooks.Success) {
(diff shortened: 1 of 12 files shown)


examples/nodejs/cache/doc-example-files/doc-examples-js-apis.ts
```diff
@@ -1481,64 +1481,64 @@ async function example_API_TopicSubscribe(topicClient: TopicClient, cacheName: s
...
-  const result = await topicClient.getWebhookSecret(cacheName, 'examples webhook');
-  if (result instanceof GetWebhookSecret.Success) {
-    console.log('successfully retrieved the webhook secret');
-  } else if (result instanceof GetWebhookSecret.Error) {
-    throw new Error(
-      `An error occurred while attempting to fetch the secret for the webhook 'examples webhook' inside of cache '${cacheName}': ${result.errorCode()}: ${result.toString()}`
-    );
-  }
-}
+// async function example_API_ListWebhooks(topicClient: TopicClient, cacheName: string) {
```
rishtigupta, line 1484: I had to comment out the example code for webhooks since the CI kept failing for it.
Will publish this sdk version and update the tests accordingly in a follow-up pr.

cprice404: i'm confused as to why these would fail, the package.json in the examples dir should be pinned to a released version of the SDK. Also it's important for the sake of backward compatibility that this syntax still work, so I think we will need to get to the bottom of it before we merge this.

examples/nodejs/cache/package-lock.json
```diff
@@ -122,6 +122,7 @@
       "version": "0.9.5",
```
cprice404, line 122: why did this file change? just curious

packages/common-integration-tests/src/webhooks.ts
```diff
@@ -98,16 +97,17 @@ export function runWebhookTests(
           webhook.id.cacheName,
           webhook.id.webhookName
         );
-        if (resp instanceof GetWebhookSecret.Success) {
```
cprice404, line 101: it would be good to keep at least one test around that uses the old syntax.

cprice404 (commented):
code looks good at a glance. just a few comments about keeping some test coverage for the old pattern, and getting to the bottom of whatever is going on with the examples

examples/nodejs/cache/package-lock.json
rishtigupta: While I was experimenting with the ci build failures, I reinstalled my node modules

PR #1409, opened 2024-08-15
feat: adds order opt to sortedset getRank

eaddingtonwhite:
Currently for `sortedSetGetRank` we only support fetching in `Ascending` order.
```Typescript
await momento.sortedSetGetRank('my-cache', 'my-sorted-set', 'my-element');
```

This PR adds optional call options to `sortedSetGetRank` to sort the sorted set by `Descending` or `Ascending` order when determining the rank of the passed element.
So can call like this now:
```typescript
await momento.sortedSetGetRank('my-cache', 'my-sorted-set', 'my-element', {
    order: SortedSetOrder.Descending
});
```

### DIFF
--- a/examples/nodejs/cache/basic.ts
+++ b/examples/nodejs/cache/basic.ts
@@ -1,10 +1,12 @@
 import {
   CacheClient,
-  Configurations,
-  CredentialProvider,
-  CreateCacheResponse,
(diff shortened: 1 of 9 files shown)


anitarua (approved):
lgtm!

PR #1410, opened 2024-08-15
refactor: organize integration tests by service

malandis:
Organizes the integration tests by service, and importantly by directory. By organizing by directory, we have an unambiguous regex to use for jest to match integration tests by service, eg `jest cache/` to run all cache service tests. Integration test authoring stays maintainable because developers just need to place their integration test file under the appropriate service directory.

For clarity we have refactored `common-integration-tests`, the integration tests under node-js, and those under web to have the same directory structure.

We have updated the scripts in the `package.json` files to match on the service (as a directory) as opposed to individual test files.

A note on the organization:
- We moved the webhooks test under the topics service test directory.
- We kept the client constructor tests in the top-level integration test directory.

In a following PR we will introduce makefiles that will invoke these scripts and update GitHub Actions. That is, the makefiles will run `npm run integration-test-<>`. That way the source of truth for the tests will be the scripts in `package.json`, since various shell scripts still reference these.


packages/client-sdk-nodejs/package.json
```diff
@@ -15,12 +15,15 @@
...
     "prebuild": "eslint . --ext .ts",
     "test": "jest --testPathIgnorePatterns auth-client.test.ts --maxWorkers 1",
-    "integration-test-auth": "jest auth-client.test.ts --maxWorkers 1",
-    "integration-test-store": "jest storage.test.ts --maxWorkers 1",
-    "unit-test": "jest unit",
-    "integration-test-leaderboard": "jest leaderboard --maxWorkers 1",
-    "integration-test-control-cache-topics": "jest integration --testPathIgnorePatterns \"auth-client.test.ts|leaderboard.test.ts|storage.test.ts\" --maxWorkers 1",
+    "integration-test-auth": "jest auth/ --maxWorkers 1",
+    "integration-test-cache": "jest cache/ --maxWorkers 1",
+    "integration-test-control-cache-topics": "npm run integration-test-cache && npm run integration-test-topics",
```
malandis, line 20: The canaries reference this script, so we leave it in place for now (but updated) to not break it.

packages/client-sdk-nodejs/package.json
```diff
@@ -15,12 +15,15 @@
...
+    "integration-test-leaderboard": "jest leaderboard/ --maxWorkers 1",
+    "integration-test-storage": "jest storage/ --maxWorkers 1",
+    "integration-test-store": "npm run integration-test-storage",
```
malandis, line 23: The canaries reference this script, so we leave it in place for now (but updated) to not break it.

packages/client-sdk-nodejs/package.json
```diff
@@ -15,12 +15,15 @@
...
+    "integration-test-topics": "jest topics/ --maxWorkers 1",
     "integration-test": "jest integration --maxWorkers 1",
```
malandis, line 25: The `integration-test` script is still used by the shell scripts under `./scripts` in this repo.

packages/common-integration-tests/src/index.ts
```diff
@@ -1,16 +1,17 @@
...
+export * from './cache/sorted-set';
+export * from './cache/item-get-ttl';
+export * from './cache/item-get-type';
+export * from './topics/topic-client';
+export * from './cache/keys-exist';
+export * from './cache/update-ttl';
+export * from './leaderboard/leaderboard-client';
+export * from './topics/webhooks';
+export * from './cache/batch-get-set';
+export * from './storage/storage';
```
malandis, line 17: In a future PR, we could export the service names only as opposed to all the names within each of the modules. That way we can properly namespace them in tests. Since this is an internal refactor, we leave that for future work.

rishtigupta (approved):
:shipit:

packages/common-integration-tests/src/storage/storage.ts
```diff
@@ -8,7 +8,7 @@ import {
   StorageGetResponse,
   StoragePutResponse,
 } from '@gomomento/sdk-core';
-import {testStoreName} from './common-int-test-utils';
+import {testStoreName} from '..//common-int-test-utils';
```
cprice404, line 11: ?

cprice404 (commented):
lgtm w/2 questions:

1. Are you planning to do a Makefile in a follow-up PR, and, as mentioned on your other PR,
2. Don't we want to still keep control plane tests in their own targets since it's a different DNS name?

packages/common-integration-tests/src/storage/storage.ts
malandis: I don't know how this still built! Fixed in https://github.com/momentohq/client-sdk-javascript/pull/1410/commits/fb8eec51c612dba7d2b56a60347d5192645b02a7.

malandis:
> lgtm w/2 questions:
> 
> 1. Are you planning to do a Makefile in a follow-up PR, and, as mentioned on your other PR,
Yes. In the description of this one I mentioned I would follow up on this one too :-). The volume of changes was large and I kept getting merge conflicts with other ongoing PRs.

> 2. Don't we want to still keep control plane tests in their own targets since it's a different DNS name?
I don't think so. At least for the canaries, the canaries test based on service availability. So while every cell has a control plane, they may not have the leaderboard control plane (storage control plane). It seems the right dimension to split the Make targets is by service. Can talk more details on Slack if needed.

cprice404 (approved):

PR #1412, opened 2024-08-16
chore: add web and nodejs sdk makefiles

malandis:
In order to standardize testing from the canaries, we add makefiles
with the standard per-service target names.


pgautier404 (approved):
🚢

PR #1413, opened 2024-08-19
feat: remove explicit references to `undefined` for better support of `null`

cprice404:
Users who are consuming our SDK directly from JavaScript (as opposed to TypeScript)
may attempt to pass in `null` in various places where we expose `undefined` in our
APIs. Based on the strict equality checking that we are doing (`===`), `null`s would
not be accepted in those places in the API and would result in errors for the users.

This commit applies some eslint rules that @tvald suggested, which forbid the explicit
use of `undefined` as a value, and require the less rigid `==` equality checks for `null`
(these checks will pass for both `null` and `undefined`).

This should allow JS users to pass in `null` for optional fields in our SDK APIs. I do not
believe there are any backward incompatible changes in this PR, nor any functional changes
other than being more tolerant of `null`.


packages/client-sdk-nodejs/src/config/middleware/cache-request-logging.ts
```diff
@@ -110,11 +110,11 @@ interface SetIfRequestLog extends WriteRequestLog {
...
   present: boolean;
-  presentAndNotEqual: string | undefined;
+  presentAndNotEqual: string | null;
   absent: boolean;
-  equal: string | undefined;
-  absentOrEqual: string | undefined;
-  notEqual: string | undefined;
+  equal: string | null;
+  absentOrEqual: string | null;
+  notEqual: string | null;
```
tvald, line 117: This is a breaking change if the interface is exposed externally. It should be fine to use either `<field>: string | undefined` or `<field>?: string`.

More generally, guidance should be to always use `undefined` as the type, `null` for comparison, and neither anywhere else. (Sorry, I know this is confusing.)

packages/client-sdk-nodejs/src/config/middleware/cache-request-logging.ts
```diff
@@ -366,7 +366,7 @@ const convertSetUnionRequest: RequestToLogInterfaceConverterFn<
 
 interface SetDifferenceRequestLog extends SetCollectionRequestLog {
   action: 'minuend' | 'subtrahend_set' | 'subtrahend_identity';
-  elements?: string[];
+  elements: string[] | null;
```
tvald, line 369: Similarly, this is a breaking change if the type is exposed, and we should prefer `?` notation.

packages/client-sdk-nodejs/src/config/middleware/cache-request-logging.ts
```diff
@@ -387,7 +387,7 @@ const convertSetDifferenceRequest: RequestToLogInterfaceConverterFn<
       ? request.subtrahend.set.elements.map(element =>
           convertBytesToString(element)
         )
-      : undefined,
+      : null,
```
tvald, line 390: 🤔  Ok, I need to revisit my thinking. This should 100% be `undefined`, not `null`.

packages/client-sdk-nodejs/test/unit/cache-service-error-mapper.test.ts
```diff
@@ -27,13 +27,19 @@ const generateServiceError = (status: Status): ServiceError => {
...
+
 describe('CacheServiceErrorMapper', () => {
   const cacheServiceErrorMapper = new CacheServiceErrorMapper(false);
   let resolved: SdkError | unknown;
   let rejected: SdkError | unknown;
   beforeEach(() => {
-    resolved = undefined;
-    rejected = undefined;
+    resolved = returnUndefined();
+    rejected = returnUndefined();
```
tvald, line 42: Hmm, this also feels like acrobatics that make me reconsider my stance.

packages/client-sdk-nodejs/src/config/middleware/cache-request-logging.ts
cprice404: This is not an interface anyone will have written any code against. This seemed like the path of least resistance for fixing this one.

packages/client-sdk-nodejs/src/config/middleware/cache-request-logging.ts
cprice404: same response for all of the stuff in this file - the stuff in this level is very low-level middleware stuff that I'm not expecting any users to touch directly.

packages/client-sdk-nodejs/src/config/middleware/cache-request-logging.ts
cprice404: lol. :)

Ternary operators won't let me omit the block after the `:`, and I can't use `undefined` there. This class/file isn't public enough to warrant too much churn but if you do have another idea how to solve this particular one lmk.

packages/client-sdk-nodejs/test/unit/cache-service-error-mapper.test.ts
cprice404: yeah it's janky, but in this case it's just some test setup logic.

cprice404:
not sure when this is going to be revisited, closing for now.

PR #1414, opened 2024-08-20
fix: switch to a jest reporter that shows correct error counts, avoid network calls in unit tests

cprice404:
This commit just does a few housekeeping things.

1. Switch from `jest-ci-spec-reporter` to `jest-spec-reporter`. They are similar,
   and neither of them is perfect, and neither of them has hardly any stars on github,
   but the `jest-spec-reporter` one prints the correct error counts at the end of the run.
   The `jest-ci-spec-reporter` one does not. (They both have the desired trait of using
   an output format that prints success/failure results for each test case on a single line,
   which makes it easier for operators to diagnose issues.
2. Unit tests should not make network calls. Our cache client unit tests were doing so. This
   changes them to prevent them from doing that.


rishtigupta (approved):
:shipit:

anitarua (approved):

cprice404:
@anitarua I will take a swing at updating the notion runbook to reflect this, and ping you to sanity check me if that's okay

PR #1418, opened 2024-08-22
feat: add sequence number to TopicItem and add optional onDiscontinuity callback

anitarua:
Work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/973

Tested new onDiscontinuity callback locally using a [stubbed topics service](https://github.com/momentohq/momento-local/pull/100) that returns only discontinuities.

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/pubsub-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/pubsub-client.ts
@@ -10,6 +10,7 @@ import {middlewaresInterceptor} from './grpc/middlewares-interceptor';
 import {
   CredentialProvider,
   StaticGrpcConfiguration,
+  TopicDiscontinuity,
   TopicGrpcConfiguration,
   TopicItem,
   TopicPublish,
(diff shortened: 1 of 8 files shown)


packages/core/src/internal/clients/pubsub/AbstractPubsubClient.ts
```diff
@@ -28,6 +29,7 @@ export interface SendSubscribeOptions {
     error: TopicSubscribe.Error,
     subscription: TopicSubscribe.Subscription
   ) => void;
+  onDiscontinuity?: (discontinuity: TopicDiscontinuity) => void;
```
cprice404, line 32: Since this is internal I think you could probably make it required like the other ones?

packages/core/src/internal/clients/pubsub/AbstractPubsubClient.ts
```diff
@@ -149,6 +151,7 @@ export abstract class AbstractPubsubClient<TGrpcError>
       topicName: topicName,
       onItem: onItem,
       onError: onError,
+      onDiscontinuity: options.onDiscontinuity,
```
cprice404, line 154: about 10 lines up from here is where we create the no-op implementations of these if the caller didn't pass one in, so that from this point forward they don't have to be nullable.

packages/core/src/messages/responses/topic-item.ts
```diff
@@ -14,10 +14,16 @@ export interface TopicItemOptions {
 export class TopicItem {
   private readonly _value: string | Uint8Array;
   private readonly _tokenId?: string;
+  private readonly _sequenceNumber: number;
 
-  constructor(_value: string | Uint8Array, options?: TopicItemOptions) {
+  constructor(
```
cprice404, line 19: This is technically a breaking change to this constructor but I don't expect anyone external to be constructing these, so I think what you've done here is probably the best way to go.

cprice404 (commented):
overall looks great, just a few nits/questions

packages/client-sdk-nodejs/src/internal/pubsub-client.ts
```diff
@@ -238,16 +239,24 @@ export class PubsubClient extends AbstractPubsubClient<ServiceError> {
             options.subscription
           );
         }
-      } else if (resp?.heartbeat) {
+      } else if (resp.heartbeat) {
```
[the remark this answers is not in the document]
anitarua: Yeah, I inspected the type (`_SubscriptionItem`) when I saw that and I think `resp` will never be undefined; just that it may not contain any data if something goes wrong, and that would be handled by the `else` case.

```typescript
        constructor(data?: any[] | ({} & (({
            item?: _TopicItem;
            discontinuity?: never;
            heartbeat?: never;
        } | {
            item?: never;
            discontinuity?: _Discontinuity;
            heartbeat?: never;
        } | {
            item?: never;
            discontinuity?: never;
            heartbeat?: _Heartbeat;
        }))));
```

PR #1419, opened 2024-08-22
feat: add default credential providers to all clients

anitarua:
Work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/975

Added a new function `getDefaultCredentialProvider` to the core package that's used in both node and web sdks when creating clients with optional credential providers (should be all 5 of them now: cache, topics, auth, leaderboard, storage)


packages/client-sdk-nodejs/src/index.ts
```diff
@@ -117,6 +117,7 @@ import {
   CredentialProvider,
   StringMomentoTokenProvider,
   EnvMomentoTokenProvider,
+  getDefaultCredentialProvider,
```
cprice404, line 120: Do we intend for this to be a user-facing function? If not, I think we might not want to export it from the nodejs/web sdk packages.

packages/core/src/index.ts
```diff
@@ -107,6 +107,7 @@ import {
   CredentialProvider,
   StringMomentoTokenProvider,
   EnvMomentoTokenProvider,
+  getDefaultCredentialProvider,
```
cprice404, line 110: it makes sense to export it here so the SDKs can use it.

cprice404 (commented):
just one nit about exports

packages/client-sdk-nodejs/src/index.ts
anitarua: oh probably not, will remove it from there

cprice404 (approved):
default

PR #1420, opened 2024-08-22
fix: stop leaking stores

anitarua:
Adds a `WithStore` integration test helper to avoid leaking stores.
Previously, we leaked a store with each run of the storage tests.


rishtigupta (approved):
:shipit:

PR #1423, opened 2024-08-27
chore: replacing more 'instanceof's in examples with 'switch'

cprice404:


examples/web/nextjs-chat/src/utils/momento-web.ts
```diff
@@ -5,8 +5,9 @@ import {
   TopicClient,
   type TopicItem,
   TopicPublish,
-  TopicSubscribe,
+  type TopicSubscribe,
 } from "@gomomento/sdk-web";
+import { TopicPublishResponse, TopicSubscribeResponse } from "@gomomento/sdk";
```
rishtigupta, line 10: This needs to be imported from web-sdk as well: https://github.com/momentohq/client-sdk-javascript/pull/1424/files

rishtigupta:
lgtm!

One question (nit): I noticed some places where we have access to logger but we are still using `console.log`/`console.info` where we can use `logger.info`?

cprice404:
> lgtm!
> 
> One question (nit): I noticed some places where we have access to logger but we are still using `console.log`/`console.info` where we can use `logger.info`?

hrm, good call. looks like that is always in those `ensureCacheExists` functions? Or did you see it somewhere else? I can fix those.

anitarua (approved):

PR #1428, opened 2024-08-28
chore: adding examples that are currently hard-coded in dev docs

cprice404:
This commit copies over a lot of example JS code that is currently
hard-coded in the dev docs, into example files here that we can
validate in CI. Also hopefully makes it easier to make sweeping
changes in the future (such as our instanceof migration).

A subsequent dev docs PR will consume these and get rid of the
hard-coded examples.


examples/nodejs/cache/doc-example-files/working-with-files.ts
```diff
@@ -0,0 +1,76 @@
+import * as fs from 'node:fs';
+import {CacheClient, CacheGetResponse, CacheSetResponse, Configurations, CredentialProvider} from '@gomomento/sdk';
+
```
malandis, line 3: Are we going to commit this file? Otherwise main just bails without error.

cprice404: ? sorry i'm not following your question

cprice404: oh you're asking about the json file. hrm let me look

cprice404: this example is really for the docs site, but i can add that file.

malandis: Woops. Yes I meant the json file. As is we can syntax check but if we run the program, without the file, the run function will return fast.

malandis (approved):

PR #1429, opened 2024-08-28
chore: update readme and examples for default credential provider and new topics callbacks

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/975

- Updates the basic cache, topics, and storage client examples to minimize code now that a default CredentialProvider is used.
- Updates the main repo README to include info about the default CredentialProvider and best practice 
- Adds example for adding optional `onDiscontinuity` and `onHeartbeat` topics callbacks

### DIFF
--- a/examples/nodejs/cache/doc-example-files/cheat-sheet-main.ts
+++ b/examples/nodejs/cache/doc-example-files/cheat-sheet-main.ts
@@ -1,9 +1,8 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import {CacheClient, Configurations} from '@gomomento/sdk';
+import {CacheClient} from '@gomomento/sdk';
 
 async function main() {
   const cacheClient = await CacheClient.create({
-    configuration: Configurations.Laptop.v1(),
     defaultTtlSeconds: 60,
   });
 }
--- a/examples/nodejs/cache/doc-example-files/cheat-sheet-main.ts
+++ b/examples/nodejs/cache/doc-example-files/cheat-sheet-main.ts
@@ -1,10 +1,9 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import {CacheClient, Configurations, CredentialProvider} from '@gomomento/sdk';
+import {CacheClient, Configurations} from '@gomomento/sdk';
 
 async function main() {
   const cacheClient = await CacheClient.create({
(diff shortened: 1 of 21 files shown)


examples/nodejs/cache/doc-example-files/cheat-sheet-main.ts
```diff
@@ -1,10 +1,9 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import {CacheClient, Configurations, CredentialProvider} from '@gomomento/sdk';
+import {CacheClient, Configurations} from '@gomomento/sdk';
 
 async function main() {
   const cacheClient = await CacheClient.create({
     configuration: Configurations.Laptop.v1(),
-    credentialProvider: CredentialProvider.fromEnvironmentVariable('MOMENTO_API_KEY'),
```
cprice404, line 7: `configuration` is actually optional now too; I just never went and updated the docs. (so thank you for being diligent about it!)

I think we should leave plenty of examples of how to explicitly set these so that people will know, but in the examples where we omit them, we should omit both. I think this file is a good place to omit both, and whatever shows up in the main README is another good spot.  Maybe also in the most basic example file for each product. But probably the rest of the places (e.g. the dictionary example) I'd leave the explicit versions in.

You can use your judgment on which files to update vs leave explicit, but let's go ahead and remove the explicit `configuration` option from any of the ones where you remove `credentialProvider`.

examples/nodejs/storage/basic.ts
```diff
@@ -1,17 +1,12 @@
...
   StorageGetResponse,
   StoragePutResponse,
 } from '@gomomento/sdk';
 
 async function main() {
-  const storageClient = new PreviewStorageClient({
-    configuration: StorageConfigurations.Laptop.latest(),
-    credentialProvider: CredentialProvider.fromEnvironmentVariable('MOMENTO_API_KEY'),
-  });
+  const storageClient = new PreviewStorageClient({});
```
anitarua, line 9: should the props object itself become optional then? 
removing the `{}` results in an `Expected 1 arguments, but got 0` error.

cprice404: good call, we could do that later, but I don't think it's a blocker for now.

cprice404 (approved):

PR #1433, opened 2024-09-05
chore: make sure storage tests use unique stores

anitarua:
Should use the `WithStore` helper function to create a unique store for each storage control plane test.

Creates single shared store for storage data plane tests using a `beforeAll` and is cleaned up using an `afterAll`.

Remove the "should successfully make two of the same requests after 5s retry timeout" test, which I added to test an edge case while developing the storage retry strategy, but it is no longer relevant after collapsing the fixed timeout interceptor into the retry interceptor.

### DIFF
--- a/packages/client-sdk-nodejs/test/integration/integration-setup.ts
+++ b/packages/client-sdk-nodejs/test/integration/integration-setup.ts
@@ -247,13 +247,10 @@ export function SetupTopicIntegrationTest(): {
 
 export function SetupStorageIntegrationTest(): {
   storageClient: PreviewStorageClient;
-  integrationTestStoreName: string;
 } {
-  const {integrationTestCacheName} = SetupIntegrationTest();
   const storageClient = momentoStorageClientForTesting();
   return {
     storageClient,
-    integrationTestStoreName: integrationTestCacheName,
   };
 }
 
--- a/packages/client-sdk-nodejs/test/integration/integration-setup.ts
+++ b/packages/client-sdk-nodejs/test/integration/integration-setup.ts
@@ -1,4 +1,9 @@
-import {testCacheName} from '@gomomento/common-integration-tests';
+import {
+  createStoreIfNotExists,
+  deleteStoreIfExists,
+  testCacheName,
+  testStoreName,
+} from '@gomomento/common-integration-tests';
 import {
   AuthClient,
   CreateCache,
@@ -247,10 +252,22 @@ export function SetupTopicIntegrationTest(): {
 
 export function SetupStorageIntegrationTest(): {
   storageClient: PreviewStorageClient;
+  integrationTestStoreName: string;
 } {
+  const integrationTestStoreName = testStoreName();
   const storageClient = momentoStorageClientForTesting();
+
+  beforeAll(async () => {
+    await createStoreIfNotExists(storageClient, integrationTestStoreName);
+  });
+
+  afterAll(async () => {
+    await deleteStoreIfExists(storageClient, integrationTestStoreName);
+  });
+
   return {
     storageClient,
+    integrationTestStoreName,
   };
 }
 
--- a/packages/common-integration-tests/src/storage/storage.ts
+++ b/packages/common-integration-tests/src/storage/storage.ts
@@ -14,12 +14,8 @@ import {
   WithStore,
 } from '../common-int-test-utils';
 import {v4} from 'uuid';
-import {sleep} from '@gomomento/sdk-core/dist/src/internal/utils';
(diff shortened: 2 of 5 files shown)


packages/common-integration-tests/src/storage/storage.ts
```diff
@@ -107,13 +103,14 @@ export function runStorageServiceTests(
   });
   describe('#store get put and delete', () => {
     it('put get and delete key in a store', async () => {
-      await WithStore(storageClient, testingStoreName, async () => {
+      const storeName = testStoreName();
+      await WithStore(storageClient, storeName, async () => {
```
cprice404, line 107: Creating and deleting stores is a fairly expensive operation, I'm worried that it might not be scalable to do that for every test. (Thoughts, @anp13 ?)

Zooming out a bit - what's the problem we're trying to solve that would preclude us from creating a single store to use for the whole test suite? potentially via `BeforeAll` and `AfterAll`?

anp13: I agree with you @cprice404. I would avoid create/deleting a store unless you are actually testing the functionality of creation/deletion of a store. Also we have seen ddb tables left behind when we timeout and deleting the store does not clean it up in the case where it is still stuck in creating state. So best to cautious with these operations.

I think the issue here was that we were doing a get operation on a store which was deleted in the previous test case. But that should not be a problem if the delete store test case is the last one which it is now since you removed the 5s timeout retry test case.

packages/client-sdk-nodejs/test/integration/integration-setup.ts
```diff
@@ -249,11 +254,20 @@ export function SetupStorageIntegrationTest(): {
...
+  const integrationTestStoreName = testStoreName();
   const storageClient = momentoStorageClientForTesting();
+
+  beforeAll(async () => {
+    await createStoreIfNotExists(storageClient, integrationTestStoreName);
+  });
+
+  afterAll(async () => {
+    await deleteStoreIfExists(storageClient, integrationTestStoreName);
+  });
```
anitarua, line 266: added a `beforeAll` and `afterAll` to create a shared store that can be used by storage data plane tests. this will prevent leaked stores and ensure minimal creation of stores where possible.

packages/common-integration-tests/src/storage/storage.ts
```diff
@@ -71,169 +70,26 @@ export function runStorageServiceTests(
...
-        const bytesValue = new Uint8Array([1, 2, 3, 4]);
-        const putBytesResponse = await storageClient.putBytes(
-          testingStoreName,
-          key,
-          bytesValue
-        );
-        switch (putBytesResponse.type) {
-          case StoragePutResponse.Success: {
+      // WithStore creates a store then deletes it
+      await WithStore(storageClient, storeName, async () => {
```
anitarua, line 74: used WithStore in this test to make it more concise, but it does the same thing as before (calls create, create, delete)

packages/common-integration-tests/src/storage/storage.ts
```diff
@@ -71,169 +70,26 @@ export function runStorageServiceTests(
...
-      await WithStore(storageClient, testingStoreName, async () => {
-        const key = v4();
-        const getResponse = await storageClient.get(testingStoreName, key);
-        expectWithMessage(() => {
-          expect(getResponse.type).toEqual(StorageGetResponse.NotFound);
-        }, `expected NotFound, received ${getResponse.toString()}`);
-        expect(getResponse.value()).toBeUndefined();
-      });
-    });
     it('should return store not found error for deleting a store that doesnt exist', async () => {
```
anitarua, line 93: moved 'should return store not found error for deleting a store that doesnt exist' to the storage control plane block of tests

packages/common-integration-tests/src/storage/storage.ts
```diff
@@ -253,21 +109,125 @@ export function runStorageServiceTests(
...
-        const getResponse1 = await storageClient.get(testingStoreName, key);
-        expectWithMessage(() => {
-          expect(getResponse1.type).toEqual(StorageGetResponse.NotFound);
-        }, `expected NotFound, received ${getResponse1.toString()}`);
+  });
 
-        await sleep(5000);
+  // Data plane tests should use the shared integration test store
+  // with the BeforeAll setup and AfterAll teardown
+  describe('#store get put and delete', () => {
```
anitarua, line 116: all the storage data plane tests should be in this describe block and should use the shared store with unique keys

cprice404 (approved):

PR #1439, opened 2024-09-26
chore: run tests on github and canaries with consistent reads

nand4011:
Enable consistent reads in the integration tests of the different packages if the useConsistentReads custom flag is detected.

Add new npm commands that run the tests with the consistent reads flag, and copies of the test scripts that call the npm commands.

Change the integration-test-cache npm command to use the consistent reads flag in the node and web packages so the canaries will have them enabled.

Use the new script in the github build scripts.

Add a check in the node and web packages to only send the read concern header it is configured to not be BALANCED, because balanced is equivalent to no header.


packages/client-sdk-web/src/internal/client-metadata-provider.ts
```diff
@@ -40,7 +40,8 @@ export class ClientMetadataProvider {
         ClientMetadataProvider.agentName
       }:${getBrowserName(navigator.userAgent)}`;
     }
-    if (this.readConcern) {
+    // Not sending a head concern header is treated the same as sending a BALANCED read concern header
+    if (this.readConcern && this.readConcern !== ReadConcern.BALANCED) {
```
malandis, line 44: 👍

malandis (approved):

anitarua (approved):

PR #1445, opened 2024-10-11
test: add negative test to detect replication failures

malandis:
Previously, our integration tests were written assuming consistent
reads. To avoid false positives caused by stale reads in low-latency
environments, we enforced consistent reads by setting the consistent
read header. While this reduced noise from false positives, it also
meant we lacked test coverage for replica reads and replication behavior.

To address this gap, we've added a negative test that detects
replication failures by verifying that no stale reads occur after the
expected replication delay.

The new test suite `replica-reads` implements a test that runs 10
trials in parallel. Each trial performs the following steps:
- Set: Writes a random key-value pair to the cache
- Wait: 1 second for reads to propagate
- Get: Reads the value for the key using `cacheClientWithBalancedReadConcern`.
- Assert: Confirms that the retrieved value matches the value set
earlier.

Trials are started with slight delays between them to distribute
load (delayBetweenTrials and a small random offset).

### DIFF
--- a/packages/common-integration-tests/src/cache/replica-reads.ts
+++ b/packages/common-integration-tests/src/cache/replica-reads.ts
@@ -15,43 +15,44 @@ export function runReplicaReadTests(
       const replicationDelayMs = 1000;
       const trials = [];
 
+      const trialFn = async (trialNumber: number) => {
+        // Start this trial at it's own time plus a random delay
+        const startDelay =
+          trialNumber * delayBetweenTrials + (Math.random() - 0.5) * 10;
+        await new Promise(resolve => setTimeout(resolve, startDelay));
+
+        const cacheKey = v4();
+        const cacheValue = v4();
+
+        // Perform a set operation
+        const setResponse = await client.set(
+          integrationTestCacheName,
+          cacheKey,
+          cacheValue
+        );
+        expectWithMessage(() => {
+          expect(setResponse).toBeInstanceOf(CacheSet.Success);
+        }, `expected SUCCESS but got ${setResponse.toString()}`);
+
+        // Wait for replication to complete
+        await new Promise(resolve => setTimeout(resolve, replicationDelayMs));
+
+        // Verify that the value can be read
+        const getResponse = await client.get(
+          integrationTestCacheName,
+          cacheKey
+        );
+        expectWithMessage(() => {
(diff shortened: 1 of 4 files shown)


packages/common-integration-tests/src/cache/replica-reads.ts
```diff
@@ -0,0 +1,62 @@
...
+          // Verify that the value can be read
+          const getResponse = await client.get(
+            integrationTestCacheName,
+            cacheKey
+          );
+          expectWithMessage(() => {
+            expect(getResponse).toBeInstanceOf(CacheGet.Hit);
+          }, `expected HIT but got ${getResponse.toString()}`);
+
+          if (getResponse instanceof CacheGet.Hit) {
```
anitarua, line 50: lgtm overall!

but would it be better to check using `getResponse.type` here instead of `instanceof`?

malandis: Fixed in [f2ab706](https://github.com/momentohq/client-sdk-javascript/pull/1445/commits/f2ab706874e84e63a2f7239c123a5e14d8b252b9)

anitarua (dismissed):
lgtm!

packages/common-integration-tests/src/cache/replica-reads.ts
```diff
@@ -0,0 +1,60 @@
...
+          }, `expected HIT but got ${getResponse.toString()}`);
+
+          expectWithMessage(() => {
+            expect(getResponse.value()).toEqual(cacheValue);
+          }, `expected ${cacheValue} but got ${getResponse.value() ?? 'undefined'}`);
+        })();
+        trials.push(trialPromise);
+      }
+
+      await Promise.all(trials);
```
krispraws, line 57: Just to confirm, the success criteria for this test is that **all** the trials should return a hit with the expected value?

krispraws (dismissed):
Looks good, left a question for clarification.

packages/common-integration-tests/src/cache/replica-reads.ts
malandis: That's correct

krispraws: Thanks!

anitarua (approved):

PR #1447, opened 2024-10-15
fix: add NotSet path for increase/decrease ttl apis

rishtigupta:
## PR Description:
- Add the "NotSet" path in increase/decrease ttl apis

## Issue
closes https://github.com/momentohq/client-sdk-javascript/issues/1446


anitarua (approved):

malandis (commented):
We'll also need tests to cover this path

rishtigupta:
> We'll also need tests to cover this path

The tests already existed with check for an Error instead of NotSet path. Updated the tests.

malandis (approved):

PR #1448, opened 2024-10-22
chore: clarify limit exceeded in resource exhausted errors

anitarua:
Work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/982

Removes the confusing preamble from the RESOURCE_EXHAUSTED error message. Makes more clear which limit was exceeded if that information is available in the grpc error details or `err` metadata field. 

Using the `err` metadata is preferred so that we don't rely on basic string matching if possible as error details can change any time in the future.

If all else fails, we return a generic error message: `'Limit exceeded for this account'`


rishtigupta (approved):
lgtm :shipit:

PR #1458, opened 2024-10-28
chore: use unique topic names in tests

anitarua:
While investigating some canary failures, discovered that some auth topics tests were reusing topic names.
Added a new util function to ensure unique topic names are used in all topics tests.


kvcache (approved):
seems legit to me

rishtigupta (approved):
:shipit:

PR #1463, opened 2024-11-06
feat: support topics sequence page

anitarua:
Work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/1059

Updates the topic subscription to take the sequence page into account:

- Keeps track of last known sequence page in addition to sequence number in `SubscriptionState` so we can reconnect gracefully
- Includes sequence page detail for `TopicDiscontinuity` objects

As part of this change, had to update the client-protos dependency version as well.

Tested locally to verify both node and web clients gracefully reconnect when wifi is toggled and sequence page is used to resume from the next expected sequence number.
Also tested locally and confirmed that subscribing with an initial (arbitrary) sequence page will cause a discontinuity before getting assigned a new sequence page.

Also had forgotten to export `TopicDiscontinuity` in the web sdk in a previous PR, so that was added here as well.

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/pubsub-client.ts
+++ b/packages/client-sdk-nodejs/src/internal/pubsub-client.ts
@@ -169,11 +169,13 @@ export class PubsubClient extends AbstractPubsubClient<ServiceError> {
       topic: options.topicName,
       resume_at_topic_sequence_number:
         options.subscriptionState.resumeAtTopicSequenceNumber,
+      sequence_page: options.subscriptionState.lastTopicSequencePage,
     });
 
     this.getLogger().trace(
-      'Subscribing to topic with resume_at_topic_sequence_number: %s',
-      options.subscriptionState.resumeAtTopicSequenceNumber
+      'Subscribing to topic with resume_at_topic_sequence_number %s and sequence_page %s',
+      options.subscriptionState.resumeAtTopicSequenceNumber,
+      options.subscriptionState.lastTopicSequencePage ?? 'undefined'
     );
     const call = this.client.Subscribe(request, {
       interceptors: this.streamingInterceptors,
@@ -210,11 +212,14 @@ export class PubsubClient extends AbstractPubsubClient<ServiceError> {
 
       if (resp.item) {
         const sequenceNumber = resp.item.topic_sequence_number;
+        const sequencePage = resp.item.sequence_page;
         options.subscriptionState.lastTopicSequenceNumber = sequenceNumber;
+        options.subscriptionState.lastTopicSequencePage = sequencePage;
         this.getLogger().trace(
-          'Received an item on subscription stream; topic: %s; sequence number: %s',
+          'Received an item on subscription stream; topic: %s; sequence number: %s; sequence page: %s',
           truncateString(options.topicName),
-          sequenceNumber
+          sequenceNumber,
+          sequencePage
         );
         if (resp.item.value.text) {
           options.onItem(
@@ -254,7 +259,8 @@ export class PubsubClient extends AbstractPubsubClient<ServiceError> {
         options.onDiscontinuity(
           new TopicDiscontinuity(
             resp.discontinuity.last_topic_sequence,
-            resp.discontinuity.new_topic_sequence
+            resp.discontinuity.new_topic_sequence,
+            resp.discontinuity.new_sequence_page
           )
         );
       } else {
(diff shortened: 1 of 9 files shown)


rishtigupta (commented):
lgtm! :shipit:

packages/core/src/messages/responses/topic-discontinuity.ts
```diff
@@ -28,6 +34,14 @@ export class TopicDiscontinuity {
...
+  /**
+   * Returns the new sequence page after the discontinuity.
+   * @returns number
+   */
+  public newSequencePage(): number {
+    return this._newSequencePage;
+  }
+
   public toString(): string {
     const displayValue = `Last Sequence Number: ${this._lastSequenceNumber}; New Sequence Number: ${this._newSequenceNumber}`;
```
malandis, line 46: We should also add the sequence page to the string representation. This can be really helpful for debugging.

packages/client-sdk-web/src/index.ts
```diff
@@ -284,6 +285,7 @@ export {
   TopicClientConfiguration,
   TopicClient,
   TopicItem,
+  TopicDiscontinuity,
```
malandis, line 288: Good catch!

packages/client-sdk-web/src/internal/pubsub-client.ts
```diff
@@ -134,6 +134,15 @@ export class PubsubClient<
...
     );
+    if (options.subscriptionState.lastTopicSequencePage !== undefined) {
+      request.setSequencePage(options.subscriptionState.lastTopicSequencePage);
+    }
+
+    this.getLogger().trace(
+      'Subscribing to topic with resume_at_topic_sequence_number %s and sequence_page %s',
+      options.subscriptionState.resumeAtTopicSequenceNumber,
+      options.subscriptionState.lastTopicSequencePage ?? 'undefined'
+    );
```
malandis, line 145: I recommend adding a property accessor `resumeAtTopicSequencePage` on subscription state class which defaults to 0. That way we can avoid the null checks here and throughout (see `public get resumeAtTopicSequenceNumber` for reference.)

packages/client-sdk-web/src/internal/pubsub-client.ts
```diff
@@ -211,13 +228,16 @@ export class PubsubClient<
         options.onHeartbeat(new TopicHeartbeat());
       } else if (discontinuity) {
         this.getLogger().trace(
-          'Received discontinuity from subscription stream; topic: %s',
-          truncateString(options.topicName)
+          'Received a discontinuity; topic: %s; new sequence number: %s; new sequence page: %s',
+          truncateString(options.topicName),
+          discontinuity.getNewTopicSequence(),
+          discontinuity.getNewSequencePage()
```
malandis, line 234: Could also use the discontinuity `toString` here

packages/client-sdk-nodejs/src/internal/pubsub-client.ts
```diff
@@ -169,11 +169,13 @@ export class PubsubClient extends AbstractPubsubClient<ServiceError> {
...
         options.subscriptionState.resumeAtTopicSequenceNumber,
+      sequence_page: options.subscriptionState.lastTopicSequencePage,
     });
 
     this.getLogger().trace(
-      'Subscribing to topic with resume_at_topic_sequence_number: %s',
-      options.subscriptionState.resumeAtTopicSequenceNumber
+      'Subscribing to topic with resume_at_topic_sequence_number %s and sequence_page %s',
+      options.subscriptionState.resumeAtTopicSequenceNumber,
+      options.subscriptionState.lastTopicSequencePage ?? 'undefined'
```
malandis, line 178: See also below my comment about how `resumeAtTopicSequenceNumber` is implemented. We can do the same for `resumeAtTopicSequencePage` to avoid null checks.

malandis (commented):
Logic looks good to me. Couple suggestions to avoid null checks in the code, otherwise gtg!

packages/client-sdk-web/src/internal/pubsub-client.ts
```diff
@@ -134,6 +134,13 @@ export class PubsubClient<
     request.setResumeAtTopicSequenceNumber(
       options.subscriptionState.resumeAtTopicSequenceNumber
     );
+    request.setSequencePage(
+      options.subscriptionState.resumeAtTopicSequencePage
+    );
+
+    this.getLogger().trace(
+      `Subscribing to topic with resume_at_topic_sequence_number ${options.subscriptionState.resumeAtTopicSequenceNumber} and sequence_page ${options.subscriptionState.resumeAtTopicSequencePage}`
+    );
```
anitarua, line 143: while testing locally, I noticed the logs coming from the web sdk looked poorly formatted, reverted to using the string interpolation instead

```
[2024-11-08T18:08:38.251Z] TRACE (Momento: PubsubClient): Issuing publish request; topic: %s, message length: %stopic-web,15
[2024-11-08T18:08:38.436Z] TRACE (Momento: PubsubClient): Received an item on subscription stream; topic: %s; sequence number: %s; sequence page: %stopic-web,3,1731089315
```

malandis (approved):

PR #1467, opened 2024-11-13
chore: change cancelled error message to remove the word server

pratik151192:
Changing cancelled error messages. Went back and forth on the message, and made it simple to say was `cancelled` without assigning the "blame" to anyone. Although gRPC [status code](https://grpc.io/docs/guides/status-codes/) says `The operation was cancelled, typically by the caller.`


rishtigupta (approved):
:shipit:

nand4011 (dismissed):

malandis (commented):
I would change this to only contact us if this was unexpected. If they are using a cancellation token or other client-side cancellation mechanism, then it's expected.

rishtigupta (approved):
:shipit:

PR #1471, opened 2024-11-19
feat: add http api tests to nodejs sdk

rishtigupta:
## PR Description:
- Add HTTP API tests to NodeJS and Web SDK for canaries. 
- Fix the copyright in LICENSE file

## Issue
https://github.com/momentohq/dev-eco-issue-tracker/issues/1068

### DIFF
--- a/packages/client-sdk-nodejs/package.json
+++ b/packages/client-sdk-nodejs/package.json
@@ -16,6 +16,7 @@
     "prebuild": "eslint . --ext .ts",
     "test": "jest --testPathIgnorePatterns auth-client.test.ts --maxWorkers 1",
     "integration-test-auth": "jest auth/ --maxWorkers 1 -- useConsistentReads",
+    "integration-test-http": "jest http/ --maxWorkers 1 -- useConsistentReads",
     "integration-test-cache": "jest cache/ --maxWorkers 1 -- useConsistentReads",
     "integration-test-control-cache-topics": "npm run integration-test-cache && npm run integration-test-topics",
     "integration-test-leaderboard": "jest leaderboard/ --maxWorkers 1 -- useConsistentReads",
--- a/packages/client-sdk-nodejs/package.json
+++ b/packages/client-sdk-nodejs/package.json
@@ -14,7 +14,7 @@
   },
   "scripts": {
     "prebuild": "eslint . --ext .ts",
-    "test": "jest --testPathIgnorePatterns auth-client.test.ts --maxWorkers 1",
+    "test": "jest --testPathIgnorePatterns=\"auth-client\\.test\\.ts|http-apis\\.test\\.ts\" --maxWorkers 1",
     "integration-test-auth": "jest auth/ --maxWorkers 1 -- useConsistentReads",
     "integration-test-http": "jest http/ --maxWorkers 1 -- useConsistentReads",
     "integration-test-cache": "jest cache/ --maxWorkers 1 -- useConsistentReads",
@@ -27,7 +27,7 @@
     "integration-test": "jest integration --maxWorkers 1",
     "unit-test": "jest unit",
     "build-deps": "cd ../core && npm run build && cd - && cd ../common-integration-tests && npm run build && cd -",
-    "build-and-run-tests": "npm run build-deps && jest --testPathIgnorePatterns auth-client.test.ts --maxWorkers 1",
+    "build-and-run-tests": "npm run build-deps && jest --testPathIgnorePatterns=\"auth-client\\.test\\.ts|http-apis\\.test\\.ts\" --maxWorkers 1",
     "lint": "eslint . --ext .ts",
     "format": "eslint . --ext .ts --fix",
     "watch": "tsc -w",
--- a/packages/client-sdk-nodejs/test/integration/shared/http/http-apis.test.ts
+++ b/packages/client-sdk-nodejs/test/integration/shared/http/http-apis.test.ts
@@ -1,0 +1,7 @@
+import {SetupAuthClientIntegrationTest} from '../../integration-setup';
+import {runHttpApiTest} from '@gomomento/common-integration-tests/dist/src/http/http-apis';
+
+const {mgaAccountSessionTokenAuthClient, cacheName} =
+  SetupAuthClientIntegrationTest();
+
+runHttpApiTest(mgaAccountSessionTokenAuthClient, cacheName);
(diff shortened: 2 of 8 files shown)


packages/common-integration-tests/src/http/http-apis.ts
```diff
@@ -0,0 +1,139 @@
...
+      expect(getRes.statusCode).toBe(200);
+
+      // Use DELETE API to update the value
+      const delRes = await deleteValue(baseUrl, apiKey, cacheName, key);
+      expect(delRes.statusCode).toBe(204);
+
+      // Use GET API to retrieve the value
+      const getRes2 = await getValue(baseUrl, apiKey, cacheName, key);
+      expect(getRes2.statusCode).toBe(404);
+    });
```
malandis, line 137: We should add in some more checks to be comprehensive:
- what is the expected result when we get a value that is not there?
- do we get the expected value back on a key that exists?
- what is the expected result when deleting a key that is not there?

packages/common-integration-tests/src/http/http-apis.ts
```diff
@@ -0,0 +1,139 @@
...
+) {
+  describe('Momento HTTP API', () => {
+    let apiKey: string;
+    let baseUrl: string;
+
+    beforeAll(async () => {
+      const resp = await sessionTokenAuthClient.generateApiKey(
+        SUPER_USER_PERMISSIONS,
+        ExpiresIn.minutes(5)
+      );
```
malandis, line 103: Why are we using this as opposed to the token we use for other tests?

packages/common-integration-tests/src/http/http-apis.ts
```diff
@@ -0,0 +1,139 @@
...
+function putValue(
+  baseUrl: string,
+  apiKey: string,
+  cacheName: string,
+  key: string,
+  value: string,
+  ttl: number
+): Promise<IncomingMessage> {
+  const encodedKey = encodeURIComponent(key);
+  const path = `/cache/${cacheName}?key=${encodedKey}&value=${value}&ttl_seconds=${ttl}`;
```
malandis, line 50: Do we not need to encode the cache name and value as well?

packages/client-sdk-nodejs/test/integration/shared/http/http-apis.test.ts
```diff
@@ -0,0 +1,7 @@
+import {SetupAuthClientIntegrationTest} from '../../integration-setup';
+import {runHttpApiTest} from '@gomomento/common-integration-tests/dist/src/http/http-apis';
+
+const {mgaAccountSessionTokenAuthClient, cacheName} =
+  SetupAuthClientIntegrationTest();
```
malandis, line 5: Are these tests specific to mga account sessions?

malandis (commented):
Couple questions:
- Is there no need to run these tests on the web sdk? I see we added a script for the nodejs but not web.
- Were the updates to the package locks necessary for this PR? If not let's revert those and if we need to bump dependencies can do in a new PR

packages/common-integration-tests/src/http/http-apis.ts
rishtigupta: I used this token to get the api endpoint that will be used in creating the url for invoking the http api.

packages/client-sdk-nodejs/test/integration/shared/http/http-apis.test.ts
rishtigupta: Not really, but the only way I could get the api-endpoint is by creating a new token using the mga account session token and then calculating the api endpoint from it.

malandis: Is the MGA Account Session token a customer-facing concern? Ie can a customer run these integration tests?

malandis (commented):
Just a question about the MGA token and if that's a customer facing path to running the tests.

packages/client-sdk-nodejs/test/integration/shared/http/http-apis.test.ts
rishtigupta: The customer won't be able to run these tests since they do not have any way to access mga's account session token.

nand4011:
>     * Is there no need to run these tests on the web sdk? I see we added a script for the nodejs but not web.

Since the http api isn't a part of any of the SDKs, these are in the javascript repo mostly for convenience. I don't think they ever need to be run locally. Would it make more sense to use a separate module so they don't run when running the JS integration tests?

rishtigupta:
> > ```
> > * Is there no need to run these tests on the web sdk? I see we added a script for the nodejs but not web.
> > ```
> 
> Since the http api isn't a part of any of the SDKs, these are in the javascript repo mostly for convenience. I don't think they ever need to be run locally. Would it make more sense to use a separate module so they don't run when running the JS integration tests?

We use this script to ignore auth client tests when running JS integ test suite:
```
    "build-and-run-tests": "npm run build-deps && jest --testPathIgnorePatterns auth-client.test.ts --maxWorkers 1",
```
Does that work if we add http tests to the ignore path in the above script?

packages/client-sdk-nodejs/package.json
```diff
@@ -14,8 +14,9 @@
   },
   "scripts": {
     "prebuild": "eslint . --ext .ts",
-    "test": "jest --testPathIgnorePatterns auth-client.test.ts --maxWorkers 1",
+    "test": "jest --testPathIgnorePatterns=\"auth-client\\.test\\.ts|http-apis\\.test\\.ts\" --maxWorkers 1",
```
anitarua, line 17: is this the right syntax? I think [Jest CLI](https://jest-archive-august-2023.netlify.app/docs/27.x/cli#--testpathignorepatternsregexarray) lets you pass in an array

packages/common-integration-tests/src/http/http-apis.ts
```diff
@@ -0,0 +1,216 @@
...
+
+    beforeAll(async () => {
+      const resp = await sessionTokenAuthClient.generateApiKey(
+        SUPER_USER_PERMISSIONS,
+        ExpiresIn.minutes(5)
+      );
+      expect(resp).toBeInstanceOf(GenerateApiKey.Success);
+      const successResp = resp as GenerateApiKey.Success;
+      apiKey = successResp.apiKey;
+      baseUrl = `api.cache.${successResp.endpoint}`;
```
anitarua, line 127: I might be overlooking something, but would it be possible to call `credentialProvider.getCacheEndpoint()` or similar here in order to construct the endpoint instead of making a new API key each time?

rishtigupta: it's in the beforeAll hook, so should be called once?

anitarua (approved):
lgtm, though not sure if @malandis had more mga-related follow up questions

malandis (approved):

PR #1472, opened 2024-11-26
chore: add http api tests to makefile

rishtigupta:
## PR Description:
To add http api tests to canaries, we use the make file base bath defined here: https://github.com/momentohq/canaries-kitchensink/blob/main/canaries/sdk-canary/index.ts#L79-L88. 

Therefore, this commit adds http-tests to makefile for both nodejs and web sdks

## Issue
https://github.com/momentohq/dev-eco-issue-tracker/issues/1068


anitarua (approved):

malandis (approved):
LGTM. Note that to run this from the canary, we'll have to add trivial make targets to the .NET and Go SDKs as well. That is, those Makefiles will have a `test-http-service` target that doesn't do anything.

PR #1484, opened 2025-01-03
feat: deprecate withMomentoLocal override in favor of new static method forMomentoLocal

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1090

Deprecates `withMomentoLocal()` which overrode the endpoints on a `StringTokenMomentoProvider`, but route requires a user pass in a valid API key before overriding the endpoints for momento-local, which doesn't require an API key.

Creates a new static method on the `CredentialProvider` instead, `forMomentoLocal()`, that creates a new instance of `MomentoLocalProvider` which does not require an API key. `MomentoLocalProvider` will now accept configurable hostname and port number as well, but defaults to `127.0.0.1:8080`, which is what the old method used.


rishtigupta (approved):
love it :shipit:

PR #1485, opened 2025-01-07
ci: check permissions of triggering actor before running PR checks

anitarua:
Addresses https://github.com/momentohq/dev-eco-issue-tracker/issues/1096

For PR checks that need an auth token to run successfully, we first check that the github actor that triggered the workflow has write permissions to the repo before sharing the secrets and running the tests.

This should allow us the ability to approve workflows on PRs that come in via forked branches, making it easier to accept open-source contributions.

**Intended behavior**

1. PR from fork comes in, requires workflow approval from a maintainer
2. Maintainer approves, starts workflow, becomes the "triggering actor"
3. GitHub Actions workflow checks triggering actor's permissions, proceeds if they have at least "write" access to the repo
4. PR checks run with access to secrets due to usage of `on: pull_request_target`

PRs from a maintainer should run the PR checks as usual because they are the ones that trigger the workflow. Workflow approval can be configured to be required on all PRs from forks.

[Reference article for the workflow](https://michaelheap.com/access-secrets-from-forks/#check-collaborator-permissions)

[Reference article for `on: pull_request` vs `on: pull_request_target`](https://runs-on.com/github-actions/pull-request-vs-pull-request-target/)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,6 +9,25 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  verify-user-permissions:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get User Permission
+        id: checkAccess
+        uses: actions-cool/check-user-permission@v2
+        with:
+          require: write
+          username: ${{ github.triggering_actor }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Check User Permission
+        if: steps.checkAccess.outputs.require-result == 'false'
+        run: |
+          echo "${{ github.triggering_actor }} does not have permissions on this repo."
+          echo "Current permission level is ${{ steps.checkAccess.outputs.user-permission }}"
+          echo "Job originally triggered by ${{ github.actor }}"
+          exit 1
+
   verify:
     runs-on: ubuntu-24.04
     steps:
@@ -359,6 +378,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   cache-tests:
+    needs: verify-user-permissions
     strategy:
       matrix:
         node: [18]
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -372,6 +392,8 @@ jobs:
     steps:
       - name: Setup repo
         uses: actions/checkout@v3
+        with:
+          ref: ${{  github.event.pull_request.head.sha }} # This is dangerous without the first access check
```
malandis, line 396: Why do we need this?

.github/workflows/build.yml
```diff
@@ -9,6 +9,25 @@ concurrency:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get User Permission
+        id: checkAccess
+        uses: actions-cool/check-user-permission@v2
+        with:
+          require: write
+          username: ${{ github.triggering_actor }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
malandis, line 22: Is this necessary?

malandis (commented):
Couple questions about the implementation. Also can you explain the flow for someone submitting a PR via a fork vs directly to the repo?

.github/workflows/build.yml
anitarua: It was included in the [reference article](https://michaelheap.com/access-secrets-from-forks/), but based on the action's [homepage](https://github.com/marketplace/actions/check-user-permission), I think perhaps not. I'll remove it and see if it still runs without.

.github/workflows/build.yml
anitarua: Was also following the reference article here, but perhaps not necessary either. Will remove and see what happens as well

anitarua:
Going to take Pete's suggestion and test this more thoroughly on a sample repo before proceeding here

pgautier404 (approved):

PR #1488, opened 2025-01-07
fix: upgrade to open pull request v7 for correct attribution

malandis:
The commit attribution is wrong when we run the update-dependencies
workflow. While it should only be committing using the bot account, an
individual developer is also tagged on the commit. This is incorrect.

The create-pull-request GitHub action fixed the attribution for the
commits in v6. We upgrade to v7 for the latest.


anitarua (approved):

rishtigupta (approved):
:shipit:

PR #1494, opened 2025-01-10
fix: copy constructors should use spread operator

malandis:
# Overview
In order to make our config copy constructors maintainable and less
error-prone, we should use the spread operator over `this`.

Previously in the copy constuctors we built a copy of the
corresponding properties by hand. Because some properties are
optional, and because we add new properties over time, we introduced
drift between the new properties and various copy constructors.

That is, when we added new properties, some copy constructors were out
of date and hence falling back to defaults instead of copying those
new properties over.

# Sample affected code

From the cache static transport strategy:
```typescript
  withNumClients(numClients: number): GrpcConfiguration {
    return new StaticGrpcConfiguration({
      deadlineMillis: this.deadlineMillis,
      maxSessionMemoryMb: this.maxSessionMemoryMb,
      numClients: numClients,
    });
  }
```

At a glance this may look correct, but we had since added 6 other properties 
which are not being copied over. That means customers using non-default
values for those properties will have them erased on a copy.

# Fix
We can fix this across the board by:
1. aligning the field names in each config class with the corresponding
property name, and
2. building the props copy using a spread over `this` followed by the
new value.

# Example fix
```typescript
interface ThingProps {
    field1: number;
    field2: number;
    field3: number;
}

class Thing {
    private readonly field1: number;
    private readonly field2: number;
    private readonly field3: number;

    constructor(props: ThingProps) {
        this.field1 = props.field1;
        this.field2 = props.field2;
        this.field3 = props.field3;
    }

    getField1(): number {
        return this.field1;
    }

    withField1(field1: number): Thing {
        return new Thing({
            ...this,
            field1
        })
    }

    getField2(): number {
        return this.field2;
    }

    getField3(): number {
        return this.field3;
    }

    toString(): string {
        return `${this.field1}, ${this.field2}, ${this.field3}`;
    }
}

const thing = new Thing({field1: 1, field2: 2, field3: 3});
console.log(thing.toString()); // 1, 2, 3

const thing2 = thing.withField1(5);
console.log(thing2.toString()); // 5, 2, 3
```

### DIFF
--- a/packages/client-sdk-nodejs/src/config/configuration.ts
+++ b/packages/client-sdk-nodejs/src/config/configuration.ts
@@ -183,13 +183,8 @@ export class CacheConfiguration implements Configuration {
 
   withRetryStrategy(retryStrategy: RetryStrategy): Configuration {
     return new CacheConfiguration({
-      loggerFactory: this.loggerFactory,
-      retryStrategy: retryStrategy,
-      transportStrategy: this.transportStrategy,
-      middlewares: this.middlewares,
-      throwOnErrors: this.throwOnErrors,
-      readConcern: this.readConcern,
-      compression: this.compression,
+      ...this,
+      retryStrategy,
     });
   }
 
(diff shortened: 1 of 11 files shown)


anitarua (approved):
nice catch!

PR #1500, opened 2025-01-13
feat: add retry metrics middleware

rishtigupta:
## PR Description:
- ~~Had to modify the requestType from MomentoRPCMethod enum to string in oprder to support the requestType from the middleware (which we extract from the request body)~~
- Implement the RetryTestMetricsMiddleware
- Add unit test

## Issue
closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1092


packages/client-sdk-nodejs/test/unit/test-retry-metrics-middleware.test.ts
```diff
@@ -0,0 +1,43 @@
...
+import {CredentialProvider, MomentoLogger} from '@gomomento/sdk-core';
+import {CacheClient, Configurations} from '../../src';
+
+describe('TestRetryMetricsMiddleware', () => {
+  let middleware: TestRetryMetricsMiddleware;
+  let cacheClient: CacheClient;
+  let testMetricsCollector: TestRetryMetricsCollector;
+  let momentoLogger: MomentoLogger;
+
+  beforeEach(async () => {
```
anitarua, line 12: should this be `beforeAll` so we create only one cache client for the whole test suite? I'm not sure if we'll write more tests in this suite though

packages/client-sdk-nodejs/test/unit/test-retry-metrics-middleware.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    // Create CacheClient with middleware enabled
+    cacheClient = await CacheClient.create({
+      configuration: Configurations.Laptop.v1().withMiddlewares([middleware]),
+      credentialProvider:
+        CredentialProvider.fromEnvironmentVariable('MOMENTO_API_KEY'),
+      defaultTtlSeconds: 60,
+    });
+  });
+
+  test('should add a timestamp on request body when cache name is available', async () => {
```
anitarua, line 32: could we test that metrics can be retrieved for different caches too?

packages/client-sdk-nodejs/test/integration/test-retry-metrics-middleware.ts
```diff
@@ -0,0 +1,86 @@
...
+    metadata: MiddlewareMetadata
+  ): Promise<MiddlewareMetadata> {
+    return Promise.resolve(metadata);
+  }
+
+  onResponseBody(
+    response: MiddlewareMessage | null
+  ): Promise<MiddlewareMessage | null> {
+    return Promise.resolve(response);
+  }
```
anitarua, line 61: hadn't occurred to me before, but we could also track received timestamps if it would be useful for verifying retry strategy behavior when we write the tests

doesn't have to be implemented in this pr though!

anitarua (commented):
looking good! had some suggestions for the tests

you mentioned revising the enum in a follow-up, I think that would still be nice to have once we write the retry tests 👍

packages/client-sdk-nodejs/test/integration/test-retry-metrics-collector.ts
```diff
@@ -1,35 +1,3 @@
-export enum MomentoRPCMethod {
```
malandis, line 1: I feel like this approach would be more maintainable. Can we discuss the blocker you ran in to? Probably easier synchronously.

rishtigupta: yes sure. 
I will also try and implement the MomentoRPCMethod enum in the same PR till tomorrow for better discussion tomorrow.

packages/client-sdk-nodejs/test/unit/test-retry-metrics-middleware.test.ts
rishtigupta: using `beforeAll` conflicts with the deep equality in expect methods. Therefore, to keep each test independent of each other, we would need to use `beforeEach`

anitarua (dismissed):
lgtm!

malandis (commented):
This is much better thanks. Just a couple minor changes and GTG.

packages/client-sdk-nodejs/test/integration/momento-rpc-method.ts
```diff
@@ -0,0 +1,31 @@
+export enum MomentoRPCMethod {
```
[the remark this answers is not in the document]
rishtigupta: Actually that is true for metrics collector and middleware too. Will move those too.

malandis (commented):
Very nice!

anitarua (approved):

PR #1509, opened 2025-01-21
feat: add automated retry tests to ci

rishtigupta:
## PR Description:
- Add automated retry tests to ci

## Issue
https://github.com/momentohq/dev-eco-issue-tracker/issues/1095


anitarua (commented):
I'm a little surprised to see the new tests run in PR given the weirdness with `pull_request_target` recently 😅 but I guess it's because this PR doesn't update the github actions workflow itself, just the script.

It seems like these tests are run serially with the other cache tests, which I think slows down them down, especially when spinning docker containers up/down.

These tests don't need an API key so I think it would be better if they ran in parallel as a separate github actions job and using different docker ports.

I wonder if you can make a new github actions file that uses `on: pull_request` instead of `on: pull_request_target` to avoid some of the weirdness too? 
And I think the scripts look good, but instead of calling them from `scripts/build-and-test-all-packages-consistent-reads.sh`, call them from a new `.github/workflows/retry-tests.yml` or something.

scripts/build-and-test-package-full-network-outage-retry-tests.sh
```diff
@@ -0,0 +1,27 @@
...
+docker run --name ${DOCKER_CONTAINER_NAME} -d -p 8080:8080 gomomento/momento-local \
+    --return-error unavailable --error-rpcs get
+
+# Ensure the container is stopped when the script exits
+function cleanup {
+    echo "Stopping and removing Docker container..."
+    docker stop ${DOCKER_CONTAINER_NAME}
+    docker rm ${DOCKER_CONTAINER_NAME}
+}
+trap cleanup EXIT
```
malandis, line 21: In Github Actions, the native way of running this is to set the services property in the job. See [here](https://github.com/momentohq/momento-java-lettuce-client/blob/main/.github/workflows/on-pull-request.yml#L27-L31). We can set the various docker settings there as well.

We will need a script like this as a prerequisite to run locally.

.github/workflows/retry-tests.yml
```diff
@@ -0,0 +1,89 @@
...
+          echo "${{ github.triggering_actor }} does not have permissions on this repo."
+          echo "Current permission level is ${{ steps.checkAccess.outputs.user-permission }}"
+          echo "Job originally triggered by ${{ github.actor }}"
+          exit 1
+
+  test-momento-local-1:
+    needs: verify-user-permissions
+    runs-on: ubuntu-latest
+    env:
+      MOMENTO_API_KEY: ${{ secrets.ALPHA_TEST_AUTH_TOKEN }}
```
anitarua, line 29: momento-local doesn't need api key, can remove this

.github/workflows/retry-tests.yml
```diff
@@ -0,0 +1,89 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
anitarua, line 34: we're using `on: pull_request` in this file so we're good to remove this and let `actions/checkout` use the default ref

.github/workflows/retry-tests.yml
```diff
@@ -0,0 +1,89 @@
...
+
+      - name: Install Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+
+      - name: Start Momento Local 2
+        run: |
+          docker run -d -p 8081:8080 gomomento/momento-local --return-error unavailable --error-rpcs get --error-count 2
```
anitarua, line 76: I think you'll need to use a different port for both sides, like: `-p 8081:8081`

And you'll need to get momento-local know to change port: `--port 8081`

.github/workflows/retry-tests.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Start Momento Local 1
+        run: |
+          docker run -d -p 8080:8080 gomomento/momento-local --return-error unavailable --error-rpcs get
+
+      - name: Set script permissions
+        run: chmod +x ./scripts/build-and-test-package-full-network-outage-retry-tests.sh
+
+      - name: Build and test with Momento Local 1
+        env:
+          TEST_URL: http://localhost:8080
```
anitarua, line 50: is `TEST_URL` an expected env var that the test client looks for? 
maybe we need to make an env var so the test knows which port to use when it constructs a momento-local credential provider

.github/workflows/retry-tests.yml
```diff
@@ -0,0 +1,89 @@
+name: Retry Tests
+
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  verify-user-permissions:
```
anitarua, line 8: also since we don't need api key, we don't need to verify permissions before running the retry tests

.github/workflows/retry-tests.yml
rishtigupta: I don't think that is necessary. In the case of Momento Local 2, inside the container, the application will run on 8080, but outside (on your local machine or in the CI environment), we will access it via 8081.

.github/workflows/retry-tests.yml
rishtigupta: This is what the build scripts look for I have hardcoded 8081 as of now for the temporary tests to run

.github/workflows/retry-tests.yml
rishtigupta: Yeah, I thought it will help me get rid of the chmod command, but it did not

.github/workflows/retry-tests.yml
anitarua: so overall docker command would look something like `docker run -d -p 8081:8081 gomomento/momento-local --port 8081 --return-error unavailable --error-rpcs get --error-count 2`

later, we can figure out a way to use the [native github actions syntax](https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/running-jobs-in-a-container#example-running-a-job-within-a-container) to maybe make docker setup and cleanup easier/automatic

anitarua: > I don't think that is necessary. In the case of Momento Local 2, inside the container, the application will run on 8080, but outside (on your local machine or in the CI environment), we will access it via 8081.

oh yeah that's right

.github/workflows/retry-tests.yml
```diff
@@ -0,0 +1,66 @@
...
+          node -v
+          ./scripts/build-and-test-package-full-network-outage-retry-tests.sh
+
+      - name: Clean up Docker Container
+        run: docker rm -f momento-local-1 || true
+
+  test-momento-local-2:
+    runs-on: ubuntu-latest
+    env:
+      MOMENTO_API_KEY: ${{ secrets.ALPHA_TEST_AUTH_TOKEN }}
```
anitarua, line 39: one more cleanup
```suggestion
```

.github/workflows/retry-tests.yml
```diff
@@ -0,0 +1,66 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
anitarua, line 44: another cleanup
```suggestion
```

packages/client-sdk-nodejs/test/integration/integration-setup.ts
```diff
@@ -56,19 +56,25 @@ export async function WithCache(
...
 ) {
   const cacheName = testCacheName();
   const testMiddleware = new TestRetryMetricsMiddleware(
     new DefaultMomentoLoggerFactory().getLogger('TestRetryMetricsMiddleware'),
     testMetricsCollector,
     v4()
   );
+  if (connectionPort) {
+    momentoLocalProvider = new MomentoLocalProvider({
+      port: connectionPort,
```
anitarua, line 75: I think changing `TEST_URL` to `MOMENTO_LOCAL_PORT` and `MOMENTO_LOCAL_HOSTNAME` and then initializing `new MomentoLocalProvider` using the env vars or a default value would be preferable to a hardcoded value in the tests

packages/client-sdk-nodejs/test/integration/integration-setup.ts
```diff
@@ -56,19 +56,25 @@ export async function WithCache(
   }
 }
 
-export const momentoLocalProvider = new MomentoLocalProvider();
+export let momentoLocalProvider = new MomentoLocalProvider();
```
anitarua, line 59: is it possible to initialize `momentoLocalProvider` inside of the `WithCacheAndCacheClient` scope instead of out here?

rishtigupta:
- hostname and port to be an env var
- move momento local provider into the cache and client method
- remove TEST_URL env var
- clean up comments

scripts/build-and-test-package-full-network-outage-retry-tests.sh
```diff
@@ -0,0 +1,18 @@
...
+
+set -x
+set -e
+
+ROOT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )"/.. &> /dev/null && pwd )
+
+echo "building and testing JS SDK with full network outage"
+
+${ROOT_DIR}/scripts/build-package.sh "core"
+${ROOT_DIR}/scripts/build-package.sh "common-integration-tests"
```
anitarua, line 11: nit: do we need to compile common-integration-tests if we're just running node-specific tests?
if we can remove this, it would save a little bit of CI time since building each package takes a while

anitarua: everything else lgtm though!

rishtigupta: Yes, it complains about not finding modules from core and common integ test.

rishtigupta: Actually, I have been trying so many things, let me try again by removing them and see if that works

rishtigupta: Yup, it complains about not finding the modules: https://github.com/momentohq/client-sdk-javascript/actions/runs/12919370823/job/36029750432?pr=1509

anitarua (approved):
lgtm then! exciting to see the new tests pass in 1min 45s 😄

.github/workflows/retry-tests.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: actions/checkout@v3
+
+      - name: Install Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+
+      - name: Start Momento Local 1
+        run: |
+          docker run -d -p 8080:8080 gomomento/momento-local --return-error unavailable --error-rpcs get
```
malandis, line 21: Once the request-based momento-local is available, we can just have one service that the tests can use right?

malandis (approved):
LGTM. Will have to revisit once the latest momento-local is out.

PR #1517, opened 2025-01-29
feat: add momento-local-test-config-middleware

rishtigupta:
## PR Description:
- Add Momento local test config middleware with "Experimental" label and unit tests for it
- Add MomentoErrorCodeMetadataConverter class and refactor tests based off of it.
- s/MomentoRpcMethodConverter/MomentoRPCMethodMetadataConverter, refactor tests
- Make the TestRetryMetricsMiddleware and ExperimentalMomentoLocalTestConfigMiddleware convert the enum values to metadata that momento-local can accept. 

## Issue
https://github.com/orgs/momentohq/projects/68/views/2

PR #1518, opened 2025-01-29
chore: move rpc method to main package, add 1:1 mapping for conversion

rishtigupta:
## PR Description:
- Move the MomentoRPCMethod to the main package so gthat we can make it accessible to the users when implementing the MOmentoLocalTestConfigMiddleware (https://github.com/momentohq/client-sdk-javascript/pull/1517)
- Replace the current metadata converter such that we have 1:1 mapping with the method strings accepted by momento-local. Ref: https://github.com/momentohq/momento-local/blob/main/src/cache/service/middlewares/rpc.rs#L51
- Refactor the tests to import the MomentoRPCMethod enum correctly.


anitarua (approved):
thanks!! will definitely help others avoid some of the issues I ran into when writing tests against momento-local

PR #1522, opened 2025-02-12
chore: add topic token refresh client example

rishtigupta:
## PR Description:
- Adds an example demonstrating how to use the Node.js SDK to generate disposable tokens for topics. The example includes functionality to refresh the client before the token expires.

## Issue
https://github.com/momentohq/dev-eco-issue-tracker/issues/1129


anitarua (commented):
lgtm overall! 

this is basically a nodejs version of the [web example](https://github.com/momentohq/client-sdk-javascript/blob/main/examples/web/cache/refresh-disposable-tokens.ts), right? could similar comments be added to this one to explain the key components?

anitarua (approved):

PR #1528, opened 2025-03-03
fix: manually update example dependencies

rishtigupta:
## PR Description:
Attempt to fix test failure in https://github.com/momentohq/client-sdk-javascript/pull/1527

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -491,7 +491,6 @@ jobs:
           pushd examples/nodejs/aws/lambda-examples/advanced-compression
             # This example shows how to build and package a lambda function for a different target architecture,
             # and thus it requires this `--force` flag.
-            rm -rf node_modules package-lock.json
             npm ci --force
             npm run build
           popd
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -491,6 +491,7 @@ jobs:
           pushd examples/nodejs/aws/lambda-examples/advanced-compression
             # This example shows how to build and package a lambda function for a different target architecture,
             # and thus it requires this `--force` flag.
+            rm -rf node_modules package-lock.json
             npm ci --force
             npm run build
           popd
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -491,6 +491,7 @@ jobs:
           pushd examples/nodejs/aws/lambda-examples/advanced-compression
             # This example shows how to build and package a lambda function for a different target architecture,
             # and thus it requires this `--force` flag.
+            rm -rf node_modules package-lock.json
```
malandis, line 494: We shouldn't have to do this; the `package-lock.json` should be consistent and `node_modules` is created when we install, not checked in, so no need to delete.

examples/nodejs/aws/lambda-examples/advanced-compression/package.json
```diff
@@ -11,11 +11,13 @@
   },
   "dependencies": {
     "@gomomento/sdk": "^1.108.0",
-    "@gomomento/sdk-nodejs-compression-zstd": "^0.92.2",
-    "@mongodb-js/zstd-linux-x64-gnu": "^1.2.0",
```
malandis, line 15: We're on major version 2 of `mongodb/zstd` in the compression library. Let's try marking the dependency as required (as before) but bump the major version.

malandis (commented):
Suggestion re: major version of library

.github/workflows/build.yml
rishtigupta: yeah, this was just for a trial run

examples/nodejs/aws/lambda-examples/advanced-compression/package.json
rishtigupta: Weird:
I do not see `@mongodb-js/zstd-linux-x64-gnu` registered at [npmjs](https://www.npmjs.com/search?q=%40mongodb-js%2Fzstd-linux-x64-gnu). It might have been deprecated?

rishtigupta: https://www.npmjs.com/package/@mongodb-js/zstd-linux-x64-gnu

PR #1529, opened 2025-03-04
fix: aws lambda zstd example needs postinstall step

malandis:
The current version of `@mongodb-js/zstd` has a deficiency loading the
native node module as compared to the previous major version. The
loader needs fallbacks to
`@mongodb/zstd/build/{Release,Debug}/zstd.node` to fully work when
bundling for lambda. We will open an issue with mongo about this.

In the interim we add a postinstall step to copy the file in the
`Debug` directory so the build will pass.

We also provide scripts to build and package for aws `npm run rebuild:aws` and dev `npm run rebuild:dev`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -489,9 +489,7 @@ jobs:
             npm run build
           popd
           pushd examples/nodejs/aws/lambda-examples/advanced-compression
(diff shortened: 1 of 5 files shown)


rishtigupta (approved):
:shipit:

rishtigupta (approved):
:shipit:

PR #1531, opened 2025-03-05
fix: force resolve to release path in nativeNodeModulesPlugina and include that in esbuild config

rishtigupta:
## PR Description:
- Removes `postInstall` script from `package.json`
- Updates the `nativeNodeModulesPlugin` in esbuild to force resolve path
- Add plugin to esbuild config


malandis (approved):
🎉

PR #1533, opened 2025-03-14
feat: add momento local tests for topics

rishtigupta:
## PR Description  
This commit introduces several improvements and fixes related to the topic client, including:  

- Adding middleware support for the topic client.  
- Renaming `TestMetricsMiddleware` to `MomentoLocalMiddleware` and refactoring the relevant code.  
- Adding momento local tests for topics using middleware.  
- Fixing the reconnect logic for subscriptions when an error occurs (details below).  

### **Improved Retry Logic in [`handleSubscribeError`](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/core/src/internal/clients/pubsub/AbstractPubsubClient.ts#L217)**
- The [`shouldReconnectSubscription` flag is always set to `true`](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/internal/pubsub-client.ts#L298), causing the client to always attempt a reconnect when an error occurs.  
- The [`sleep(reconnectDelayMillis)` function](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/core/src/internal/clients/pubsub/AbstractPubsubClient.ts#L268) introduces a delayed retry, which may still execute even after `unsubscribeFn` has been invoked.  

### **Race Condition Between Unsubscribe and Retry**  
- If an error occurs, [the retry logic schedules a `sendSubscribe` call after a 500ms delay.](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/core/src/internal/clients/pubsub/AbstractPubsubClient.ts#L263-L272)  
- However, if `unsubscribeFn` is called just before the retry executes, the subscription state might not be correctly checked, leading to unintended retries.  

### **Missing `subscriptionState.isSubscribed` Check in `handleSubscribeError`**  
- The [`prepareErrorCallback` function correctly verifies](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/internal/pubsub-client.ts#L290-L292) whether the subscription has already been unsubscribed before handling the error.  
- However, `handleSubscribeError` does not perform this check before retrying, which can result in an unnecessary or unintended reconnect attempt.  

### **Fix**  
To prevent unintended retries, a validation check has been added in `handleSubscribeError` to ensure that `unsubscribeFn` has not already been called before attempting a reconnect.

### DIFF
--- a/packages/core/src/utils/topic-call-options.ts
+++ b/packages/core/src/utils/topic-call-options.ts
@@ -40,4 +40,9 @@ export interface SubscribeCallOptions {
    * @param heartbeat The heartbeat received from the topic subscription.
    */
   onHeartbeat?: (heartbeat: TopicHeartbeat) => void;
+
+  /**
+   * The callback to invoke when the connection is lost.
+   */
+  onConnectionLost?: () => void;
 }
--- a/packages/client-sdk-nodejs/src/config/retry/momento-rpc-method.ts
+++ b/packages/client-sdk-nodejs/src/config/retry/momento-rpc-method.ts
@@ -43,6 +43,8 @@ enum MomentoRPCMethod {
   SortedSetGetRank = '_SortedSetGetRankRequest',
   SortedSetLength = '_SortedSetLengthRequest',
   SortedSetLengthByScore = '_SortedSetLengthByScoreRequest',
+  TopicPublish = '_PublishRequest',
+  TopicSubscribe = '_SubscriptionRequest',
 }
 
 class MomentoRPCMethodMetadataConverter {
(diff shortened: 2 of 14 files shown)


packages/core/src/utils/topic-call-options.ts
```diff
@@ -40,4 +40,9 @@ export interface SubscribeCallOptions {
    * @param heartbeat The heartbeat received from the topic subscription.
    */
   onHeartbeat?: (heartbeat: TopicHeartbeat) => void;
+
+  /**
+   * The callback to invoke when the connection is lost.
+   */
+  onConnectionLost?: () => void;
```
anitarua, line 47: should probably explain how this differs from the onError callback for end users. They won't receive an error or anything if they subscribe to this, so it would just be for logging or their own custom reconnect logic?

anitarua (commented):
looking good! just want to confirm that keepalive subscribe tests would be part of a follow-up pr?

rishtigupta:
Yes, the keep-alive tests would be a separate pr. I still need to discuss with Nate what all tests are we planning on writing for that.

packages/core/src/utils/topic-call-options.ts
rishtigupta: It works similar to the java sdk one: i.e custom logic could be added here by the user, and hence the function would be triggered when there is a lost connection to the stream.

anitarua (approved):

PR #1539, opened 2025-03-25
chore: bump generated-types on nodejs

malandis:
Bumps the generated-types dependency on the nodejs sdk to the latest
version (v `1.24.0`).

### DIFF
--- a/packages/client-sdk-nodejs/src/config/middleware/cache-request-logging.ts
+++ b/packages/client-sdk-nodejs/src/config/middleware/cache-request-logging.ts
@@ -856,6 +856,17 @@ const convertSortedSetLengthByScoreRequest: RequestToLogInterfaceConverterFn<
   };
 };
 
+// TODO elaborate on all the fields here. This is a stub.
+const convertSortedSetUnionStoreRequest: RequestToLogInterfaceConverterFn<
+  cache.cache_client._SortedSetUnionStoreRequest,
+  SortedSetRequestLog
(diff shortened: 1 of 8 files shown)


packages/client-sdk-nodejs/src/config/middleware/cache-request-logging.ts
```diff
@@ -61,6 +61,13 @@ const convertGetBatchRequest: RequestToLogInterfaceConverterFn<
   };
 };
 
+const convertGetWithHashRequest: RequestToLogInterfaceConverterFn<
```
rishtigupta, line 64: Do we support `withHash` apis?

rishtigupta:
Everything looks good, but I’m unclear on the purpose of the `withHash` APIs. Why do we need to add them?

rishtigupta (approved):
:shipit:

packages/client-sdk-nodejs/src/config/middleware/cache-request-logging.ts
malandis: No but the tests exhaustively check all the RPCs as defined in the protos, and will fail if we are missing one! It was a surprise to me.

anitarua: not yet but incoming, will first be implemented in go sdk but probably should be implemented in node soon after since it's our canonical sdk

PR #1540, opened 2025-03-25
feat: grpc channel state detection and auto-reconnect

rishtigupta:
## PR Description:
- Adds logic to inspect the gRPC channel’s ConnectivityState via client.getChannel().getConnectivityState(true).
- If the client’s channel is in TRANSIENT_FAILURE or SHUTDOWN state, the wrapper will now:
    - Log the reason
    - Close the existing client
    - Create a new client via the provided factory
- This is in addition to existing logic that rotates the client based on:
    - Idle timeout (maxIdleMillis)
    - Maximum client age (maxClientAgeMillis)

## Testing
- Added unit tests for the wrapper class

## Issue
work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/1197

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/grpc/grpc-client-wrapper.ts
+++ b/packages/client-sdk-nodejs/src/internal/grpc/grpc-client-wrapper.ts
@@ -1,7 +1,13 @@
+import {Channel} from '@grpc/grpc-js';
+
 export interface CloseableGrpcClient {
   close(): void;
 }
 
 export interface GrpcClientWrapper<T extends CloseableGrpcClient> {
   getClient(): T;
+}
+
+export interface GrpcClientWithChannel extends CloseableGrpcClient {
+  getChannel(): Channel;
 }
--- a/packages/client-sdk-nodejs/src/internal/grpc/idle-grpc-client-wrapper.ts
+++ b/packages/client-sdk-nodejs/src/internal/grpc/idle-grpc-client-wrapper.ts
@@ -56,7 +56,6 @@ export class IdleGrpcClientWrapper<T extends CloseableGrpcClient>
     const now = Date.now();
 
     // Reconnect if channel is in a bad state.
-    // Reconnect if channel is in a bad state.
     // Although the generic type `T` only extends `CloseableGrpcClient` (which doesn't define `getChannel()`),
     // we know that in practice, the client returned by `clientFactoryFn()` is a gRPC client that inherits from
     // `grpc.Client`: https://grpc.github.io/grpc/node/grpc.Client.html
--- a/packages/client-sdk-nodejs/src/internal/grpc/idle-grpc-client-wrapper.ts
+++ b/packages/client-sdk-nodejs/src/internal/grpc/idle-grpc-client-wrapper.ts
@@ -53,9 +53,9 @@ export class IdleGrpcClientWrapper<T extends CloseableGrpcClient>
   }
 
   getClient(): T {
-    let shouldRecreate = false;
-    let reason: string | undefined;
+    const now = Date.now();
 
+    // Reconnect if channel is in a bad state.
     // Reconnect if channel is in a bad state.
     // Although the generic type `T` only extends `CloseableGrpcClient` (which doesn't define `getChannel()`),
     // we know that in practice, the client returned by `clientFactoryFn()` is a gRPC client that inherits from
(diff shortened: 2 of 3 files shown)


packages/client-sdk-nodejs/src/internal/grpc/idle-grpc-client-wrapper.ts
```diff
@@ -48,31 +53,56 @@ export class IdleGrpcClientWrapper<T extends CloseableGrpcClient>
...
       }
     }
 
+    // Recreate the client if needed
+    if (shouldRecreate) {
+      this.logger.info(`${reason as string}; reconnecting client.`);
+      this.client.close();
+      this.client = this.clientFactoryFn();
+      this.clientCreatedTime = Date.now();
+    }
```
pgautier404, line 104: I think this would be more readable if this logic was in a function called somethig like `recreateClient(reason: string)` that returned a new client. You could then just call that method as soon as you detect the client needs to be reconstructed. So everywhere you currently have `shouldRecreate = true` can be replaced with 

```
this.client = recreateClient("reason text here");
retrurn;
```

That way you wouldn't need to track the `shouldRecreate` boolean or consider cases that aren't relevant.

anitarua (approved):
lgtm!

rishtigupta:
While gRPC provides built-in reconnection behavior for `TRANSIENT_FAILURE` (which we can fine-tune via channel options like `grpc.initial_reconnect_backoff_ms`), it does not attempt to reconnect once a channel reaches the `SHUTDOWN` state. At that point, the client is considered permanently unusable and must be explicitly recreated by the application.

This means that gRPC settings alone are insufficient to fully guard against persistent failures. For environments like EC2 (where long-lived connections are common), tuning gRPC reconnect backoff is helpful. But to robustly handle edge cases (like silent disconnects, idle timeouts, or unhandled SHUTDOWNs), we need application-level recovery logic — which we're addressing in the SDK via `IdleGrpcClientWrapper`.

In short: I think we can add both the approaches:
- gRPC settings = useful baseline safety net,
- SDK wrapper = essential for complete resilience.

wdyt? @malandis

rishtigupta:
If we decide to add the exponential backoff to default grpc settings, we would need these options:

```
'grpc.initial_reconnect_backoff_ms': 1000,       // (1s) The time between the first and second connection attempts 
'grpc.min_reconnect_backoff_ms': 1000,           // (1s) The min time between subsequent connection attempts
'grpc.max_reconnect_backoff_ms': 10000,          // (10s) The max time between subsequent connection attempts
```

Reference: https://grpc.github.io/grpc/core/group__grpc__arg__keys.html

rishtigupta:
More thoughts, especially when using lambda environment:

We already document in our codebase that [`grpc.keepalive_*` settings should generally be **disabled in AWS Lambda**](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/internal/grpc/grpc-channel-options.ts#L29-L33) environments. This is because the Lambda runtime can be suspended between invocations, and keepalive ACKs may not be received in time — leading to **false disconnects** and unnecessary channel teardown.

While gRPC does automatically attempt to recover from `TRANSIENT_FAILURE` using exponential backoff, this recovery is not always sufficient or reliable in Lambda, since:
- The client may be reused across invocations, but frozen in the middle of reconnect logic.
- There's no guarantee the reconnect timers or pings will fire before the next invocation begins.

Because of this, we should not rely solely on gRPC’s internal reconnect behavior. Instead, it’s important that we handle bad channel states like `TRANSIENT_FAILURE` explicitly in our `IdleGrpcClientWrapper`, alongside the existing `maxIdleMillis` and `maxClientAgeMillis` checks.

For longer-lived environments (e.g., EC2 or containers), gRPC's backoff policy can still provide value. To support this, we can optionally expose exponential backoff settings like:

```ts
'grpc.initial_reconnect_backoff_ms'
'grpc.min_reconnect_backoff_ms'
'grpc.max_reconnect_backoff_ms'
```

via our `GrpcConfiguration`, allowing clients to tune reconnection behavior in environments where it makes sense.

**TL;DR:**  
- Lambda: wrapper logic is essential, gRPC retry tuning has limited impact  
- EC2/containers: wrapper + gRPC backoff tuning together provide robust recovery  
- Proposal: expose reconnect tuning options in `GrpcConfiguration` so clients can opt in where appropriate (can be handled in a separate pr if that's acceptable)

packages/client-sdk-nodejs/src/internal/grpc/grpc-client-wrapper.ts
```diff
@@ -1,7 +1,13 @@
...
   close(): void;
 }
 
 export interface GrpcClientWrapper<T extends CloseableGrpcClient> {
   getClient(): T;
 }
+
+export interface GrpcClientWithChannel extends CloseableGrpcClient {
+  getChannel(): Channel;
+}
```
malandis, line 13: Could this method just be added to the `CloseableGrpcClient` interface? Or added as an additional generic constraint on `GrpcClientWrapper<T>`?

That way we can avoid the type erasure in `idle-grpc-client-wrapper`

packages/client-sdk-nodejs/src/internal/grpc/idle-grpc-client-wrapper.ts
```diff
@@ -48,32 +53,60 @@ export class IdleGrpcClientWrapper<T extends CloseableGrpcClient>
...
+
+    // Reconnect if channel is in a bad state.
+    // Although the generic type `T` only extends `CloseableGrpcClient` (which doesn't define `getChannel()`),
+    // we know that in practice, the client returned by `clientFactoryFn()` is a gRPC client that inherits from
+    // `grpc.Client`: https://grpc.github.io/grpc/node/grpc.Client.html
+    //
+    // Since `grpc.Client` defines a public `getChannel()` method, we can safely assume it's present on the client.
+    // To make TypeScript accept this, we cast the client to `unknown` first and then to `GrpcClientWithChannel`.
+    // This double-cast is necessary to override structural type checking and is safe in our controlled use case.
+    const clientWithChannel = this.client as unknown as GrpcClientWithChannel;
```
malandis, line 66: Can we just add `getChannel` to the `CloseableGrpcClient` interface? It's an internal interface so not a breaking change (see other comment).

malandis (approved):
Can follow up on outstanding issues later. 👍

packages/client-sdk-nodejs/src/internal/grpc/grpc-client-wrapper.ts
rishtigupta: We could, but then we lose flexibility — all `CloseableGrpcClient` implementations must now implement `getChannel()`, even if they don’t use it. Also, if we ever have another kind of client (e.g., a future control-plane gRPC client that doesn't expose channel), we would be forced to implement a no-op or stub `getChannel()` method. 
That being said, I doubt we would be using a different kind of channel in the near future, but just a callout for now.

rishtigupta: on second thought, we know are we are confident that all production clients will always expose getChannel() (i.e., they're always grpc.Client subclasses), we could add the method to `ClosableGrpcClient` without overthinking about my above comment. That would lead to simpler type hierarchy and no more casting `GrpcClientWithChannel` in our wrapper class.

rishtigupta: nvm, 

I explored simplifying the type system by moving the `getChannel(): Channel` method directly into the `CloseableGrpcClient` interface. The goal was to eliminate the need for casting in the `IdleGrpcClientWrapper` and reflect the fact that all gRPC clients expose this method.

However, in practice this change caused TypeScript type errors when instantiating the wrapper with gRPC-generated clients like `ScsControlClient`:

```
TS2322: Type 'ScsControlClient' is not assignable to type 'CloseableGrpcClient'
```

This happens because the type definitions generated by `@grpc/grpc-js` for these clients don’t explicitly declare `getChannel()` as part of their public interface — even though it exists at runtime.

Rather than introducing type assertions or manually wrapping every client to satisfy the extended interface, I’ve decided to **keep `getChannel()` in the `GrpcClientWithChannel` type** and retain the type cast inside the wrapper. This provides the right balance between type safety and flexibility without leaking gRPC internals into all client interfaces.

PR #1541, opened 2025-03-25
chore(nodejs): bump generated types and grpc-js to latest version

malandis:
Bumps the generated types dependency to the latest version, which also
bumps grpc-js to v1.13.1. We also bump the grpc-js dependency to match
that.


rishtigupta (approved):
:shipit:

PR #1543, opened 2025-03-26
fix: compression libraries don't need to depend on generated-types

malandis:
Previously the compression and compression-zstd libraries depended
directly on generated types. This is not necessary since they depend
on the nodejs SDK, which in turn depends on generated types.

This introduced the possibility of the generated types in the
compression (zstd) libraries having conflicting versions.


rishtigupta (approved):
:shipit:

PR #1545, opened 2025-03-27
chore: bump generated types to 0.124.3

malandis:
Updates the generated types dependencies to 0.124.3 where grpc is a peer dependency. This should resolve deduplication failures our users have seen when they install their own version of grpc.


rishtigupta (approved):
:shipit:

PR #1546, opened 2025-03-27
chore: bump generated types to 0.124.4

malandis:
Similar to #1545, we update the generated types but this time with protobuf as a peer dependency. This protects application developers from the same dual package hazard we saw with grpc.


rishtigupta (approved):
:shipit:

PR #1549, opened 2025-03-28
fix: make sure retry deadline does not exceed overall deadline

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1206

Assigns the Min(retry deadline, overall deadline) on each retry attempt to ensure we don't exceed the client's overall deadline.

Turns out that like the Java retry interceptor, the JS retry interceptor needs to manually check whether the current time exceeds the overall deadline. Otherwise, a timeout error is never returned. The status code on retry attempts do not change to DEADLINE_EXCEEDED if a retry attempt's deadline was exceeded; it remains the original status code and trying to change it will cause an UNKNOWN error.

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/grpc/retry-interceptor.ts
+++ b/packages/client-sdk-nodejs/src/internal/grpc/retry-interceptor.ts
@@ -114,11 +114,9 @@ export class RetryInterceptor {
                   return;
                 }
                 // Do not exceed the overall deadline when setting the retry attempt's deadline.
-                nextDeadline = new Date(
-                  Math.min(
-                    overallDeadline.getTime(),
-                    calculateDeadline(deadlineOffset).getTime()
-                  )
+                nextDeadline = calculateDeadline(
+                  deadlineOffset,
+                  overallDeadline
                 );
                 logger.debug(
                   `Setting next deadline (via offset of ${deadlineOffset} ms) to: ${nextDeadline.toISOString()}`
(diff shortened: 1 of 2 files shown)


packages/client-sdk-nodejs/src/internal/grpc/retry-interceptor.ts
```diff
@@ -107,7 +113,13 @@ export class RetryInterceptor {
...
                   return;
                 }
-                nextDeadline = calculateDeadline(deadlineOffset);
+                // Do not exceed the overall deadline when setting the retry attempt's deadline.
+                nextDeadline = new Date(
+                  Math.min(
+                    overallDeadline.getTime(),
+                    calculateDeadline(deadlineOffset).getTime()
+                  )
+                );
```
malandis, line 122: Would it make sense for this logic to be part of `calculateDeadline`? Eg an optional parameter with to clip the max value to.

rishtigupta:
Did we not add any momento-local tests for this scenario that needs to be updated?

anitarua:
> Did we not add any momento-local tests for this scenario that needs to be updated?

Adding a test for this scenario now, but weirdly it passes and returns deadline exceeded error (as I had been expecting) when the manual deadline check is removed, but other tests just hang. Looking into whether there's a bug in the retry interceptor now and will probably write more tests to cover these cases.

packages/client-sdk-nodejs/src/internal/grpc/retry-interceptor.ts
```diff
@@ -186,8 +227,11 @@ export class RetryInterceptor {
   }
 }
 
-function calculateDeadline(offsetMillis: number): Date {
+function calculateDeadline(offsetMillis: number, maxDeadline?: Date): Date {
   const deadline = new Date(Date.now());
   deadline.setMilliseconds(deadline.getMilliseconds() + offsetMillis);
+  if (maxDeadline && deadline > maxDeadline) {
```
malandis, line 233:
```suggestion
  if (maxDeadline !== undefined && deadline > maxDeadline) {
```

malandis (approved):
The comments are really helpful. Thanks!

rishtigupta (approved):
:shipit:

PR #1551, opened 2025-04-01
chore: let the old grpc client drain after configured timeout

rishtigupta:
## PR Description
This PR enhances the `IdleGrpcClientWrapper` to more gracefully handle gRPC client reconnections by introducing a new `ClientReaper` utility. This addresses scenarios where `.close()` was being called immediately on clients that may still have in-flight requests — potentially causing disruptions like `DEADLINE_EXCEEDED` errors.

### DIFF
--- a/packages/client-sdk-nodejs/src/internal/grpc/idle-grpc-client-wrapper.ts
+++ b/packages/client-sdk-nodejs/src/internal/grpc/idle-grpc-client-wrapper.ts
@@ -6,6 +6,27 @@ import {
 import {MomentoLogger, MomentoLoggerFactory} from '@gomomento/sdk-core';
 import {ConnectivityState} from '@grpc/grpc-js/build/src/connectivity-state';
 
+export class ClientReaper<T extends CloseableGrpcClient> {
+  private readonly reapingDelayMs: number;
+  private logger: MomentoLogger;
+
+  constructor(props: {delayMs?: number; loggerFactory: MomentoLoggerFactory}) {
+    this.reapingDelayMs = props.delayMs ?? 1000; // default 1 second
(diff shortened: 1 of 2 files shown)


packages/client-sdk-nodejs/src/internal/grpc/idle-grpc-client-wrapper.ts
```diff
@@ -50,6 +73,10 @@ export class IdleGrpcClientWrapper<T extends CloseableGrpcClient>
     this.lastAccessTime = Date.now();
     this.maxClientAgeMillis = props.maxClientAgeMillis;
     this.clientCreatedTime = Date.now();
+    this.clientReaper = new ClientReaper({
+      delayMs: 1000, // 1 second
+      loggerFactory: props.loggerFactory,
+    });
```
malandis, line 79: Waiting 1s doesn't guarantee the RPCs will have finished. Can we wait until the connection state is idle?

rishtigupta: @nand4011  observed that the customer is using an older version of the SDK that relies solely on idle-time-based client recycling. As a result, the recent improvements to the SDK (such as gRPC channel state checks) are not affecting their current implementation.

Additionally, we noticed the log message "Client has been idle" was printed twice in quick succession. This suggests that multiple concurrent requests were made just as the client was considered idle, causing the wrapper to attempt multiple recreations simultaneously.

Given this, we are deferring the more complex inflight request tracking strategy for now. As a short-term mitigation, we are introducing a simple flag to prevent multiple concurrent recreations. This ensures that only one client recreation occurs at a time - check this pr: https://github.com/momentohq/client-sdk-javascript/pull/1552

PR #1555, opened 2025-04-03
chore: remove storage tests

pgautier404:


malandis (commented):
All good on the committed changes. We also need to remove the storage examples from running in our GitHub actions.

rishtigupta (approved):
:shipit:

PR #1556, opened 2025-04-03
chore: remove storage example code

pgautier404:


rishtigupta:
The `doc-example-files` might still be referenced in the Momento docs, which could cause issues if we remove them now. Should we check with Allen first to confirm whether the references to the storage sections have been removed? Alternatively, we could leave the `doc-example-files` in place for now and revisit this later.

pgautier404:
> The `doc-example-files` might still be referenced in the Momento docs, which could cause issues if we remove them now. Should we check with Allen first to confirm whether the references to the storage sections have been removed? Alternatively, we could leave the `doc-example-files` in place for now and revisit this later.

The example files are pulled in dynamically, so removing them will just cause them to be removed from the docs. The docs for storage aren't even linked to from the rest of the site because storage is an allow-listed preview that was never publicly available, so I'm not worried about them disappearing. As far as I know this shouldn't cause any problems with the docs process itself, but if it does we can revert the commit.

malandis (commented):
I was suggesting just removing the steps in GitHub actions that runs the examples (vs removing the example code). That's the minimum we can do at this stage to ensure CI doesn't fail.

rishtigupta (approved):
:shipit:

pgautier404:
> I was suggesting just removing the steps in GitHub actions that runs the examples (vs removing the example code). That's the minimum we can do at this stage to ensure CI doesn't fail.

Running the examples were never part of the canaries, so it was never strictly necessary to remove them as part of this effort. However, since I'm in these files anyway, it seems a bit of a waste of time to not just remove them while I'm here. We've already removed the Go examples and API docs snippets, so I'd vote we go ahead and finish what we've started here.

I have a PR open for the dev docs that removes the storage pages from the node, web, and java docs. I'd like to merge this and the related java sdk PR once that's approved and merged, but I'll hold off on merging these docs examples PRs until we've had a chance to discuss.

malandis:
> We've already removed the Go examples and API docs snippets, so I'd vote we go ahead and finish what we've started here.

Ah yeah that's right. Thanks for the reminder. Lots of moving parts here!

malandis (approved):

PR #1559, opened 2025-04-08
fix: fixed timeout retry strategy should overwrite deadline only after first request fails

anitarua:
Work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/1214 

The JS fixed timeout retry strategy was overwriting the grpc deadline from the initial request when it should update the deadline only after the initial request fails. Once the initial request fails, the retry deadline (responseDataReceivedTimeoutMillis) should be used.

Updates the tests to ensure this logic is exercised.

### DIFF
--- a/packages/client-sdk-nodejs/src/config/retry/fixed-timeout-retry-strategy.ts
+++ b/packages/client-sdk-nodejs/src/config/retry/fixed-timeout-retry-strategy.ts
@@ -6,7 +6,7 @@ import {EligibilityStrategy} from './eligibility-strategy';
 import {MomentoLoggerFactory, MomentoLogger} from '../..';
 import {DefaultStorageEligibilityStrategy} from './storage-default-eligibility-strategy';
 import {Status} from '@grpc/grpc-js/build/src/constants';
-import {getCurrentTimeAsDateObject} from '../../internal/utils';
+import {hasExceededDeadlineRelativeToNow} from '../../internal/utils';
 
 export interface FixedTimeoutRetryStrategyProps {
   loggerFactory: MomentoLoggerFactory;
@@ -47,7 +47,7 @@ export class FixedTimeoutRetryStrategy implements RetryStrategy {
     if (
       props.attemptNumber > 0 &&
       props.grpcStatus.code === Status.DEADLINE_EXCEEDED &&
-      props.overallDeadline >= getCurrentTimeAsDateObject()
+      hasExceededDeadlineRelativeToNow(props.overallDeadline)
     ) {
       this.logger.debug(
         `Request is eligible for retry (attempt ${props.attemptNumber}), retrying after ${this.retryDelayIntervalMillis} ms +/- jitter.`
--- a/packages/client-sdk-nodejs/src/config/retry/fixed-timeout-retry-strategy.ts
+++ b/packages/client-sdk-nodejs/src/config/retry/fixed-timeout-retry-strategy.ts
@@ -47,7 +47,7 @@ export class FixedTimeoutRetryStrategy implements RetryStrategy {
     if (
       props.attemptNumber > 0 &&
       props.grpcStatus.code === Status.DEADLINE_EXCEEDED &&
-      props.overallDeadline > getCurrentTimeAsDateObject()
+      props.overallDeadline >= getCurrentTimeAsDateObject()
     ) {
       this.logger.debug(
         `Request is eligible for retry (attempt ${props.attemptNumber}), retrying after ${this.retryDelayIntervalMillis} ms +/- jitter.`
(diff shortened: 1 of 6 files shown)


packages/client-sdk-nodejs/src/config/retry/fixed-timeout-retry-strategy.ts
```diff
@@ -39,6 +40,20 @@ export class FixedTimeoutRetryStrategy implements RetryStrategy {
     this.logger.debug(
       `Determining whether request is eligible for retry; status code: ${props.grpcStatus.code}, request type: ${props.grpcRequest.path}, attemptNumber: ${props.attemptNumber}`
     );
+
+    // If a retry attempt's timeout has passed but the client's overall timeout has not yet passed,
+    // we should reset the deadline and retry.
+    if (
+      props.attemptNumber > 0 &&
+      props.grpcStatus.code === Status.DEADLINE_EXCEEDED &&
+      props.overallDeadline > new Date(Date.now())
```
malandis, line 49: Can you audit the code for `new Date(Date.now())`? I see some inconsistencies in `retry-strategy.ts` specifically for comparing now vs the overall deadline. Can consider extracting to a helper.

malandis (commented):
Tests are clear; just one concern re: comparing the deadline to the current time. In some parts of the code we use strict inequality; others not.

packages/client-sdk-nodejs/src/config/retry/fixed-timeout-retry-strategy.ts
anitarua: oh good call, will extract to a helper which should help with consistency too

packages/client-sdk-nodejs/src/internal/grpc/retry-interceptor.ts
```diff
@@ -65,14 +65,16 @@ export class RetryInterceptor {
           deadlineOffset
         )}`
       );
-      const overallDeadline = calculateDeadline(overallRequestTimeoutMs);
 
+      // The first attempt at a request should use the overall request timeout,
+      // not a retry strategy-specific timeout (i.e. responseDataReceivedTimeoutMillis).
+      const overallDeadline = calculateDeadline(overallRequestTimeoutMs);
```
rishtigupta, line 71: Just confirming: overallRequestTimeoutMs is the duration after which the client should timeout, while `calculateDeadline` helper calculates the exact timestamp at which that happens, right?

anitarua: correct, yes

rishtigupta:
Based on the behavior explained [here](https://github.com/momentohq/dev-eco-issue-tracker/issues/1214), the PR lgtm!
will 👍 once Michael's comments are addressed.

rishtigupta (approved):
:shipit:

packages/client-sdk-nodejs/src/internal/utils.ts
```diff
@@ -4,3 +4,13 @@ export function convert(v: string | Uint8Array): Uint8Array {
   }
   return v;
 }
+
+export function getCurrentTimeAsDateObject(): Date {
+  return new Date();
```
malandis, line 9: Is this equivalent to what we had before? `new Date(Date.now())`

packages/client-sdk-nodejs/src/internal/grpc/retry-interceptor.ts
```diff
@@ -116,23 +122,24 @@ export class RetryInterceptor {
                 //
                 // We also need this check in case DEADLINE_EXCEEDED is marked as a retryable status code
                 // as it is in the default storage eligibility strategy.
-                if (new Date(Date.now()) >= overallDeadline) {
+                if (getCurrentTimeAsDateObject() >= overallDeadline) {
```
malandis, line 125: can replace this and the other instances with something like `hasExceededDeadlineRelativeToNow(overallDeadline)` or some better now.

packages/client-sdk-nodejs/src/internal/utils.ts
anitarua: yes

rishtigupta: I think the function coded below it: `createDateObjectFromUnixMillisTimestamp` is the same as what Michael is saying?

anitarua: yeah, `new Date()` can accept a timestamp as part of the constructor or it could be left empty and it'll create an object based on current date and time ([reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#return_value))

rishtigupta (approved):

PR #1564, opened 2025-05-19
fix: add bookkeeping for number of active subscriptions

anitarua:
- **fix: add bookkeeping for number of active subscriptions**
- **chore: add tests that can be run locally to test topic subscription bookkeeping**


malandis (dismissed):
One suggestion that's not a blocker. Very thorough tests as usual; I appreciate that!

PR #1565, opened 2025-06-12
feat: add subscription retry strategy and tests

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1172

Adds a subscription retry strategy and eligibility strategy to the nodejs sdk.
Allows for configuring how long to delay between reconnect attempts and for what error codes to retry.

Resubscribe logic is spread between the internal nodejs pubsub client and the core package's abstract pubsub client, so had to avoid grpc-specific elements in the retry and eligibility strategies.


packages/client-sdk-nodejs/src/config/retry/topic-subscription-retry-strategy.ts
```diff
@@ -0,0 +1,51 @@
...
+  private readonly logger: MomentoLogger;
+  readonly eligibilityStrategy: SubscriptionRetryEligibilityStrategy;
+  readonly retryDelayMillis: number;
+
+  constructor(props: TopicSubscriptionRetryStrategyProps) {
+    this.logger = props.logger;
+    this.eligibilityStrategy =
+      props.eligibilityStrategy ??
+      new TopicSubscriptionRetryEligibilityStrategy(props.logger);
+    this.retryDelayMillis = props.retryDelayMillis ?? 500;
```
malandis, line 31: We should externalize the default to the core package in one place.

malandis (commented):
Looks good. One minor suggestion.

packages/client-sdk-nodejs/src/config/retry/topic-subscription-retry-strategy.ts
```diff
@@ -5,6 +5,7 @@ import {
   SubscriptionRetryStrategy,
 } from '@gomomento/sdk-core';
 import {TopicSubscriptionRetryEligibilityStrategy} from './topic-subscription-retry-eligibility-strategy';
+import {DEFAULT_SUBSCRIPTION_RETRY_DELAY_MILLIS} from '@gomomento/sdk-core/dist/src/config/retry/subscription-retry-strategy';
```
malandis, line 8: Should also export the name in the `index.ts` of core to avoid the source-based lookup.

PR #1567, opened 2025-07-30
chore: Improved Runtime-Version header info

alubbad:
The `Runtime-Version` header to the web SDK previously used a getBrowserName function found in Mozilla docs to extract the web platform from the navigator.userAgent object. 
I removed this function and instead added a `UAParser` to parse information from navigator.userAgent about the browser, device, and OS used. This information has been added to the `Runtime-Version` header.


malandis (commented):
Looks good. We can remove the dead code.

malandis (approved):

PR #1569, opened 2025-08-05
chore: Refactored Credential Provider Endpoint Struct

alubbad:
The Endpoint class for CredentialProvider was written assuming that individual endpoints could be overridden. However, only base endpoint or all endpoint overrides are allowed. This means we were storing redundant information about whether each endpoint is secure. To simplify CredentialProvider, I got rid of the Endpoint struct and instead added the isSecure boolean to AllEndpoints and made each endpoint a string.


anitarua (approved):
lgtm!

PR #1570, opened 2025-08-08
feat: remove deprecated storage client and references from updated protos

tylerburdsall:
Removes our now-defunct Storage-related clients, types, etcs. upon upgrading our client protos


malandis (approved):
lgtm. I verified there are no examples that exercise these code paths either. Once merged we should do a clean release.

PR #1572, opened 2025-08-12
chore: added missing sortedSetUnionStore API

alubbad:
I added the sortedSetUnionStore API to the NodeJS and Web SDKs. However, the [length()](packages/core/src/messages/responses/cache-sorted-set-union-store.ts) method from the Success response is returning wrong values in the tests.


packages/core/src/index.ts
```diff
@@ -298,6 +299,7 @@ export {
   CacheSortedSetRemoveElements,
   CacheSortedSetLength,
   CacheSortedSetLengthByScore,
+  CacheSortedSetUnionStore,
```
anitarua, line 302: I think you might also have to export SortedSetUnionStoreOptions, SortedSetSource, and SortedSetAggregate as well so that they're publicly available

packages/client-sdk-web/src/internal/cache-data-client.ts
```diff
@@ -4115,6 +4125,121 @@ export class CacheDataClient<
...
+    for (const source of sources) {
+      sortedSources.push(this.convertSortedSetSource(source));
+    }
+    const request = new _SortedSetUnionStoreRequest();
+    request.setSetName(sortedSetName);
+    request.setSourcesList(sortedSources);
+    request.setAggregate(aggregate);
+    request.setTtlMilliseconds(this.ttlOrDefaultMilliseconds(options?.ttl));
+
+    // return await new Promise((resolve, reject) => {
```
anitarua, line 4197: looks like some commented out code can be removed in this file

packages/core/src/internal/clients/cache/AbstractCacheClient.ts
```diff
@@ -1596,7 +1599,33 @@ export abstract class AbstractCacheClient implements ICacheClient {
...
+   * sotring the result of the union
+   * @param {string} cacheName - The cache containing the sorted set.
+   * @param {string} sortedSetName - The sorted set name.
+   * @param {sources} SortedSetSource[] -
+   * @param {number} [options.minScore] - The lower bound on the score range to search in.
+   * @param {number} [options.maxScore] - The upper bound on the score range to search in.
+   * @returns {Promise<CacheSortedSetLengthByScore.Response>}
+   * {@link CacheSortedSetLengthByScore.Hit} containing the length if the sorted set exists.
+   * {@link CacheSortedSetLengthByScore.Miss} if the sorted set does not exist.
+   * {@link CacheSortedSetLengthByScore.Error} on failure.
```
anitarua, line 1613: looks like the parameters in the docs comment hasn't been updated yet

packages/core/src/internal/clients/cache/IDataClient.ts
```diff
@@ -370,6 +373,12 @@ export interface IDataClient {
     minScore?: number,
     maxScore?: number
   ): Promise<CacheSortedSetLengthByScore.Response>;
+  sortedSetUnionStore(
+    cacheName: string,
+    sortedSetName: string,
+    sources: SortedSetSource[],
+    options?: SortedSetUnionStoreCallOptions
```
anitarua, line 380: I think these options should be unpacked into the individual options in this interface (refer to sortedSetLengthByScore and similar methods).

AbstractCacheClient will do the mapping from options object to the individual options

packages/core/src/utils/cache-call-options.ts
```diff
@@ -143,6 +142,23 @@ export interface SortedSetLengthByScoreCallOptions {
...
+export interface SortedSetUnionStoreCallOptions {
+  /**
+   * A function to determine the final score for an element that exists in multiple source sets.
+   * If the function is not specified, the aggregate function defaults to SUM.
+   */
+  aggregate?: SortedSetAggregate;
+  /**
+   * The time to live in seconds of the object being modified.
+   */
+  ttl?: number;
```
anitarua, line 160: should be `CollectionTtl` type since the ttl would be applied to the sorted set collection. 

more info about CollectionTtl in the docs: https://docs.momentohq.com/cache/develop/api-reference/collection-ttl

packages/core/src/internal/utils/validators.ts
```diff
@@ -239,3 +239,38 @@ export function validateMaxConcurrentRequests(limit: number) {
     );
   }
 }
+
+export function validateAggregate(aggregate?: SortedSetAggregate) {
```
anitarua, line 243: because SortedSetAggregate is an enum, you probably don't need this validator. end users should be able to pass in only the allowed values anyway

packages/core/src/internal/utils/validators.ts
```diff
@@ -239,3 +239,38 @@ export function validateMaxConcurrentRequests(limit: number) {
...
+      `Invalid aggregate value: ${aggregate}. Valid values are ${validValues.join(
+        ', '
+      )}`
+    );
+  }
+}
+
+export function validateSortedSetSources(sources: SortedSetSource[]) {
+  if (!sources) {
+    throw new InvalidArgumentError('source is required');
```
anitarua, line 261: more descriptive error would be good, perhaps "at least one source sorted set is required"
(or is at least two sets required? I'm not sure we've written any tests to check that yet)

packages/client-sdk-nodejs/src/index.ts
```diff
@@ -395,6 +396,7 @@ export {
   CacheSortedSetRemoveElements,
   CacheSortedSetLength,
   CacheSortedSetLengthByScore,
+  CacheSortedSetUnionStore,
```
anitarua, line 399: similarly, the sorted set union store aggregate and sources types might have to be exported from the nodejs and web sdk index.ts files too

can verify by importing the local version of the packages into some example code and writing some code using the sdk

packages/client-sdk-web/src/internal/cache-data-client.ts
```diff
@@ -4115,6 +4125,121 @@ export class CacheDataClient<
...
+    sortedSetName: Uint8Array,
+    sources: SortedSetSource[],
+    options?: SortedSetUnionStoreCallOptions
+  ): Promise<CacheSortedSetUnionStore.Response> {
+    let aggregate: _SortedSetUnionStoreRequest.AggregateFunction;
+    if (options?.aggregate === undefined) {
+      aggregate = _SortedSetUnionStoreRequest.AggregateFunction.SUM;
+    } else {
+      aggregate = this.convertAggregateResult(options.aggregate);
+    }
```
anitarua, line 4186: seems like convertAggregateResult could maybe also handle the undefined case?

packages/client-sdk-web/src/internal/cache-data-client.ts
```diff
@@ -4115,6 +4125,121 @@ export class CacheDataClient<
...
+
+    this.logger.trace(
+      "Issuing 'sortedSetUnionStore' request; sources: %s, aggregate: %s:, ttl: %s",
+      sources.toString(),
+      options?.aggregate?.toString() ?? 'null',
+      options?.ttl?.toString() ?? 'null'
+    );
+    const result = await this.sendSortedSetUnionStore(
+      cacheName,
+      this.convertToUint8Array(sortedSetName),
```
anitarua, line 4165: Seems like other places in this file use `convertToB64String` instead of `convertToUint8Array` on strings

packages/client-sdk-web/src/internal/cache-data-client.ts
```diff
@@ -4654,4 +4779,29 @@ export class CacheDataClient<
         return ItemType.SORTED_SET;
     }
   }
+
+  private convertAggregateResult(
```
anitarua, line 4783: nice helper functions!

packages/client-sdk-web/src/internal/cache-data-client.ts
```diff
@@ -4115,6 +4125,121 @@ export class CacheDataClient<
...
+          ...this.clientMetadataProvider.createClientMetadata(),
+          ...createCallMetadata(cacheName, this.deadlineMillis),
+        },
+        (err, resp) => {
+          if (resp) {
+            resolve(new CacheSortedSetUnionStore.Success(resp.getLength()));
+          } else {
+            this.cacheServiceErrorMapper.resolveOrRejectError({
+              err: err,
+              errorResponseFactoryFn: e => new CacheSet.Error(e),
```
anitarua, line 4233:
```suggestion
              errorResponseFactoryFn: e => new CacheSortedSetUnionStore.Error(e),
```

packages/client-sdk-nodejs/src/internal/cache-data-client.ts
```diff
@@ -4138,6 +4169,85 @@ export class CacheDataClient implements IDataClient {
...
+  }
+
+  private async sendSortedSetUnionStore(
+    cacheName: string,
+    sortedSetName: Uint8Array,
+    sources: SortedSetSource[],
+    options?: SortedSetUnionStoreCallOptions
+  ): Promise<CacheSortedSetUnionStore.Response> {
+    let aggregate: grpcCache._SortedSetUnionStoreRequest.AggregateFunction;
+    if (options?.aggregate === undefined) {
```
anitarua, line 4210: similar suggestion here as in the web internal data client

anitarua (commented):
looking good overall!

packages/client-sdk-web/src/index.ts
```diff
@@ -211,6 +213,8 @@ export {
   TopicName,
   CacheSelector,
   TopicSelector,
+  SortedSetAggregate,
+  SortedSetSource,
```
anitarua, line 217: does web also need to export the sorted set union store response class?

packages/client-sdk-web/src/internal/cache-data-client.ts
```diff
@@ -4115,6 +4123,88 @@ export class CacheDataClient<
     });
   }
 
+  public async sortedSetUnionStore(
+    cacheName: string,
+    sortedSetName: string,
+    sources: SortedSetSource[],
+    aggregate?: SortedSetAggregate,
+    ttl?: CollectionTtl
```
anitarua, line 4131: I think you can provide a default argument here, like how sortedSetPutElements does it ([example](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-web/src/internal/cache-data-client.ts#L3484))

packages/core/src/internal/utils/validators.ts
```diff
@@ -239,3 +239,22 @@ export function validateMaxConcurrentRequests(limit: number) {
     );
   }
 }
+
+export function validateSortedSetSources(sources: SortedSetSource[]) {
+  if (!sources) {
+    throw new InvalidArgumentError('At least one sources is required');
```
anitarua, line 245: thanks for adding the test "errors when one source is provided"! 
I think this error message should be updated to require at least 2 sources

anitarua (commented):
looking good! few more comments

packages/client-sdk-web/src/internal/cache-data-client.ts
alubbad: I tried that but it said "Parameter cannot have question mark and initializer."

packages/core/src/internal/utils/validators.ts
alubbad: oh I had that initially but it turns out you can input one source set and it would still work.

packages/client-sdk-web/src/internal/cache-data-client.ts
anitarua: Let's go with the default argument then, can remove the `?`

PR #1574, opened 2025-08-12
feat: add support for functions-related permissions in key/token generation

tylerburdsall:
With the addition of the latest client protos, we can add support for restricting access to specific functions (currently a preview feature) within specific caches


anitarua (commented):
looking good! I think just missing exports of new permissions stuff like `AllFunctions` ([reference](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/core/src/index.ts#L170-L192))

anitarua (commented):
found a couple more spots to export as well:

- [node sdk](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/index.ts#L154)
- [web sdk](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-web/src/index.ts#L137)

PR #1579, opened 2025-08-15
feat: Added getWithHash and setWithHash APIs

alubbad:
I added the missing getWithHash and setWithHash APIs to the Nodejs and Web sdks.


packages/core/src/index.ts
```diff
@@ -154,6 +156,7 @@ export {
   DictionarySetFieldOptions,
   DictionarySetFieldsOptions,
   SetOptions,
+  SetWithHashOptions,
```
anitarua, line 159: also I think both GetWithHashOptions and SetWithHashOptions needs to be exported from both core and node packages

alubbad: oh, I'm not sure they're needed in the node package because GetOptions is not exported either, and none of the SetIf options are, so I'm assuming they use SetOptions instead, since they're all the same.

anitarua: I think if SetOptions is exported, it must've been an oversight that GetOptions was not also exported.
From local testing, I think it would be nice if we export GetOptions, SetOptions, GetWithHashOptions, and SetWithHashOptions, just in case.

It would allow for writing code with explicit typing and more IDE support. Example:
```typescript
  const setOptions: SetWithHashOptions = {
    compress: false,
  };
  const setResponse = await momento.setWithHash('cache', 'foo', 'FOO', setOptions);
```

alubbad: ohh okay then I'll export all of them. thanks for the explanation!

anitarua (approved):
lgtm!

PR #1584, opened 2025-08-20
feat: ability to use AbortSignal to cancel get and set requests

anitarua:
Work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/1255

Adds a new optional `AbortSignal` argument to `get` and `set` methods to allow users to cancel requests before exiting a serverless (e.g. Lambda) environment.

In the `sendGet` and `sendSet` methods of the internal data client, an event listener waits for an `abort` event to occur, then invokes the grpc-provided `cancel()` method. The `CancellationInterceptor` is also added to the list of interceptors for `get` and `set` to avoid a retry strategy foot-gun.

Added tests to verify the `AbortSignal` works when both `NoRetryStrategy` and the default `FixedCountRetryStrategy` are used. The `CancellationInterceptor` helps to short-circuit retry attempts by using an `ABORTED` grpc status code instead of `CANCELLED`, and the error converter checks both cases to make sure we return a `CancelledError` for the `AbortSignal` case.

### DIFF
--- a/packages/client-sdk-nodejs/test/integration/cache/abort-signal.test.ts
+++ b/packages/client-sdk-nodejs/test/integration/cache/abort-signal.test.ts
@@ -26,7 +26,7 @@ describe('AbortSignal', () => {
       if (setResponse.type === CacheSetResponse.Error) {
         expect(setResponse.errorCode()).toEqual('CANCELLED_ERROR');
         expect(setResponse.message()).toContain(
-          'Request cancelled by an AbortSignal'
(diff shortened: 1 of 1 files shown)


allenheltondev (changes requested):

allenheltondev (dismissed):
Thank you for the quick changes! These look good to me, but I see the build is failing on node 18. Happy to re-review if that requires another change

allenheltondev (approved):

PR #1586, opened 2025-08-21
fix: make AbortSignal tests less flaky

anitarua:
In low latency environments, like in our canaries, these tests can become flaky when the get/set commands complete before the AbortSignal can be processed.

This PR updates the tests so that each one runs 3 times and it expects at least one to pass. This should hopefully make canary alerts less noisy.


honeyAcorn (approved):
Looks go to me but it would be nice if @malandis can approve it :)

PR #1611, opened 2026-05-07
fix: pin ua-parser-js to v1.0.41 (last MIT release)

eaddingtonwhite:
ua-parser-js v2.x switched to AGPL-3.0-or-later, requiring a paid PRO license for commercial use. Pin to v1.0.41 — still actively maintained on the v1.x branch — to keep @gomomento/sdk-web license-clean for downstream consumers.

Side benefit: drops 8 packages from the dep graph that v2.0.4 pulled in (node-fetch, tr46, whatwg-url, ua-is-frozen, detect-europe-js, is-standalone-pwa, @types/node-fetch, webidl-conversions). v1.0.41 has zero runtime dependencies.

Call site in client-metadata-provider.ts is unchanged — v1 supports the same UAParser(navigator.userAgent) call shape. Adds @types/ua-parser-js as a devDependency since v1 does not ship its own type declarations.

### DIFF
--- a/packages/client-sdk-web/package-lock.json
+++ b/packages/client-sdk-web/package-lock.json
@@ -15,13 +15,14 @@
         "google-protobuf": "3.21.2",
         "grpc-web": "1.4.2",
         "jwt-decode": "3.1.2",
-        "ua-parser-js": "2.0.4"
+        "ua-parser-js": "1.0.41"
       },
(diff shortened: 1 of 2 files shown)


ben-kugler (approved):

danielamiao:
Thanks!

